PR #3, opened 2025-03-04
Update integrators format

abhijeetgangan:
Integrators ~~and optimizers~~ now take model are a arg and return `init` and `update` functions. `init` function takes a dict of inputs like positions. masses etc or the `BaseState`.

### DIFF
--- a/torchsim/runners.py
+++ b/torchsim/runners.py
@@ -136,11 +136,11 @@ def optimize(
         )
     state: BaseState = initialize_state(system, model.device, model.dtype)
 
-    state, update_fn = optimizer(
-        state=state,
+    init_fn, update_fn = optimizer(
         model=model,
         **optimizer_kwargs,
     )
+    state = init_fn(state)
 
     step: int = 1
     last_energy = state.energy + 1


orionarcher:
```python

StateDict = Dict[Literal["positions", "masses", "cell", "pbc", "atomic_numbers"], torch.Tensor]

    def nve_init(state: Union[BaseState, StateDict], kT: torch.Tensor, **extra_state_kwargs: Any):
        """Initialize an NVE state from input data.
        
        Args:
            input_data: Either a BaseState object or a dictionary containing positions, masses, cell, pbc
            kT: Temperature in energy units for initializing momenta
            **extra_state_kwargs: Additional state arguments
            
        Returns:
            MDState: Initialized state for NVE integration
        """
        device = model.device
        dtype = model.dtype


        # Extract required data from input
        if not isinstance(state, BaseState):
            state = BaseState(**state)

        # Override with extra_state_kwargs if provided
        atomic_numbers = extra_state_kwargs.get("atomic_numbers", atomic_numbers)

        model_output = model(
            positions=state.positions,
            cell=state.cell,
            atomic_numbers=state.atomic_numbers,
            batch=state.batch
        )

        momenta = (
            extra_state_kwargs.get("momenta")
            if extra_state_kwargs.get("momenta") is not None
            else calculate_momenta(positions, masses, kT, device, dtype)
        )

        initial_state = MDState(
            positions=state.positions,
            momenta=state.momenta,
            energy=model_output["energy"],
            forces=model_output["forces"],
            masses=state.masses,
            cell=state.cell,
            pbc=state.pbc,
            atomic_numbers=state.atomic_numbers,
        )
        return initial_state
  ```
  
  What do you think of this?
  
  Thinking behind changes.
  1. more descriptive name for main argument, clearer typing
  2. reduced verbosity by instantiating state rather than unpacking it
  3. as I understand python, this shouldn't change any logic

abhijeetgangan:
Maybe the `input_data` shouldn't be called state. Maybe `input_state` / `initial_state`

orionarcher:
Other places where state is an input the variable is called `state`, (like the update function or temperature or kinetic_energy) which is why I favor that convention. What's your thinking behind `input_state`?

abhijeetgangan:
I don't think very basic functions should take in a state. Users should be able to pass any tensor without creating a state.

orionarcher:
The function is returning `MDState` which inherits from `BaseState`, so they'll be creating a `BaseState` either way. Does creating it slightly earlier change any important logic?

EDIT: maybe misread, I am thinking that allowing the dict will prevent them from having to create a state outside the function

abhijeetgangan:
Let me be clear. In the integrators and optimizers it is fine. I am saying if you want to compute kinetic energy why would you need a state to be initialized? Maybe I want to use such functions outside the integrators. Why would one need to create a base state?

orionarcher:
Ahh yeah I misremembered the signature for `temperature` and `kinetic_energy` didn't mean to suggest they should take state.

I am thinking having one name for `state` in functions that accept it would be a good standard. Rather than having `input_state` here. By the same notion, we could call the state in update_fn `update_state`, but I think having a single standard is preferable.

orionarcher:
Otherwise I think the change looks good. If we want to call it input state that could help differentiate `BaseState` from `BaseState | StateDict`

abhijeetgangan:
I think then we can give users the three options. First with existing methods they can either use a `BaseState` (preferable) or a `StateDict` (more flexible?). If they want to their own method they can use the same logic basically having their own `State` inherit the `BaseState` or create a new `StateDict` with more data keys.

For the update `State` it has to match the state which it is trying to update.

So I am okay with having a `State` everywhere but users may pass it directly to `update_fn` and that won't work ofc.

PR #6, opened 2025-03-04
Run example scripts in GitHub Actions

janosh:
to ensure a good UX for newcomers, we should make sure all example scripts actually work

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -136,7 +136,18 @@ jobs:
       - name: Find example scripts
         id: set-matrix
         run: |
-          EXAMPLES=$(find examples -name "*.py" | jq -R -s -c 'split("\n")[:-1]')
+          # Find all example scripts but exclude known failing ones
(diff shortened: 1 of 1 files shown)


examples/9_Others/9.4_Elastic_Tensor_MACE.py
```diff
@@ -1,25 +1,31 @@
...
     BravaisType,
+    ElasticState,
     get_elastic_tensor,
+    get_elementary_deformations,
     get_full_elastic_tensor,
 )
-from mace.calculators.foundations_models import mace_mp
 
 
 def get_bravais_type(
```
janosh: @abhijeetgangan can we import `get_bravais_type` from somewhere? it's very long. i assume you copy pasted it? ideally replace with `moyopy`

abhijeetgangan: We can put it in the elastic module or we can use moyopy for this example. Whatever works.

PR #9, opened 2025-03-04
Update optimizers

abhijeetgangan:


orionarcher:
Still need updates to `fire` and `test_optimizers.py` but otherwise looks good so far

PR #12, opened 2025-03-05
More robust tests

abhijeetgangan:
This PR introduces new and more robust tests for integrators ~~and optimizers.~~


abhijeetgangan:
Optimizer test is a bit tricky and needs to be compared with the same calculator in ASE. Will do another PR for it.

PR #31, opened 2025-03-06
Optimizer tests

abhijeetgangan:


tests/test_integrators.py
```diff
@@ -8,22 +8,23 @@
...
+from torchsim.units import MetalUnits
 
 
 @pytest.fixture
-def si_batched_base_state(si_base_state: BaseState) -> BaseState:
-    """Create a basic state from si_structure."""
-    batch = torch.repeat_interleave(torch.arange(2), si_base_state.positions.shape[0])
+def ar_double_base_state(ar_base_state: BaseState) -> BaseState:
+    """Create a basic state from ar_fcc_base_state."""
+    batch = torch.repeat_interleave(torch.arange(2), ar_base_state.positions.shape[0])
```
orionarcher: This name is already used in conftest, it shouldn't be redefined here

tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def si_double_base_state(si_atoms: Atoms, device: torch.device) -> Any:
 
 
 @pytest.fixture
-def ar_fcc_base_state(device: torch.device) -> BaseState:
+def ar_base_state(device: torch.device) -> BaseState:
```
orionarcher, line 66: Could we modify this function to use `ase` `bulk` instead of constructing lattice manually? non-blocking

tests/test_runners.py
```diff
@@ -124,19 +124,19 @@ def test_integrate_double_nvt_with_reporter(
...
+        temperature=100.0,  # K
         timestep=0.001,  # ps
         unit_system=UnitSystem.metal,
         trajectory_reporter=reporter,
         gamma=0.1,  # ps^-1
     )
 
     assert isinstance(final_state, BaseState)
-    assert final_state.n_atoms == 16
+    assert final_state.n_atoms == 512
```
orionarcher: Consider making the `ar_base_state` smaller in general, maybe 2x2x2 supercells instead of 4x4x4. A larger system will just make the tests a bit slower with little upside.

PR #37, opened 2025-03-07
Fix examples with errors

### LINKED ISSUE #33: Fix and test all example scripts
remove the filter for 9 example scripts not being tested in CI due to failures [e322123](https://github.com/Radical-AI/torch-sim/pull/6/commits/e3221233889fbe5fafc15c93cfe35830425bb62a)

https://github.com/Radical-AI/torch-sim/blob/e3221233889fbe5fafc15c93cfe35830425bb62a/.github/workflows/test.yml#L139-L151

also reduce MD step count in `examples/3_Dynamics/3.6_MACE_NVT_Nose_Hoover_temp_profile.py` and `examples/3_Dynamics/3.8_MACE_NPT_Nose_Hoover.py` which run for well over 20 min

abhijeetgangan:
closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,35 +184,35 @@ for filename in trajectory_files:
 
 TorchSim is built around the following core modules:
 
-- `torchsim.integrators`: Provides batched molecular dynamics integrators for simulating the time evolution of molecular systems.
+- [`torchsim.integrators`](torchsim/integrators.py): Provides batched molecular dynamics integrators for simulating the time evolution of molecular systems.
 
-- `torchsim.optimizers`: Provides optimization algorithms for molecular systems, including gradient descent and the FIRE algorithm.
+- [`torchsim.optimizers`](torchsim/optimizers.py): Provides optimization algorithms for molecular systems, including gradient descent and the FIRE algorithm.
 
-- `torchsim.unbatched_integrators`: Provides unbatched molecular dynamics integrators for simulating the time evolution of molecular systems.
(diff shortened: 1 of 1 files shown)


examples/1_Introduction/1.5_Fairchem.py
```diff
@@ -3,6 +3,7 @@
 # /// script
 # dependencies = [
 #     "fairchem-core>=1.6",
+#     "torch",
```
janosh: no need to add deps that are already listed in `pyproject.toml` since `uv run --with . ${{ matrix.example }}` will install them

examples/3_Dynamics/3.10_Hybrid_swap_mc.py
```diff
@@ -42,7 +42,7 @@
     model=loaded_model,
     device=device,
     dtype=dtype,
-    enable_cueq=True,
+    enable_cueq=False,
```
janosh, line 45: i would go with this such that people running the examples who have a GPU get the speedup
```suggestion
    enable_cueq=torch.cuda.is_available(),
```

examples/7_Others/7.1_Soft_sphere_autograd.py
```diff
@@ -3,6 +3,7 @@
 # /// script
 # dependencies = [
 #     "plotly>=6",
+#     "kaleido"
```
janosh: be sure to include a minimum version that the example is known to work with. i think you need to specify https://pypi.org/project/kaleido/1.0.0rc0 since the last non-rc release is 4 years old

PR #47, opened 2025-03-15
Structure directories

abhijeetgangan:


torch_sim/utils/math.py
janosh, line 1: i think this should just be `torch_sim/math.py`

torch_sim/utils/batching.py
janosh, line 1: how about just `torch_sim/batching.py`

torch_sim/utils/utils.py
janosh, line 1: how about `torch_sim/utils/utils.py`->`torch_sim/utils/__init__.py`

torch_sim/models/mattersim/m3gnet.py
janosh, line 1: what's the long-term plan for this module? you have a separate repo right? maybe makes sense to remove this folder and make the other repo an optional dep?

abhijeetgangan: Agreed. Will remove the redundant code in the next PR. I will add my new code later.

torch_sim/utils/math.py
abhijeetgangan: Should it be in utils and be a top level import?

janosh: i try to avoid `utils` in filenames unless i really don't know what else to name it. in this case, i don't think having it in `utils` adds anything

abhijeetgangan: I see. For now it can stay in `utils` but I agree it's better to avoid give such names. This is something worth discussing.

PR #48, opened 2025-03-15
Remove unused code and format examples to new API

abhijeetgangan:
Removes legacy / unused code.

Adapt the examples to new API.

Mattersim model will be added later as an optional dependency.


examples/6_Phonons/6.1_Phonon_dos_batched_MACE.py
```diff
@@ -17,6 +17,72 @@
 
 from torch_sim.models.mace import MaceModel
 from torch_sim.neighbors import vesin_nl_ts
+from torch_sim.runners import BaseState
+
+
+def phonopy_atoms_to_state(
```
orionarcher: maybe move to torchsim.runners and/or write a test?

Could consider renaming to `phonopy_to_state`

Also, would be nice to have `state_to_phonopy_atoms` equivalent.

PR #76, opened 2025-03-25
Batchable `static()` runner

janosh:
- support for reporters without file I/O and optional file writing in reporter.report()
- fix trajectory files left open warnings in tests
- more modular trajectory writing, now used in static runner
- more reporting docs
- `static` runner tests


orionarcher:
Seems to be all good, would you take a look @janosh? Feel free to push any changes directly

PR #112, opened 2025-04-05
Add MatterSim and fix ASE cell convention issues.

CompRhys:
* adds MatterSim (not all torch implementation as the neighbor list and triplet methods differ in strange ways).
* Reverts to use original SevenNet neighbor list based on covariate discussion with @AdeeshKolluru (c.f. mattersim neighbor list comment).
* defaults the neighbor list used for unbatched mace to match batched mace, update examples accordingly.
* parameterizes the contributed model tests adding structures with non-cubic cells. Fix column/row issues for some contributed models.
* parameterizes the elastic tests.
* Adds round trip tests for io models fixing column/row vector issues for converting into SimState.
* Adds properties for column and row vector cells, makes minimal changes to unitcell/frechet methods implementations due to column/row vector issues.


torch_sim/models/orb.py
```diff
@@ -382,7 +385,7 @@ def forward(self, state: SimState | StateDict) -> dict[str, torch.Tensor]:
             state = state.to(self._device)
 
         half_supercell = (
-            torch.max(state.volume) > 1000
+            torch.max(torch.det(state.cell)) > 1000
```
CompRhys, line 388: this is because volume isn't defined for PBC=False

torch_sim/optimizers.py
```diff
@@ -405,11 +404,13 @@ def gd_step(
...
         cell_update = cell_positions_new / cell_factor_expanded
-        new_cell = torch.bmm(state.reference_cell, cell_update.transpose(1, 2))
+        new_row_vector_cell = torch.bmm(
+            state.reference_row_vector_cell, cell_update.transpose(-2, -1)
+        )
 
         # Update state
         state.positions = atomic_positions_new
-        state.cell = new_cell
+        state.row_vector_cell = new_row_vector_cell
```
CompRhys, line 413: do we not need to update the orig/reference_cell each step?

abhijeetgangan: Nope. Just need to make sure both the updated cell and reference cell (when initialized) have the same conventions.

tests/models/test_mace.py
```diff
@@ -72,148 +41,82 @@ def ase_mace_calculator() -> MACECalculator:
...
+        "cu_sim_state",
+        "mg_sim_state",
+        "sb_sim_state",
+        "tio2_sim_state",
+        "ga_sim_state",
+        "niti_sim_state",
+        "ti_sim_state",
+        "si_sim_state",
+        "sio2_sim_state",
+        "benzene_sim_state",
```
abhijeetgangan: We need to be a bit careful about using models for non-pbc systems which were only trained on periodic systems.

torch_sim/optimizers.py
CompRhys: but if i'm not following the implementation incorrectly then the gradient matrix then will depend on a historic state? and so if I take a midpoint of a trajectory and reinit won't they diverge? A X = B has different solutions if the reference A changes?

abhijeetgangan: I see your point. However, in the algorithm the updated cell at any given optimization step is defined as the deformation gradient times the original cell.

Yes, the trajectories will diverge but should converge to the same minimum as the direction would be the same (The trajectories would diverge anyway because the models introduce randomness at each step).

To your point of re-initializing the optimizer state changing the trajectory, it is worth doing some tests and also checking with the ASE / Original implementation as they would do the same.

Do you think it is worth adding some sort of warning for now?

torch_sim/models/mattersim.py
```diff
@@ -0,0 +1,150 @@
...
+            threebody_cutoff=self.three_body_cutoff,
+        )
+
+        self.implemented_properties = [
+            "energy",
+            "forces",
+            "stress",
+        ]
+
+    def forward(self, state: SimState | StateDict) -> dict[str, torch.Tensor]:
```
abhijeetgangan, line 108: Looks like you are using the graph convertor from mattersim code which uses pymatgen + cython code. This is fine as they currently don't match with the neighbor list implementations we have. I would check if there is some performance drop compared to ASE for single structure and a large batch.

I am working porting the functionality to torch but I am not sure if it will be faster.

tests/conftest.py
```diff
@@ -173,3 +351,157 @@ def torchsim_trajectory(si_sim_state: SimState, lj_model: Any, tmp_path: Path):
...
+        )
+
+    # Rename the function to include the test name
+    test_unbatched_model_calculator_consistency.__name__ = (
+        f"test_unbatched_{test_name}_consistency"
+    )
+    return test_unbatched_model_calculator_consistency
+
+
+consistency_test_simstate_fixtures = [
```
abhijeetgangan: Should we include some randomly perturbed structures? Most of them have close to zero forces.

torch_sim/models/sevennet.py
```diff
@@ -226,19 +222,16 @@ def forward(self, state: SimState | StateDict) -> dict[str, torch.Tensor]:
...
         else:
             results["energy"] = torch.zeros(self.n_systems, device=self._device)
 
-        # Process forces
         forces = output[key.PRED_FORCE]
         if forces is not None:
             results["forces"] = forces.detach()
 
-        # Process stress
         stress = output[key.PRED_STRESS]
```
abhijeetgangan, line 239: Sevennet always returns forces and stress unlike other models where you can pass a bool to compute them. Is there a way make it similar to other models?

torch_sim/models/sevennet.py
```diff
@@ -24,6 +22,7 @@
     import torch
     from sevenn.atom_graph_data import AtomGraphData
     from sevenn.calculator import torch_script_type
+    from sevenn.train.dataload import _graph_build_f
```
abhijeetgangan: Why are we using the ASE / matscipy neighbor list? The neighborlist implementations we have are already tested with ASE implementation.

torch_sim/models/orb.py
orionarcher: This also won't be defined for `PBC=False` because `state.cell` will be `None`.

What about:

```python
        half_supercell = (
            torch.max(state.volume) > 1000
            if state.pbc and self._half_supercell is None
            else self._half_supercell
        )
```

torch_sim/models/mattersim.py
CompRhys: https://github.com/luthaf/vesin benchmarking here shows that the pmg cython code is faster than ase. I also tried to port to torch but to match the logic involves lots of loops and so then it will definitely be slower.

torch_sim/models/sevennet.py
CompRhys: I originally used the vesin but then the issues with mattersim made me question whether that might be a source of uncontrolled error.

abhijeetgangan: I haven't compared with the pmg cython code or triplet code used in mattersim but for 7net we can safely use vesin because https://github.com/MDIL-SNU/SevenNet/blob/8ce2c9d4e2aa6e5fe19bcf0882cbaab82a6ec0e8/sevenn/train/dataload.py#L82

CompRhys: looking further I agree that the neighbor lists all match ase, as a result I think all classical models may also need updating as they're passing the `column_vector_cell` as `cell` rather than the `row_vector_cell`. I swapped sevennet back to vesin

torch_sim/models/sevennet.py
CompRhys: https://github.com/MDIL-SNU/SevenNet/blob/8ce2c9d4e2aa6e5fe19bcf0882cbaab82a6ec0e8/sevenn/nn/force_output.py They don't seem to have a way to turn it off. @YutackPark is this a feature you could add in a backwards compatible manner?

tests/models/test_mace.py
CompRhys: I agree for science but for testing I think it's good to test in principal that they can handle molecules.

tests/conftest.py
CompRhys: added a rattled sio2 structure.

torch_sim/models/sevennet.py
abhijeetgangan: I had a chat with @YutackPark and I think it is better to return all the values for now in the results instead of changing the 7net source code.

torch_sim/optimizers.py
CompRhys: Checking the ase implementation they also do not update the original/reference cell. This seems wrong to me but happy to mark resolved and return in the future. Maybe @lan496 could shed some thoughts based on his implementation? Did you choose not to update the reference cell just because that was how it was done for the UnitCellFilter you were improving the numerical stability of?

torch_sim/models/sevennet.py
YutackPark: I did it intentionally so that 7net can compute a batch consisting of both non-PBC and PBC structures. 

It is not hard to add a switch. I can simply replace force stress head to force only head.

torch_sim/models/sevennet.py
CompRhys: https://github.com/Radical-AI/torch-sim/tree/fix-classical-row-column

torch_sim/optimizers.py
lan496: I think using the original cell as a reference is fine from an algorithm perspective because it is only used to calculate the strain tensor. Of course, the frequency of updating the reference cell may change an optimization efficiency or robustness; for example, LAMMPS's [fix nph](https://docs.lammps.org/fix_nh.html#fix-npt-command) has an `nreset` flag to specify an interval of resetting the reference cell

abhijeetgangan: I will take a note of this and revisit this later. Resetting the original cell with a given interval is one option we can add.

PR #115, opened 2025-04-06
Feature: Correlation function module

stefanbringuier:
## Summary

* Adds module for correlation functions based on FFT method.
* Utilizes `CircularBuffer` for optimal time series data storage.
* `CorrelationCalculator` provides on-the-fly autocorrelation and cross-correlation.
* Added VACF (Velocity Autocorrelation Function) example script.

### Utility
Enables use of fluctuation-dissipation approach for extracting physical quantities like diffusion coefficients and vibrational density of states.

## Checklist

* [x] Doc strings have been added in the [Numpy docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html). (I think :shrug:)


torch_sim/correlations.py
abhijeetgangan, line 1: Can this be moved to `quantities.py`?

abhijeetgangan:
Thanks for the contribution! Looks good with the example and tests. Can you sign the CLA and address the comment above before I merge?

stefanbringuier:
> Thanks for the contribution! Looks good with the example and tests. Can you sign the CLA and address the comment above before I merge?

Signed individual CLA, don't see any comments to address though?

abhijeetgangan (changes requested):

examples/scripts/7_Others/7.4_Velocity_AutoCorrelation.py
```diff
@@ -0,0 +1,184 @@
...
+# dependencies = [
+#     "ase>=3.24",
+#     "matplotlib",
+#     "numpy",
+# ]
+# ///
+
+__status__ = "Development"
+__author__ = "Stefan Bringuier"
+__email__ = "stefan.bringuier@gmail.com"
```
orionarcher: We intend to use git blame for attribution tracking instead of `__<metadata>__`, would you mind removing these?

torch_sim/correlations.py
```diff
@@ -0,0 +1,399 @@
...
+References:
+    .. [1] D. Frenkel and B. Smit, "Understanding molecular simulation: From
+       algorithms to applications", Academic Press, 2002.
+    .. [2] `pwtools: Phonon DOS <https://elcorto.github.io/pwtools/written/background/phonon_dos.html>`_
+
+"""
+
+__status__ = "Unit tested"
+__author__ = "Stefan Bringuier"
+__email__ = "stefan.bringuier@gmail.com"
```
orionarcher: Same comment as script here

torch_sim/correlations.py
orionarcher: Discussed this briefly with @abhijeetgangan, could we move this to `torch_sim/properties/correlations.py` instead?

We'll be moving other functionality there too, `elastic.py` and upcoming `phonons.py`, so `correlations.py` would be a great start.

tests/test_correlations.py
```diff
@@ -0,0 +1,388 @@
...
+    def __init__(self, velocities: torch.Tensor, device: torch.device) -> None:
+        """Initialize mock state with provided data."""
+        self.velocities = velocities
+        self.device = device
+
+
+@pytest.fixture
+def device() -> torch.device:
+    """Fixture for computation device."""
+    return torch.device("cpu")
```
orionarcher: could you use the device fixture from `conftest.py`?

torch_sim/correlations.py
```diff
@@ -0,0 +1,399 @@
...
+        *,
+        normalize: bool = True,
+    ) -> None:
+        """Initialize a correlation calculator.
+
+        Args:
+            window_size: Number of steps to keep in memory
+            delta_t: step between correlation calculations
+            quantities: Dictionary mapping names to functions that calculate
+                       quantities from a SimState
```
orionarcher: I would suggest calling this `properties` to keep consistency with the `prop_calculators` in the TorchSimTrajectory. Open to discussion.

torch_sim/correlations.py
```diff
@@ -0,0 +1,399 @@
...
+        correlations: Current correlation results
+        device: Device where calculations are performed
+    """
+
+    def __init__(
+        self,
+        window_size: int,
+        delta_t: int = 1,
+        quantities: dict[str, Callable[[SimState], torch.Tensor]] | None = None,
+        device: torch.device | None = None,
```
orionarcher: i would suggest moving *, after self and make these all mandatory kwargs.

torch_sim/correlations.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def update(self, state: SimState, step: int) -> None:
+        """Update correlation calculations with new state data.
+
+        Args:
+            state: Current simulation state
+            step: Current simulation step
+        """
+        if step % self.delta_t != 0:
+            return
```
orionarcher: Consider removing self.delta_t as a parameter and allow the user to instead decide delta_t by calling `update` more or less frequently.

Imagine I set `delta_t=12` and then in `prop_calculators` I set the frequency to 5. Then my effective delta_t is 60 because the timesteps don't line up often. Bit of an edge case but potentially very confusing behavior for a user.

examples/scripts/7_Others/7.4_Velocity_AutoCorrelation.py
```diff
@@ -0,0 +1,184 @@
...
+            elif use_running_average:
+                # Compute new running average
+                factor = 1.0 / window_count
+                diff = vacf_np - running_avg_vacf
+                running_avg_vacf = running_avg_vacf + diff * factor
+
+            last_window_vacf = vacf_np.copy()
+            corr_calc.reset()
+
+        return torch.tensor([window_count], device=device, dtype=dtype)
```
orionarcher: I quite like this demonstration of how to drop this into prop calculators, very nicely done.

You should consider moving the `calc_vacf` function directly into `correlations.py` so that users can import it and directly put it in prop_calculators. This would require making some reasonable choices/assumptions for the buffer size and other parameters but I think it'd make for a nice contribution.

orionarcher (changes requested):
Thank you @stefanbringuier! I have several comments but this is a very nice contribution and we should be able to merge soon.

torch_sim/correlations.py
stefanbringuier: Hmm, good point, thanks for the thoughtful feedback! Will think about it and probably get to it tomorrow night.

examples/scripts/7_Others/7.4_Velocity_AutoCorrelation.py
stefanbringuier: Yeah makes sense, was thinking something like:

```python
correlation_dt = 10
vacf_calc = VelocityAutoCorrelation(
    window_size=150,
    correlation_dt=correlation_dt,
    device=device
)

reporter = ts.TrajectoryReporter(
    trajectory,
    prop_calculators={correlation_dt: {"vacf": vacf_calc}}
)
```

Thoughts?

orionarcher: Other than the potential consolidation/removal of `correlation_dt`, I think this looks good!

torch_sim/correlations.py
stefanbringuier: Completed, better design, thanks again.

examples/scripts/7_Others/7.4_Velocity_AutoCorrelation.py
stefanbringuier: Added this class/docs/tests. Keep the example script for now.

orionarcher: Definitely keep the script but I'd suggest replacing the redefinition of call_vacf with the `VelocityAutoCorrelation` class directly. It'd be nice to have a usage example in the codebase.

orionarcher:
Just updated branch to get a docs build test in here too. This good to merge when tests are passing @stefanbringuier?

stefanbringuier:
Yep, looks good once tests/docs pass, thanks for the feedback and creating torch-sim!

orionarcher:
Merging, test will be fixed upstream.

PR #118, opened 2025-04-08
add support for `graph-pes` models

jla-gardner:
Hi there 👋 thanks for the awesome work with `torch-sim` 😊

## Summary

I wrote and maintain [`graph-pes`](https://github.com/jla-gardner/graph-pes), a package for creating and training models of the potential energy surface that act on graph representations of atomic structures. It would be amazing if my users could make use of `torch-sim` to run fast MD in a hassle free manner! (I currently have LAMMPS and ASE calculators which are annoying to set up and ~slow respectively). 

This PR adds support for using arbitrary `GraphPESModel`s from the [graph-pes](https://github.com/jla-gardner/graph-pes) package within `torch-sim`.

I've also added a small tutorial notebook to give a concrete example of how to use the `GraphPESWrapper` class in the docs.

## Checklist

* [x]  Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.


examples/tutorials/using_graphpes_tutorial.py
```diff
@@ -0,0 +1,89 @@
...
+#     "graph-pes>=0.0.30",
+#     "torch==2.5",
+# ]
+# ///
+# </details>
+
+
+# %% [markdown]
+"""
+# Using GraphPES
```
orionarcher, line 15: Could you change the name to Integrating with GraphPES

.github/workflows/test.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           - { name: orb, test_path: "tests/models/test_orb.py" }
           - { name: sevenn, test_path: "tests/models/test_sevennet.py" }
           - { name: fairchem, test_path: "tests/models/test_fairchem.py" }
+          - { name: graphpes, test_path: "tests/models/test_graphpes.py" }
```
orionarcher, line 69: The model tests are failing because you need to modify the pyproject.toml to include a `graph_pes` optional dependency, see line 104 of `test.yml` to understand why.

examples/tutorials/using_graphpes_tutorial.py
orionarcher: Could you change the name to "Integrating with GraphPES"?

tests/models/test_graphpes.py
```diff
@@ -0,0 +1,114 @@
...
+    ts_output = ts_model(atoms_to_state(systems, device, torch.float32))
+
+    assert set(ts_output.keys()) == {"energy", "forces", "stress"}
+    assert ts_output["energy"].shape == (2,)
+    assert ts_output["forces"].shape == (sum(len(s) for s in systems), 3)
+    assert ts_output["stress"].shape == (2, 3, 3)
+
+    assert gp_energies[0].item() == pytest.approx(ts_output["energy"][0].item(), abs=1e-5)
+
+
```
orionarcher, line 111: We just added a more comprehensive set of validation tests for new `ModelInterfaces` to check results against an ASE calculator on a variety of systems and validate that the model outputs have the correct form. These are `tests/models/conftest.py::make_model_calculator_consistency_test` and `tests/models/conftest.py::make_validate_model_outputs_test`. Could you implement these for GraphPES with either the mattersim or MACE models?


See `tests/models/test_sevennet.py` for an example

```python
test_sevennet_consistency = make_model_calculator_consistency_test(
    test_name="sevennet",
    model_fixture_name="sevenn_model",
    calculator_fixture_name="sevenn_calculator",
    sim_state_names=consistency_test_simstate_fixtures,
)


test_sevennet_model_outputs = make_validate_model_outputs_test(
    model_fixture_name="sevenn_model",
)
```

This gets a bit tricky because you support so many models. I'd be happy running these tests with just 1 or 2 models. I'll also admit this "interface to an interface" use case wasn't one I'd considered, so it's ok if we need to shift our testing framework a bit to accommodate it.

torch_sim/models/graphpes.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Returns:
+            Dictionary containing the computed energies, forces, and stresses
+            (where applicable)
+        """
+        if not isinstance(state, SimState):
+            state = SimState(**state)  # type: ignore[arg-type]
+
+        atomic_graph = state_to_atomic_graph(state, self._gp_model.cutoff)
+        return self._gp_model.predict(atomic_graph, self._properties)  # type: ignore[return-value]
```
orionarcher, line 185: Just to confirm, this will then have a batched operation on the GraphPES side?

I'm a little worried about shape mismatches between the GraphPES and TorchSim conventions but the tests that I mentioned above should catch all of those.

torch_sim/models/graphpes.py
```diff
@@ -0,0 +1,166 @@
...
+        self._gp_model = _model.to(device=self._device, dtype=self._dtype)
+
+        self._compute_forces = compute_forces
+        self._compute_stress = compute_stress
+
+        self._properties: list[PropertyKey] = ["energy"]
+        if self._compute_forces:
+            self._properties.append("forces")
+        if self._compute_stress:
+            self._properties.append("stress")
```
orionarcher, line 166: Consider setting a dynamic `memory_scales_with` @propery. The default `memory_scales_with` from ModelInterface will return `n_atoms_x_density`, which is only correct for models with radial cutoffs. I am not sure if all of the models you support have radial cutoffs. If not, the memory scaler will be incorrect and the `autobatching` will subtly break. See [here](https://radical-ai.github.io/torch-sim/tutorials/autobatching_tutorial.html#Understanding-Memory-Requirements) for a more detailed discussion. I suggest adding a `memory_scales_with` property that will dynamically return the correct value depending on what model is being used.

orionarcher (changes requested):
Hi @jla-gardner, thanks for the PR! I would be very happy to have this interface in TorchSim.

For added context, I want to point you to our [current posture on model interfaces](https://github.com/Radical-AI/torch-sim/discussions/120), just to share how we are thinking about supporting models over time.

I've included a few comments to help get the tests working and make sure that the code correctly interfaces with TorchSim.

.github/workflows/test.yml
orionarcher: Edit, nevermind, I see that you have it now. Let me take a closer look.

torch_sim/models/graphpes.py
```diff
@@ -0,0 +1,166 @@
...
+
+    class GraphPESWrapper(torch.nn.Module, ModelInterface):  # type: ignore[reportRedeclaration]
+        """GraphPESModel wrapper for torch_sim.
+
+        This class is a placeholder for the GraphPESWrapper class.
+        It raises an ImportError if graph_pes is not installed.
+        """
+
+        def __init__(self, *_args: typing.Any, **_kwargs: typing.Any) -> None:  # noqa: D107
+            raise ImportError("graph_pes must be installed to use this model.")
```
orionarcher, line 42: You need to add a dummy `AtomicGraph` here because it's used in a type hint, that's why the tests are failing.

jla-gardner: Just came to the same conclusion myself 👍  thanks for the spot 😄

jla-gardner:
> Hi @jla-gardner, thanks for the PR! I would be very happy to have this interface in TorchSim.
> 
> For added context, I want to point you to our [current posture on model interfaces](https://github.com/Radical-AI/torch-sim/discussions/120), just to share how we are thinking about supporting models over time.
> 
> I've included a few comments to help get the tests working and make sure that the code correctly interfaces with TorchSim.

Thanks for the extremely quick response! Yes this is slightly a-typical as being an interface to an interface. Going through all your comments point by point now.

torch_sim/models/graphpes.py
jla-gardner: Yes, this is correct! Outputs of the `GraphPESWrapper` _should_ be of shape:  `(S,)` for energy, `(N, 3)` for forces and `(S, 3, 3)` for stresses, for a batch of S structures containing a total of N atoms.

I'll add your new tests to check that this is the case.

torch_sim/models/graphpes.py
jla-gardner: Good spot - yes _almost_ all graph-pes models have radial cutoffs. I'll update the `memory_scales_with` property to handle the edge cases where this isn't true.

torch_sim/models/graphpes.py
```diff
@@ -0,0 +1,166 @@
...
+    Returns:
+        AtomicGraph object representing the batched structures
+    """
+    graphs = []
+
+    for i in range(state.n_batches):
+        batch_mask = state.batch == i
+        R = state.positions[batch_mask]
+        Z = state.atomic_numbers[batch_mask]
+        cell = state.cell[i]
```
CompRhys:
```suggestion
        cell = state.row_vector_cell[i]
```

Our cell convention is different to ASE/vesin so we actually will need the transpose here.

jla-gardner: Thanks! Updated.

tests/models/test_graphpes.py
jla-gardner: I've implemented this for a simple `LennardJones` model for now. Adding `MACE` test shortly

examples/tutorials/using_graphpes_tutorial.py
jla-gardner: Updated

tests/models/test_graphpes.py
```diff
@@ -112,3 +117,35 @@ def test_graphpes_dtype(device: torch.device, dtype: torch.dtype):
...
+    return _lj_model.ase_calculator()
+
+
+test_graphpes_lj_consistency = make_model_calculator_consistency_test(
+    test_name="graphpes-lj",
+    model_fixture_name="ts_lj_model",
+    calculator_fixture_name="ase_lj_calculator",
+    sim_state_names=consistency_test_simstate_fixtures,
+    rtol=1e-3,
+    atol=1e-3,
```
CompRhys: These are pretty high tolerances 100x what we want to see by default, do they need to be this loose?

jla-gardner: Fixed 👍

tests/models/test_graphpes.py
jla-gardner: `MACE` and `NequIP` added 😃

jla-gardner:
Hi guys - I'm looking into the test failures from the `graph-pes` side. I note that you've had to use a high `rtol` and `atol` in your dedicated `test_mace.py` file, so it is perhaps unexpected that I am also running into some inconsistency errors. 

I'm seeing that the tests are only failing for the `rattled_sio2_sim_state`: from what I can see, this rattles atoms significantly!  In particular, one of the oxygen atoms is moved by 1.8Å, and ends up in close proximity to a Si - my hypothesis is that this is therefore a very high energy + high force conformation, and so small precision effects can lead to a large absolute (and relative) differences. Perhaps less extreme rattling here would make this test also pass for your dedicated `mace` test.

I've pushed a change to that effect now as a test

CompRhys:
> Hi guys - I'm looking into the test failures from the `graph-pes` side. I note that you've had to use a high `rtol` and `atol` in your dedicated `test_mace.py` file, so it is perhaps unexpected that I am also running into some inconsistency errors.
> 
> I'm seeing that the tests are only failing for the `rattled_sio2_sim_state`: from what I can see, this rattles atoms significantly! In particular, one of the oxygen atoms is moved by 1.8Å, and ends up in close proximity to a Si - my hypothesis is that this is therefore a very high energy + high force conformation, and so small precision effects can lead to a large absolute (and relative) differences. Perhaps less extreme rattling here would make this test also pass for your dedicated `mace` test.
> 
> I've pushed a change to that effect now as a test

Yes I did reduce the MACE tolerances due to the SiO2 rattling test breaking. It was on my TODO list to dig a little bit deeper on that incase it actually is a real difference vs numerical non-determinism that might be exaggerated by the MACE architecture. It's worth noting that I do not see that rattling test breaking locally when running on M3 macbook.

abhijeetgangan:
I did some tests locally and it seems that the `small` model predicts very high energy and large forces. If you switch to a better model like `medium-mpa-0` or `medium-omat-0` it will pass the test.

jla-gardner:
All tests pass locally on my MacBook M2 too for what it's worth. Do you have a preference between reducing the rattling amplitude and using a larger model (as suggested by @abhijeetgangan!) for testing to keep this test passing?

abhijeetgangan:
If the larger models don't have a big overhead in testing then I would probably use those. @CompRhys What do you think?

CompRhys:
> If the larger models don't have a big overhead in testing then I would probably use those. @CompRhys What do you think?

> All tests pass locally on my MacBook M2 too for what it's worth. Do you have a preference between reducing the rattling amplitude and using a larger model (as suggested by @abhijeetgangan!) for testing to keep this test passing?

biggest MACE still smaller than smallest Fairchem so I would suspect it wouldn't add too much time to the tests. I would be happy to swap to testing `medium-mpa-0`. ASL license means that we probably want to avoid `medium-omat-0` to make sure there's no potential to misconstrue torch-sim as a commercial use and create a headache.

abhijeetgangan:
> > If the larger models don't have a big overhead in testing then I would probably use those. @CompRhys What do you think?
> 
> > All tests pass locally on my MacBook M2 too for what it's worth. Do you have a preference between reducing the rattling amplitude and using a larger model (as suggested by @abhijeetgangan!) for testing to keep this test passing?
> 
> biggest MACE still smaller than smallest Fairchem so I would suspect it wouldn't add too much time to the tests. I would be happy to swap to testing `medium-mpa-0`. ASL license means that we probably want to avoid `medium-omat-0` to make sure there's no potential to misconstrue torch-sim as a commercial use.

`medium-mpa-0` should be good enough in that case.

jla-gardner:
updated to use `medium-mpa-0`

tests/models/test_graphpes.py
```diff
@@ -0,0 +1,209 @@
...
+    ts_output = ts_model(atoms_to_state([bulk_atoms], device, torch.float32))
+    assert set(ts_output.keys()) == {"energy", "forces", "stress"}
+    assert ts_output["energy"].shape == (1,)
+    assert ts_output["forces"].shape == (len(bulk_atoms), 3)
+    assert ts_output["stress"].shape == (1, 3, 3)
+
+    torch.testing.assert_close(
+        ts_output["forces"].to("cpu"),
+        gp_forces,
+        rtol=1e-3,
```
abhijeetgangan: Does this test need this high tolerance?

jla-gardner:
Looking into these continuing errors on the `graph-pes` side tonight and tomorrow morning to see if I can ascertain their root cause and reproduce locally - I'm on UK time so I'm hoping to have fixed these issues by tomorrow. Sorry for the delays, and thanks for all your help so far with this everyone 😄

tests/models/test_graphpes.py
jla-gardner: Nope! Updated

jla-gardner:
Hi guys, after some investigating I have found that `MACE` models are susceptible to surprisingly high variance when one changes the order of the entries in the neighbour list (~1e-6 absolute difference in forces on my MacBook M2, so perhaps larger on the `ubuntu-latest` runners which seem to be the ones most often failing above)

I have ensured that the `graph-pes` ASE calculator and `GraphPESWrapper` instances now use the same ordering, and am hoping that this resolves the issue in the CI 🤞

tests/models/test_graphpes.py
```diff
@@ -0,0 +1,207 @@
...
+    return mace_mp("medium-mpa-0").to(device, dtype).ase_calculator(skin=0.0)
+
+
+test_graphpes_mace_consistency = make_model_calculator_consistency_test(
+    test_name="graphpes-mace",
+    model_fixture_name="ts_mace_model",
+    calculator_fixture_name="ase_mace_calculator",
+    sim_state_names=consistency_test_simstate_fixtures,
+    # see test_mace.py for similar issue
+    rtol=6e-4,  # FIXME: unclear why this needs to be so high for mace.  # noqa: FIX001
```
orionarcher: Could you take out the comment if this has been resolved or add a note explaining your findings?

orionarcher (approved):
Looks good to me (one tiny nit), thanks for the quick edits @jla-gardner!

I pushed a small change to `docs.yml` that will test the docs build in the branch. That should be in main anyway and I'd like to check it before merging.

orionarcher:
One more thing, could you modify `docs/conf.py` to include the model in `autodoc_mock_imports`? Per step 5 in [Adding New Models](https://radical-ai.github.io/torch-sim/dev/add_model.html)

tests/models/test_graphpes.py
jla-gardner: Tentatively removed the high `rtol` requirements in the latest commit to see whether this is actually completely resolved. I fear that it won't be, as `graph-pes` is just passing data straight through to `mace-torch`, and so if the base MACE model requires these high `rtol`s in `test_mace.py`, we probably need them here too.

jla-gardner: Okay, that didn't work. Reverting now and adding comments to this effect

orionarcher:
To get the docs build to work you'll need to add `docs.yml` file to include a graph-pes install. This isn't the most scalable approach but it works fine for now, I can loop back in a future PR if it gets annoying.

jla-gardner:
> To get the docs build to work you'll need to add `docs.yml` file to include a graph-pes install. This isn't the most scalable approach but it works fine for now, I can loop back in a future PR if it gets annoying.

Added this but have now managed to break the docs build further down the track with the cryptic error `"Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x7ff7fd71a980> for event 'builder-inited' threw an exception (exception: no module named torch_sim)"`

Clearly my graph-pes install has made the docs fall over somewhere - trying to reproduce locally.

orionarcher:
Hmmm, very strange. @janosh has encountered that error when trying to build the docs locally, too. He might have input. I'll take a look too.

Ignore the 5.3 failure, thats intermittent.

jla-gardner:
> Hmmm, very strange. @janosh has encountered that error when trying to build the docs locally, too. He might have input. I'll take a look too.

I've reproduced locally - I think there is a subtle incompatibility between latest versions of graph-pes and mace-torch. 
This leads to an error that gets caught during doc-building, but the actual error itself gets hidden somewhere up the stack trace.
I'm fixing this incompatibility over on the graph-pes repo, and adding more tests to catch for this - thanks for the rigorous tests here that have led me to find this!

Will update with a new minimum requirement for `graph-pes` in due course, and that should fix these errors 😄

jla-gardner:
> > Hmmm, very strange. @janosh has encountered that error when trying to build the docs locally, too. He might have input. I'll take a look too.
> 
> I've reproduced locally - I think there is a subtle incompatibility between latest versions of graph-pes and mace-torch. This leads to an error that gets caught during doc-building, but the actual error itself gets hidden somewhere up the stack trace. I'm fixing this incompatibility over on the graph-pes repo, and adding more tests to catch for this - thanks for the rigorous tests here that have led me to find this!
> 
> Will update with a new minimum requirement for `graph-pes` in due course, and that should fix these errors 😄

Updated. docs are now building locally with the newly pinned graph-pes version.

orionarcher:
Thanks for the contribution @jla-gardner. Good to merge?

jla-gardner:
> Thanks for the contribution @jla-gardner. Good to merge?

Absolutely 😄 thanks once again for the sterling work with `TorchSim`!

PR #124, opened 2025-04-09
Add metatensor models

frostedoyster:
Hi everyone and thanks for the great work on the library. Also, let us know if you have feedback on https://github.com/Luthaf/vesin!

## Summary

* This PR adds an interface to the `metatensor` family of models (see https://docs.metatensor.org/latest/atomistic/index.html)

## Checklist


.github/workflows/test.yml
```diff
@@ -66,6 +67,8 @@ jobs:
           - { name: mattersim, test_path: "tests/models/test_mattersim.py" }
           - { name: orb, test_path: "tests/models/test_orb.py" }
           - { name: sevenn, test_path: "tests/models/test_sevennet.py" }
+          - { name: fairchem, test_path: "tests/models/test_fairchem.py" }
+          - { name: metatensor, test_path: "tests/models/test_metatensor.py" }
```
CompRhys: Can we make these alphabetical? you've duplicated the fairchem tests so maybe merge main then do a soft reset against main and then force pushing might be a useful thing to do just to check that there are no other divergence issues.

docs/conf.py
```diff
@@ -62,7 +62,7 @@
     "html_image",
 ]
 
-autodoc_mock_imports = ["fairchem", "mace", "mattersim", "orb", "sevennet"]
+autodoc_mock_imports = ["fairchem", "mace", "mattersim", "orb", "sevennet", "metatensor"]
```
CompRhys: again alphabetical please :)

tests/models/test_metatensor.py
```diff
@@ -0,0 +1,103 @@
...
+    torch.testing.assert_close(
+        metatensor_results["energy"].item(),
+        calc_energy,
+        atol=1e-4,
+        rtol=0.5,
+    )
+    torch.testing.assert_close(
+        metatensor_results["forces"],
+        calc_forces,
+        atol=1e-4,
```
CompRhys: We've updated the tests to add more test cases, can you look at the model tests on main and copy the structure that's now there

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        # Extract required data from input
+        if isinstance(state, dict):
+            state = SimState(**state, masses=torch.ones_like(state["positions"]))
+
+        # Input validation is already done inside the forward method of the
+        # MetatensorAtomisticModel class, so we don't need to do it again here.
+
+        atomic_numbers = state.atomic_numbers
+        cell = state.cell
```
CompRhys: I think this likely needs to be updated to `state.row_vector_cell`, we use column vector convention similar to LAMMPS rather than rows like ASE.

CompRhys (changes requested):
Great start, I would be a little bit cautious about including the batched autograd code in this repo for forces and stresses as it seems as though it could differ for future models and "MetaTensor" seems as though it may be a collection of models in the future c.f. `fairchem` or `graphpes` and so it may be more maintainable if that logic lives in your repo? That wouldn't be a reason not to merge but worth some thought.

.github/workflows/test.yml
frostedoyster: Yes, of course

torch_sim/models/metatensor.py
frostedoyster: Thanks! Good catch

tests/models/test_metatensor.py
frostedoyster: Updated

frostedoyster:
> I would be a little bit cautious about including the batched autograd code in this repo for forces and stresses as it seems as though it could differ for future models and "MetaTensor"

This is a conscious design decision that was made in `metatensor`, of course with its advantages and disadvantages, but meaning that any model will conform to it. The upsides are that, while the model can calculate its own forces at training time, the computation of the forces is done outside of it at evaluation time, ensuring that the forces are computed as the derivative of the energies to prevent e.g. incorrect MD. I agree with you that ideally we would like to make a single function inside `metatensor` that computes forces and stresses from energies in this way, but it's also true that the same function would only contain two/three lines (an autograd call, a division by the volume, and perhaps a concatenation). This is why we haven't done it yet and generally prefer having the calling code do it instead. Of course, we will update this interface once this function is added, although this might take a while as we are going through a sizeable refactoring effort.

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+        """Initialize the metatensor model for energy, force and stress calculations.
+
+        Sets up a metatensor model for energy, force, and stress calculations within
+        the TorchSim framework. The model can be initialized with atomic numbers
+        and batch indices, or these can be provided during the forward pass.
+
+        Args:
+            model (str | Path | None): Path to the metatensor model file or a
+                pre-defined model name (currently only "pet-mad" is supported as a
+                pre-defined model). If None, defaults to "pet-mad".
```
Luthaf: Maybe add a link to the pet-mad paper/repo here?

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+            extensions_path (str | Path | None): Optional, path to the folder containing
+                compiled extensions for the model.
+            device (torch.device | None): Device on which to run the model. If None,
+                defaults to "cuda" if available, otherwise "cpu".
```
Luthaf, line 82: Should this default to what the model requests first instead? This way the model can pick what they prefer here.

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+        elif model.endswith(".ckpt"):
+            path = model
+            self._model = load_model(path).export()
+        elif model.endswith(".pt"):
+            path = model
+            self._model = load_atomistic_model(path, extensions_path)
+        else:
+            raise ValueError('Model must be a path to a .ckpt/.pt file, or "pet-mad".')
+
+        self._device = device or ("cuda" if torch.cuda.is_available() else "cpu")
```
Luthaf, line 118: this should check devices supported by the model

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+        self._dtype = getattr(torch, self._model.capabilities().dtype)
+        self._model.to(self._device)
```
Luthaf, line 129: I would also check that the model does declare an "energy" output

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+                pbc=system_pbc,
+            )
+
+            # Calculate neighbor list(s) for this system
+            for neighbor_list_options in self._requested_neighbor_lists:
+                if not neighbor_list_options.full_list:
+                    raise ValueError(
+                        "Neighbor list options must be full for metatensor models to "
+                        "be used in torchsim."
+                    )
```
Luthaf: Is this because `neighbor_list_fn` can only return full NL? I think we should still support half NL, because that's still a relevant use case for some model. I can help with an implementation.

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,312 @@
...
+                grad_outputs=torch.ones_like(model_outputs["energy"].block().values),
+            )
+        else:
+            derivatives = []
+
+        results_by_system: dict[str, list[torch.Tensor]] = {}
+        if self._compute_forces and self._compute_stress:
+            results_by_system["forces"] = [-d for d in derivatives[: len(systems)]]
+            results_by_system["stress"] = [
+                d / torch.abs(torch.det(system.cell.detach()))
```
Luthaf, line 262: does torchsim uses the negative of the usual (i.e. in LAMMPS) stress definition?

torch_sim/models/metatensor.py
Luthaf: I would prefer not to default to any specific model, since otherwise we can't update the code to use a better model down the line without changing behavior for users. I'm not sure what are the stability guarantees of torch-sim though, @CompRhys do you have guidelines about this?

pyproject.toml
```diff
@@ -46,11 +46,11 @@ test = [
 ]
 mace = ["mace-torch>=0.3.11"]
 mattersim = ["mattersim>=0.1.2"]
-
 orb = [
     "orb-models@git+https://github.com/orbital-materials/orb-models#egg=637a98d49cfb494e2491a457d9bbd28311fecf21",
 ]
 sevenn = ["sevenn>=0.11.0"]
+metatensor = ["metatrain>=2025.4"]
```
Luthaf:
```suggestion
metatensor = ["metatensor-torch >=0.7,<0.8", "metatrain>=2025.4"]
```

This way we can more easily update the code when the metatomic refactor is done, and the two dependencies of this code are explicit (instead of implicitly pulling metatensor-torch as a dependency of metatrain). Without the version constrain I could see a world where metatrain updates to new versions of metatensor-torch, but this code does not support it.

torch_sim/models/metatensor.py
frostedoyster: Yes, as far as I understand torch-sim has an interface for the NLs which doesn't support half NLs. Currently, the only backend for this interface seems to be vesin-torch. @CompRhys please correct me if I'm wrong. Anyway, a full NL should cover >90% of models in practice

torch_sim/models/metatensor.py
frostedoyster: This calculation should be correct (stress = energy_derivative / V), just like in our ASE calculator (whose outputs we check against in the tests)

torch_sim/models/metatensor.py
frostedoyster: I went with a default because all other models seemed to have a default. If this isn't necessary, I would agree that removing the default is a good idea

torch_sim/models/metatensor.py
Luthaf: There are two conventions for the stress used by different MD engines, depending if the stress is seen as "internal" or "external". ASE uses one of the definitions, LAMMPS the other.

torch_sim/models/metatensor.py
abhijeetgangan: Currently the default neighbor list is vesin-torch but you can change that. All of them are tested against ase neighborlist implement. You can turn off the full NL here https://github.com/Radical-AI/torch-sim/blob/c79fc6591802df53a5fd0869b96715826ef0d09e/torch_sim/neighbors.py#L547

torch_sim/models/metatensor.py
abhijeetgangan: Generally it would be the user to select the model and if not then there should be a sensible default. If you feel like having no default is better then that’s fine as long as it has a descriptive error message when no model is provided.

> I'm not sure what are the stability guarantees of torch-sim though

Can you clarify what do you mean by stability here?

Luthaf: > Can you clarify what do you mean by stability here?

Should the behavior of torch-sim stay the same when using default across multiple releases?

-----

In general, metatensor is not a model per-se, but rather an API that can be used to access multiple models with a given simulation engine (or multiple simulation engines with a given model). It is a lot closer to torch-sim's own `ModelInterface` than something like MACE or SevenNet.

PET-MAD is the first of such model to be widely available, but there will likely be other models coming up! So if we can leave this empty and require the user to specify which model they want to use through the metatensor interface, it would be better IMO.

torch_sim/models/metatensor.py
Luthaf: Can models request a different NL from torch-sim? Or is the API of `neighbor_list_fn` defined around full NL?

abhijeetgangan: You can request a different NL when you initialize the model. For example, in MACE https://github.com/Radical-AI/torch-sim/blob/c79fc6591802df53a5fd0869b96715826ef0d09e/torch_sim/models/mace.py#L105

tests/models/test_metatensor.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_metatensor_initialization(device: torch.device) -> None:
+    """Test that the metatensor model initializes correctly."""
+    model = MetatensorModel(
+        model="pet-mad",
+        device=device,
+    )
+    assert model._device == device  # noqa: SLF001
+    assert model._dtype == torch.float32  # noqa: SLF001
```
orionarcher: I'd test the public attributes `model.device` and `model.dtype` instead of the private versions.

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,328 @@
...
+    from metatrain.utils.io import load_model
+
+except ImportError:
+
+    class MetatensorModel(torch.nn.Module, ModelInterface):
+        """Metatensor model wrapper for torch_sim.
+
+        This class is a placeholder for the MetatensorModel class.
+        It raises an ImportError if metatensor is not installed.
+        """
```
orionarcher, line 39: To work with the documentation build this should define a dummy __init__ method too, something like: 

```suggestion
    class MetatensorModel(torch.nn.Module, ModelInterface):
        """Metatensor model wrapper for torch_sim.

        This class is a placeholder for the MetatensorModel class.
        It raises an ImportError if metatensor is not installed.
        """
        def __init__():
            raise ImportError("metatensor must be installed to use this model.")
```

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,328 @@
...
+        if self._compute_forces:
+            if len(results_by_system["forces"]) > 0:
+                results["forces"] = torch.cat(results_by_system["forces"])
+            else:
+                results["forces"] = torch.empty_like(positions)
+        if self._compute_stress:
+            if len(results_by_system["stress"]) > 0:
+                results["stress"] = torch.cat(results_by_system["stress"])
+            else:
+                results["stress"] = torch.empty_like(cell)
```
orionarcher: May be wrong because I am not sure about the shapes of the tensors in `results_by_system` but I believe that the forces should be concatenated and the stresses should be stacked here. Tests should catch it either way.

torch_sim/models/metatensor.py
```diff
@@ -0,0 +1,328 @@
...
+                system_cell = system_cell @ strain
+            system = System(
+                positions=system_positions,
+                types=system_atomic_numbers,
+                cell=system_cell,
+                pbc=system_pbc,
+            )
+
+            # Calculate neighbor list(s) for this system
+            for neighbor_list_options in self._requested_neighbor_lists:
```
orionarcher: Do all models in metatensor use a radial cutoff for the neighbor list?

If not, consider setting a dynamic `memory_scales_with` @propery. The default memory_scales_with from ModelInterface will return "n_atoms_x_density", which is only correct for models with radial cutoffs.

For max-neighbors based cutoffs, the memory scaler will be incorrect and the autobatching will subtly break. See [here](https://radical-ai.github.io/torch-sim/tutorials/autobatching_tutorial.html#Understanding-Memory-Requirements) for a more detailed discussion.

torch_sim/models/metatensor.py
abhijeetgangan: I don't foresee any breaking API changes, as we have already discussed it before and settled on the one we currently have.

As you suggested, I think it is appropriate for users to specify the model through the interface.

Not a requirement but it would be nice to have a [tutorial](https://github.com/Radical-AI/torch-sim/pull/118/files#diff-3edfcb7ad0b651fb8d1e2171a983262b4efade0857bb4d0a784678d4776d20f1) so that users can have a quick start with the metatensor models.

Also, thank you for the PR. Would like to try the PET-MAD model myself!

orionarcher (changes requested):
Thanks for the contribution @frostedoyster, this will be very nice interface to have.

Couple comments just to make sure the interface behaves as intended.

For added context, I want to point you to our [Current Posture on Model Interfaces](https://github.com/Radical-AI/torch-sim/discussions/120), just to share how we are thinking about supporting models over time.

torch_sim/models/metatensor.py
frostedoyster: Yes, at the moment this interface only supports a fixed cutoff radius, although metatensor itself would allow for a clamped number of neighbors by simply feeding a filtered neighbor list. However, we have never tested this possibility in training or in production, and therefore I would keep it as is for now

torch_sim/models/metatensor.py
frostedoyster: Absolutely, this was a bug on my side, thanks for pointing it out. No, the tests weren't catching it... Perhaps it's because, during the tests, the model was always called on a single structure (or that's what I'm seeing with a naively injected print statement), except for one 2-structure call. Or maybe you guys call a reshape just after model evaluation which would have fixed the above bug automatically

torch_sim/models/metatensor.py
frostedoyster: Done, thanks!

torch_sim/models/metatensor.py
orionarcher: The consistency tests won't catch it up the validate model output tests should. If not, I need to fix the tests 🙂

```python
test_mattersim_model_outputs = make_validate_model_outputs_test(
    model_fixture_name="metatensor_model",
)
```

torch_sim/models/metatensor.py
frostedoyster: Thanks for the suggestion, I added a brief tutorial. Let me know what you think!

torch_sim/models/metatensor.py
frostedoyster: All the tests were already passing even before the bug fix... I wouldn't be able to give you more details than this unfortunately

frostedoyster:
Sorry for all the failures:
- the formatting pre-commit hook wasn't working for me, it should be good now
- the tests are working locally for me, using a fresh environment and the current imports. There was a small syntax issue in my `test.yaml`, should be good now
- I can't get the docs to build locally with any torch version, due to MACE failing to load. I would need some help with that, if possible

torch_sim/models/metatensor.py
Luthaf: There is also the possibility for model to set `interaction_range` to infinity, which is used for long-range models. I don't know if these would work with the current code, I'll check.

Luthaf:
Err, I already signed the CLA, and if I try to sign it again I get a `Record of your CLA already exists.` error message.

torch_sim/models/metatensor.py
Luthaf: Ok, I updated this!

CompRhys:
> Err, I already signed the CLA, and if I try to sign it again I get a `Record of your CLA already exists.` error message.

The api that the bot checks doesn't have a signed CLA associated with the Github account "Luthaf", I don't have deeper access to check the actual records. Could you try again to register the github username, if that doesn't work we will debug further.

Luthaf:
I might have registered as `luthaf` instead of `Luthaf`, but most of the github-related things accept either. Trying to register again gives me the `Record of your CLA already exists` error.

abhijeetgangan (approved):

CompRhys:
> Sorry for all the failures:
> 
> * the formatting pre-commit hook wasn't working for me, it should be good now
> * the tests are working locally for me, using a fresh environment and the current imports. There was a small syntax issue in my `test.yaml`, should be good now
> * I can't get the docs to build locally with any torch version, due to MACE failing to load. I would need some help with that, if possible

We're very glad to see contributions and there's always some teething issues trying to get new processes to work smoothly! 
Getting the doc's to build cleanly for contributions is something that is a learning curve for all of us. 

The SiO2 rattled structure test to test a high force configuration seems to be problematic on the Github runners for some models. We see quite big differences for MACE in CI when tests pass locally. Is that the same with PET-MAD here?

Luthaf:
The test errors on CI seems to only happen on ARM macos + float32 dtype. The tests are fine on ubuntu

I can also reproduce the failure locally on another ARM mac. So I'd guess that something makes the model not work quite as well on macos+float32 (maybe a different BLAS implementation?), which the CI is showing us. 

I'll see what's the best way to fix that with @frostedoyster

CompRhys:
> I might have registered as `luthaf` instead of `Luthaf`, but most of the github-related things accept either. Trying to register again gives me the `Record of your CLA already exists` error.

it's been updated it to be case insensitive, the next push to this branch should now add the signed label back

.github/workflows/docs.yml
```diff
@@ -33,10 +33,10 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv pip install .[test,docs] --system
+          uv pip install ".[test,docs]" --system
 
       - name: Install extras for tutorial generation
-        run: uv pip install ".[graphpes,mace]" --system
+        run: uv pip install ".[graphpes,mace,metatensor]" --system
```
CompRhys, line 39: Note, this may need to be refactored in future due to incompatible optional dependencies between tutorials?

CompRhys (approved):
- Fixed the docs issue, c.f. step 7 `docs/dev/add_model.md`.
- Fixed my incorrect rattling code to only use weibull for scale.

LGTM.

frostedoyster:
Thanks for the help! Just one last comment: you might want to update this line in the top-level readme to reflect all the new models that you've added recently

> Support for MACE, Fairchem, and SevenNet MLIP models with more in progress

.github/workflows/docs.yml
orionarcher: Yeah I am tracking this, will fix it if it becomes a problem

torch_sim/models/metatensor.py
orionarcher: The model scaling can be handled manually for edge cases, so I am happy leaving it as `n_atoms_x_density`

orionarcher (approved):
LGTM! I updated the README as suggested. Thanks for the contribution! I'll merge when tests are passing.

orionarcher:
Looking back at the discussion, I am realizing I may have been a bit trigger happy in merging this, apologies. Even though the docs build worked once, it's now failing in main. Should I revert this or would you prefer to take care of any remaining issues in a new PR?

EDIT: I'd suggest just opening a new PR and we can merge that in or make more changes as needed, sorry for confusion.

PR #126, opened 2025-04-09
FairChemModel Updates: PBC handling, test on OMat24 pre-trained model

AdeeshKolluru:
## Summary

- [x] Fixes https://github.com/Radical-AI/torch-sim/issues/111 
- [x] Add tests with OMat24 pre-trained model if HF_TOKEN env is set

Will include tests with OMat models as in [quacc](https://github.com/Quantum-Accelerators/quacc) in the next PR.

## Checklist

### DIFF
--- a/tests/models/test_fairchem.py
+++ b/tests/models/test_fairchem.py
@@ -66,8 +66,8 @@ test_fairchem_ocp_consistency_pbc = make_model_calculator_consistency_test(
     model_fixture_name="fairchem_model_pbc",
     calculator_fixture_name="ocp_calculator",
     sim_state_names=consistency_test_simstate_fixtures[:-1],
-    rtol=1e-4,  # NOTE: EqV2 doesn't pass at the 1e-5 level used for other models
-    atol=1e-4,
+    rtol=5e-3,  # NOTE: EqV2 doesn't pass at the 1e-5 level used for other models
+    atol=5e-3,
 )
 
 test_fairchem_non_pbc_benzene = make_model_calculator_consistency_test(
--- a/tests/models/test_fairchem.py
+++ b/tests/models/test_fairchem.py
@@ -66,8 +66,8 @@ test_fairchem_ocp_consistency_pbc = make_model_calculator_consistency_test(
     model_fixture_name="fairchem_model_pbc",
     calculator_fixture_name="ocp_calculator",
     sim_state_names=consistency_test_simstate_fixtures[:-1],
-    rtol=5e-4,  # NOTE: EqV2-OC20 doesn't pass at the 1e-5 level used for other models
-    atol=5e-4,
+    rtol=1e-4,  # NOTE: EqV2 doesn't pass at the 1e-5 level used for other models
+    atol=1e-4,
 )
 
 test_fairchem_non_pbc_benzene = make_model_calculator_consistency_test(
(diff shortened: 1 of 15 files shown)


torch_sim/models/fairchem.py
```diff
@@ -257,11 +277,9 @@ def __init__(  # noqa: C901, PLR0915
             inference_only=True,
         )
 
-        self.trainer.model = self.trainer.model.eval()
-
         if dtype is not None:
             # Convert model parameters to specified dtype
-            self.trainer.model = self.trainer.model.to(dtype=self.dtype)
+            self.trainer.model = self.trainer.model.to(dtype=self._dtype)
```
CompRhys: this is a property of the ModelInterface and so it's better to access that rather than the underlying "private" attribute

CompRhys (approved):
LGTM, one nit and should remove benezene from tests

tests/models/test_fairchem.py
```diff
@@ -41,28 +42,20 @@ def ocp_calculator(model_path: str) -> OCPCalculator:
...
-    model_fixture_name="fairchem_model",
+    model_fixture_name="fairchem_model_pbc",
     calculator_fixture_name="ocp_calculator",
-    sim_state_names=consistency_test_simstate_fixtures,
+    sim_state_names=consistency_test_simstate_fixtures[:-1],
     rtol=5e-4,  # NOTE: fairchem doesn't pass at the 1e-5 level used for other models
     atol=5e-4,
 )
 
+# TODO: add test for non-PBC model
```
CompRhys: Not clear that the OCP calculator handles non-pbc atoms so unclear how to do the comparison.

CompRhys (approved):

tests/models/test_fairchem.py
```diff
@@ -41,33 +61,32 @@ def ocp_calculator(model_path: str) -> OCPCalculator:
...
-    not torch.cuda.is_available(),
-    reason="Batching does not work properly on CPU for FAIRchem",
+    os.environ.get("HF_TOKEN") is None,
+    reason="Issues in graph construction of older models",
 )(
     make_validate_model_outputs_test(
-        model_fixture_name="fairchem_model",
+        model_fixture_name="fairchem_model_pbc",
     )
 )
```
orionarcher: From CI:

```tests/models/test_fairchem.py .............s                             [100%]```

It looks like this is being skipped in practice, could you modify the setup to get it running? It's a valuable check to have.

orionarcher (approved):
One comment on testing, LGTM other than that, thanks @AdeeshKolluru!

tests/models/test_fairchem.py
AdeeshKolluru: There's no HF_TOKEN set by default which makes the default model the older `EquiformerV2-31M-S2EF-OC20-All+MD` and it skips that test. The older models have issues in their graph construction as detailed [here](https://github.com/FAIR-Chem/fairchem/issues/428). OMat24 models are [gated](https://huggingface.co/facebook/OMAT24) and therefore can't be accessed without keys. With my HF access key, that test passes locally.

AdeeshKolluru: Ok, I've added the HF_TOKEN to the secrets and this test should work now

PR #127, opened 2025-04-11
Enhancement: Heat Flux Function

stefanbringuier:
## Summary

* Add `calc_heat_flux` function to compute heat flux vectors
* Supports both virial and convective terms
    * Can request virial term only 
* Optional centroid stress tensor form
* Supports batch processing

## Checklist

* [ ] I have read and agree to the Contributor's Certification in [CONTRIBUTING.md](https://github.com/TorchSim/torch-sim/pull/CONTRIBUTING.md).


torch_sim/quantities.py
```diff
@@ -133,6 +133,132 @@ def calc_kinetic_energy(
...
+        stress: Per-atom stress tensor components:
+            - If is_centroid_stress=False: shape (n_particles, 6) for
+              :math:`[\sigma_{xx}, \sigma_{yy}, \sigma_{zz},
+              \sigma_{xy}, \sigma_{xz}, \sigma_{yz}]`
+            - If is_centroid_stress=True: shape (n_particles, 9) for
+              :math:`[\mathbf{r}_{ix}f_{ix}, \mathbf{r}_{iy}f_{iy},
+              \mathbf{r}_{iz}f_{iz}, \mathbf{r}_{ix}f_{iy},
+              \mathbf{r}_{ix}f_{iz}, \mathbf{r}_{iy}f_{iz},
+              \mathbf{r}_{iy}f_{ix}, \mathbf{r}_{iz}f_{ix},
+              \mathbf{r}_{iz}f_{iy}]`
```
orionarcher: The stress tensor should have shape `n_batches, 3, 3`, would you convert this to follow that convention?

tests/test_quantities.py
```diff
@@ -0,0 +1,147 @@
...
+            velocities=mock_simple_system["velocities"],
+            energies=mock_simple_system["energies"],
+            stress=mock_simple_system["stress"],
+            is_virial_only=True,
+        )
+
+        expected = -torch.tensor([1.0, 4.0, 9.0], device=virial.device)
+        assert_allclose(virial.cpu().numpy(), expected.cpu().numpy())
+
+    def test_batched_calculation(self, device: torch.device) -> None:
```
orionarcher: Could you add a test to check that if I run `unbatched(A)` and `unbatched(B)` I get the same results as `batched(A, B)`?

orionarcher (changes requested):
LGTM, just request a couple small things. Tagging @abhijeetgangan for a look too.

abhijeetgangan:
- Currently, not all models return per-atom potential energies and stresses. It is out of our control if the original implementation of the model does not return those properties. The kinetic part is easy to add.
- One needs to be careful about the volume normalization. This is [not present](https://docs.lammps.org/compute_stress_atom.html) in the LAMMPS per-atom stresses so you might need to multiple by the volume.
- You can have a test case with [LJ](https://github.com/Radical-AI/torch-sim/blob/caa542357e366ca50dfdaae48cef54a46fd9a2a1/torch_sim/models/lennard_jones.py#L268) as it can return the per-atom properties.

Does this definition work for potentials that use angles? Since we are using auto-grad to get the stresses it will include all the terms beyond pair-wise interactions. LAMMPS has a [warning ](https://docs.lammps.org/compute_heat_flux.html) about that.

torch_sim/quantities.py
stefanbringuier: Sure, no problem.

stefanbringuier:
@abhijeetgangan  thanks for the comments and look-outs!

> * Currently, not all models return per-atom potential energies and stresses. It is out of our control if the original implementation of the model does not return those properties. The kinetic part is easy to add.

I  thought this might be an issue for general model use, any thoughts on how to handle?
 
> * One needs to be careful about the volume normalization. This is [not present](https://docs.lammps.org/compute_stress_atom.html) in the LAMMPS per-atom stresses so you might need to multiple by the volume.

Yeah, the formalism here does not include the domain volume normalization intentionally because I was intending to use it HFACF and for scenarios where the heatflux is calculated for a subset of atoms rather than the whole domain. But that was probably not a suitable reason to exclude it. Can add option to return domain volume normalized J.

> * You can have a test case with [LJ](https://github.com/Radical-AI/torch-sim/blob/caa542357e366ca50dfdaae48cef54a46fd9a2a1/torch_sim/models/lennard_jones.py#L268) as it can return the per-atom properties.

Working on a example script that was this.

> Does this definition work for potentials that use angles? Since we are using auto-grad to get the stresses it will include all the terms beyond pair-wise interactions. LAMMPS has a [warning ](https://docs.lammps.org/compute_heat_flux.html) about that.

Hmm, no haven't tested and need to review. The tests I've implemented just check formalism. Thanks!

abhijeetgangan:
- @stefanbringuier MACE returns per-atom potential energies, and it should be possible to get per-atom stresses. However, that means modifying the Shift-Scale MACE or doing the auto-grad in the forward in torch-sim. I am unsure what is the best way to deal with it is, but I will think about it over the weekend. We do want to use this with ML potentials, after all.

- Regarding points 2-3 - Do you have a good reference value from literature with LJ to get the Heat flux / Kappa? That would make a robust test as well. I will also search for it or maybe a similar LAMMPS example.

- We can safely use it for pairwise interaction potentials, I think. Need to be careful if the formalism itself is limiting for models like mattersim which use angles.

Thank you for the PR. This goes well with the correlation PR for Kappa calculations. I was planning to add it myself but wasn't sure because of the limitations I mentioned above.

stefanbringuier:
@abhijeetgangan

Did some searching and found a paper by [Admal & Tadmor (2015)](https://arxiv.org/pdf/1508.02250) where they derive heat flux expressions without requiring per-atom energies, using the energy balance law and force–position coupling. But the key issue is that it requires access to partial forces $f_{ij}$, which MLIPs don't typically expose to my knowledge.

Then I found [Carbogno et al. (2017)](https://arxiv.org/abs/1608.06917), where they adopt a simplified version using only the virial term:  $J_v = \frac{1}{V} \sum_i \sigma_i \cdot v_i$, with $\sigma_i \sim r_i \otimes f_i$.  Their argument is that for solids the convective term is negligible, so this form gives valid Green–Kubo $\kappa$. This avoids needing per-atom energies or per-atom stresses.

So my thinking is: when per-atom stresses aren’t provided, we extend `calc_heat_flux` to optionally accept per-atom `positions` and `forces` and construct position-force contractions internally. Then if `energies=None`, we just return the virial part with a warning that the convective term is assumed negligible.

Let me know what you think? Will proceed if you're on board with that.

abhijeetgangan:
@stefanbringuier all good points. I have tested [here](https://github.com/Radical-AI/torch-sim/blob/main/examples/scripts/7_Others/7.2_Stress_autograd.py) some of them before.

I also merged the LJ to include per-atom quantities in a batch so you can proceed with a example script for both cases where you have per atom quantities vs positions-forces contractions following the script above.

I think there is a better way to make this approach more general but for now this should be good enough for a test case. This reference has a detailed [derivation of the virial](https://doi.org/10.1063/1.3245303). I would also include the references you mentioned.

janosh:
what's the state of this PR? happy to review if it's ready

stefanbringuier:
> what's the state of this PR? happy to review if it's ready

It's still not quite ready as I need to review @abhijeetgangan [tests](https://github.com/Radical-AI/torch-sim/blob/main/examples/scripts/7_Others/7.2_Stress_autograd.py) and also work on an example/test using the LJ per-atom quantities vs. the contraction calculation. Will prioritize to bring the PR to a close (been focused on the NEB feature).

janosh:
ofc, no rush. happy to leave this open. just checking it's not held up by us

CompRhys:
Merging for proper credit assignment in advance of https://github.com/TorchSim/torch-sim/pull/264

PR #132, opened 2025-04-12
Plot `torch_sim` module graph

janosh:
add `get_module_graph_dot_file.py` script to generate DOT representation of `torch_sim` module graph

torch-sim-module-graph


janosh:
@orionarcher @AdeeshKolluru the huggingface login CI step is flaky:

```py
The token has not been saved to the git credentials helper. Pass `add_to_git_credential=True` in this function directly or `--add-to-git-credential` if using via `huggingface-cli` if you want to set the git credential as well.
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/site-packages/huggingface_hub/utils/_http.py", line 409, in hf_raise_for_status
    response.raise_for_status()
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://huggingface.co/api/whoami-v2

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/site-packages/huggingface_hub/hf_api.py", line 1737, in whoami
    hf_raise_for_status(r)
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/site-packages/huggingface_hub/utils/_http.py", line 473, in hf_raise_for_status
    raise _format(HfHubHTTPError, message, response) from e
huggingface_hub.errors.HfHubHTTPError: 403 Forbidden: None.
Cannot access content at: https://huggingface.co/api/whoami-v2.
Make sure your token has the correct permissions.
```

orionarcher:
Good catch @janosh. Raised an issue #133 to note this. As an interim solution, we can modify hugging face step in the action to only trigger for the fairchem models, that will reduce the frequency of this failure.

The `5.3_Hot_Swap_WBM.py` script is also flaky.

janosh:
@abhijeetgangan @orionarcher any reservations about showing this in the readme? could be info overload. but the docs (https://radical-ai.github.io/torch-sim/user/introduction.html) import the readme and i would like to show it in the docs...

orionarcher:
Thanks @janosh! Very cool feature

janosh:
honestly less insightful than i expected. but the work was already done so might as well use it. as expected `state`, `models.interface` and `neighbors` are frequently imported across the package. `transforms` was a bit of a surprise. glad i bumped the coverage on that before release. if we find bugs, at least the tests will be easy to update. 2nd surprise was how much code there is in `unbatched_integrators`

PR #136, opened 2025-04-13
Revert "Add metatensor models (#124)"

orionarcher:
This reverts commit 8437e6d72a90356ed0dc1d6053ae1cf5e17c18f1.

## Summary

I was trigger happy and merged this too early because the tests were preliminarily passing. Reverting to give PR authors opportunity to fix.


CompRhys:
The failures are due to the HF download? which is a flaky thing similar to the WBM download. https://github.com/Radical-AI/torch-sim/pull/134 attempted to reduce the likelihood of this but I think it's likely that the network for the github runners just isn't good enough to download big files. Could maybe look into adding these two files as artifacts: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow?

orionarcher:
oops I did not intend to merge this yet, shoot.

There was a previous build failure due to the metatensor model but it appears to be working now. Also related to a file download. Now we are seeing an error with the graph-pes tests. 

I think artifacts would be a good solution. I've temporarily disabled the MBD download in the script.

PR #138, opened 2025-04-14
Enhancement: FairChem model from_data_list support

ryanliu30:
## Summary

a small addition to use Batch.from_data_list instead of Batch constructor to support indexing the batch object via batch[i] which is used in some fairchem models.


AdeeshKolluru (approved):
LGTM, thanks for the contribution!

PR #139, opened 2025-04-14
Implement positions only FIRE

### LINKED ISSUE #107: `unit_cell_fire` requires stress for evaluation even when cell is fixed
The check for stress should be conditional or a new method should be created.

abhijeetgangan:
## Summary

Closes #107

## Checklist


torch_sim/optimizers.py
```diff
@@ -445,6 +445,252 @@ def gd_step(
...
+        ) * state.velocities + batch_wise_alpha * state.forces * v_norm / (f_norm + eps)
+
+        return state
+
+    return fire_init, fire_step
+
+
 @dataclass
 class BatchedUnitCellFireState(SimState, DeformGradMixin):
     """State information for batched FIRE optimization with unit cell degrees of
```
orionarcher: We could take this opportunity to rename `BatchedUnitCellFire` -> `UnitCellFire`, and also remove the `Batched` descriptor from the other SimStates in this file, since batching is the default.

We could also make this inherit from `FireState` instead of `SimState`.

tests/test_optimizers.py
```diff
@@ -101,6 +102,49 @@ def test_unit_cell_gradient_descent_optimization(
     assert not torch.allclose(state.cell, initial_state.cell)
 
 
+def test_fire_optimization(
+    ar_supercell_sim_state: SimState, lj_model: torch.nn.Module
+) -> None:
```
orionarcher, line 107: Another nice test would be to check that `frechet_cell_fire` and `unit_cell_fire` with fixed cells return the same result as `fire`.

orionarcher (commented):
Couple small suggestions, thanks @abhijeetgangan

PR #140, opened 2025-04-14
Orbv3 Updates

AdeeshKolluru:
## Summary

Adapts the OrbModel with recent changes in OrbCalculator and updates tests with both direct and conservative OrbV3 models.

## Checklist


torch_sim/models/orb.py
```diff
@@ -312,17 +312,15 @@ def __init__(
...
 
         # Set up system configuration
-        self.system_config = system_config or SystemConfig(
-            radius=6.0, max_num_neighbors=20
-        )
+        self.system_config = system_config or model.system_config
         self._max_num_neighbors = max_num_neighbors
         self._edge_method = edge_method
         self._half_supercell = half_supercell
+        self._conservative = conservative
```
orionarcher: should any of these be exposed as public / can be changed on the fly?

orionarcher (approved):
LGTM! one comment

torch_sim/models/orb.py
```diff
@@ -416,14 +410,16 @@ def forward(self, state: SimState | StateDict) -> dict[str, torch.Tensor]:
             _property = "energy" if prop == "free_energy" else prop
             results[prop] = predictions[_property].squeeze()
 
-        if self.conservative:
-            results["direct_forces"] = results["forces"]
-            results["direct_stress"] = results["stress"]
+        if self._conservative:
+            if self.model.forces_name in results:
+                results["direct_forces"] = results[self.model.forces_name]
```
CompRhys: If these lines aren't covered by the tests it must mean that these are not defined for conservative models? As such can lines 413-417 be deleted?

CompRhys (approved):
LGTM

One nit - I think that the code about "direct_stress" and "direct_forces" might be better to remove because it doesn't get triggered for the v3 models?

orionarcher:
Good to merge? I am going to mint a v0.2 release soon and this would be nice to include.

torch_sim/models/orb.py
AdeeshKolluru: Pulled the `conservative` attribute out.

torch_sim/models/orb.py
AdeeshKolluru: You are right that the v3 conservative models don't seem to have direct predictions heads as well. I could remove it, but I wanted to make sure that the logic closely follow that of [OrbCalculator](https://github.com/orbital-materials/orb-models/blob/main/orb_models/forcefield/calculator.py#L116-L119).

AdeeshKolluru: Removed it to make codecov happy!

Issue #133, opened 2025-04-12
Reduce flaky CI failures


orionarcher:
Examples `5.3_Hot_Swap_WBM.py` and the hugging face CI both appear to be a bit flaky and fail occasionally

orionarcher:
Flakiness with graph-pes: 

https://github.com/Radical-AI/torch-sim/actions/runs/14447521495/job/40511689547?pr=141

https://github.com/Radical-AI/torch-sim/actions/runs/14447521495/job/40511683512?pr=141

orionarcher:
As does flakiness with meta-tensor:

https://github.com/Radical-AI/torch-sim/actions/runs/14460092612/job/40550984333?pr=146

orionarcher:
And fairchem is still flaky too:

https://github.com/Radical-AI/torch-sim/actions/runs/14470883508/job/40584221809

WillEngler:
The example logs Orion posted got garbage collected by GH Actions, so reposting some more recent logs as I try to understand the root cause:

graph-pes failues:
- https://github.com/TorchSim/torch-sim/actions/runs/18176792760/job/51744500738 (casio3_sim_state)
- https://github.com/TorchSim/torch-sim/actions/runs/18144062426/job/51641704586#step:8:149 (rattled_sio2_sim_state)
- https://github.com/TorchSim/torch-sim/actions/runs/17803499812/job/50609353292 (tio2_sim_state)

can't download sevennet weights from GitHub? (less common than the graph-pes stuff):
- https://github.com/TorchSim/torch-sim/actions/runs/16889250592/job/47844588700

PR #143, opened 2025-04-14
Reduce test flakiness and hotfix docs

orionarcher:
## Summary

* [x] reduce flakiness of graph-pes test by loosening constraint #133 
* [x] fix small docs issue
* [x] Rename `HotSwappingAutobatcher` to `InFlightAutobatcher` and `ChunkingAutoBatcher` to `BinningAutoBatcher`

## Checklist

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -12,7 +12,7 @@ from itertools import chain
 
 import torch
 
-from torch_sim.autobatching import ChunkingAutoBatcher, HotSwappingAutoBatcher
+from torch_sim.autobatching import BinningAutoBatcher, InFlightAutoBatcher
 from torch_sim.models.interface import ModelInterface
 from torch_sim.quantities import batchwise_max_force, calc_kinetic_energy, calc_kT
 from torch_sim.state import SimState, StateLike, concatenate_states, initialize_state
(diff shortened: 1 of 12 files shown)


torch_sim/runners.py
```diff
@@ -491,7 +491,7 @@ class StaticState(type(state)):
 
     trajectory_reporter.finish()
 
-    if isinstance(batch_iterator, ChunkingAutoBatcher):
+    if isinstance(batch_iterator, BinningAutoBatcher):
```
janosh, line 494: while we're at it, maybe rename `batch_iterator`->`auto_batcher`? and `_configure_batches_iterator`-> `_configure_auto_batcher`

janosh (approved):

torch_sim/runners.py
orionarcher: The naming is different because if autobatcher is `False` then `_configure_auto_batcher` returns a list instead of an autobatcher. Maybe it's more confusing this way though?

janosh: no that makes sense. you could add your comment to the code to clarify

orionarcher: Done, also fixed the typing, good catch!

PR #150, opened 2025-04-15
Increase `neighbors.py` test coverage and speedup tests 5x

janosh:
looking at the current coverage report, `torch_sim/neighbors.py` stands out as crucial functionality to be correct yet with low test coverage. this PR bumps it from 14% to 19% (which is still low but a start) and speeds up the tests by 5x from 75 sec to 15 sec.

```diff
Name                                           Stmts   Miss  Cover
|------------------------------------------------------------------
torch_sim/__init__.py                             16      0   100%
torch_sim/autobatching.py                        270    243    10%
torch_sim/elastic.py                             393    359     9%
torch_sim/integrators.py                         292    262    10%
torch_sim/io.py                                  113    103     9%
torch_sim/math.py                                338    304    10%
torch_sim/models/__init__.py                      31     14    55%
torch_sim/models/fairchem.py                     134    115    14%
torch_sim/models/graphpes.py                      51     31    39%
torch_sim/models/interface.py                     83     54    35%
torch_sim/models/lennard_jones.py                 82     70    15%
torch_sim/models/mace.py                         105     90    14%
torch_sim/models/mattersim.py                     50     36    28%
torch_sim/models/metatensor.py                    98     85    13%
torch_sim/models/morse.py                         81     71    12%
torch_sim/models/orb.py                          128    110    14%
torch_sim/models/sevennet.py                      98     78    20%
torch_sim/models/soft_sphere.py                  178    161    10%
torch_sim/monte_carlo.py                          67     54    19%
- torch_sim/neighbors.py                           185    150    19%
torch_sim/optimizers.py                          426    355    17%
torch_sim/properties/__init__.py                   2      0   100%
torch_sim/properties/correlations.py             177    153    14%
torch_sim/quantities.py                           39     31    21%
torch_sim/runners.py                             152    132    13%
torch_sim/state.py                               280    211    25%
torch_sim/trajectory.py                          277    230    17%
torch_sim/transforms.py                          272    236    13%
torch_sim/unbatched/models/lennard_jones.py       81     69    15%
torch_sim/unbatched/models/morse.py               76     63    17%
torch_sim/unbatched/models/soft_sphere.py        153    137    10%
torch_sim/unbatched/unbatched_integrators.py     493    420    15%
torch_sim/units.py                                59      0   100%
torch_sim/workflows/__init__.py                    0      0   100%
torch_sim/workflows/a2c.py                       233    233     0%
|------------------------------------------------------------------
TOTAL                                           5513   4660    15%
```


tests/test_neighbors.py
```diff
@@ -116,10 +117,11 @@ def periodic_atoms_set():
         bulk("Cu", "fcc", a=3.6),
         bulk("Si", "bct", a=6, c=3),
         bulk("Ti", "hcp", a=2.94, c=4.64, orthorhombic=False),
-        # test very skewed unit cell
+        # test very skewed rhombohedral cells
         bulk("Bi", "rhombohedral", a=6, alpha=20),
-        bulk("Bi", "rhombohedral", a=6, alpha=10),
-        bulk("Bi", "rhombohedral", a=6, alpha=5),
```
janosh, line 122: @abhijeetgangan how important is the `bulk("Bi", "rhombohedral", a=6, alpha=5)` test case? given the skew, it's a huge cell and takes a long time to test. this system was taking >70% of the time

abhijeetgangan: @janosh these would be cases where I found the old neighbor list to fail. That being said alpha of 5 is an overkill for CI and can safely be either increased or the test can be removed.

janosh: your call. if it's even somewhat likely to prevent subtle bugs, then worth keeping. but `alpha=10,20` are still there and might catch those kinds of bugs as well?

abhijeetgangan: The tests failed for even for `alpha=20` so we can keep `alpha=10,20` and that should be good enough.

orionarcher (approved):
Flakiness is a separate issue, LGTM

PR #156, opened 2025-04-16
Make sure to move data to CPU before calling vesin

### LINKED ISSUE #155: `metatensor` tests fail on GPU
One shortcoming of our current CI is that all tests are only run on CPU. When running the tests on GPU, I am finding that the `metatensor` tests are failing, error shown below. It's possible that this is an environmental configuration issue on my end, but I'd like to check if the tests have been tested on GPU too. @frostedoyster, @Luthaf, would you mind taking a look? Do you have access to a CUDA device you could run the tests on?

The underlying error appears to be in `vesin`.

```py
Running tests (pytest): /workspaces/propfoliotorchsim/torch-sim/tests/models/test_metatensor.py::test_mattersim_model_outputs
Running test with arguments: --rootdir /workspaces/propfoliotorchsim/torch-sim --override-ini junit_family=xunit1 --junit-xml=/tmp/tmp-2510298TvuaQeZ2OUXQ.xml ./tests/models/test_metatensor.py::test_mattersim_model_outputs
Current working directory: /workspaces/propfoliotorchsim/torch-sim
Workspace directory: /workspaces/propfoliotorchsim/torch-sim
Run completed, parsing output
./tests/models/test_metatensor.py::test_mattersim_model_outputs Failed: [undefined]RuntimeError: device cuda:0 is not supported in vesin
request = <FixtureRequest for <Function test_mattersim_model_outputs>>
device = device(type='cuda'), dtype = torch.float32

    def test_model_output_validation(
        request: pytest.FixtureRequest,
        device: torch.device,
        dtype: torch.dtype,
    ) -> None:
        """Test that a model implementation follows the ModelInterface contract."""
        # Get the model fixture dynamically
        model: ModelInterface = request.getfixturevalue(model_fixture_name)
    
        from ase.build import bulk
    
        from torch_sim.io import atoms_to_state
    
        assert model.dtype is not None
        assert model.device is not None
        assert model.compute_stress is not None
        assert model.compute_forces is not None
    
        try:
            if not model.compute_stress:
                model.compute_stress = True
            stress_computed = True
        except NotImplementedError:
            stress_computed = False
    
        try:
            if not model.compute_forces:
                model.compute_forces = True
            force_computed = True
        except NotImplementedError:
            force_computed = False
    
        si_atoms = bulk("Si", "diamond", a=5.43, cubic=True)
        fe_atoms = bulk("Fe", "fcc", a=5.26, cubic=True).repeat([3, 1, 1])
    
        sim_state = atoms_to_state([si_atoms, fe_atoms], device, dtype)
    
        og_positions = sim_state.positions.clone()
        og_cell = sim_state.cell.clone()
        og_batch = sim_state.batch.clone()
        og_atomic_numbers = sim_state.atomic_numbers.clone()
    
>       model_output = model.forward(sim_state)

tests/models/conftest.py:152: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
torch_sim/models/metatensor.py:226: in forward
    vesin.torch.metatensor.compute_requested_neighbors(
../propfolio/.venv/lib/python3.12/site-packages/vesin/torch/metatensor/_model.py:81: in compute_requested_neighbors
    neighbors = calculator.compute(system)
../propfolio/.venv/lib/python3.12/site-packages/vesin/torch/metatensor/_neighbors.py:120: in compute
    (P, S, D) = self._nl.compute(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <vesin.torch._neighbors.NeighborList object at 0x7ffbc4a07140>
points = tensor([[0.0000, 0.0000, 0.0000],
        [1.3575, 1.3575, 1.3575],
        [0.0000, 2.7150, 2.7150],
        [1.3575,...50, 0.0000],
        [4.0725, 4.0725, 1.3575]], device='cuda:0', dtype=torch.float64,
       grad_fn=<ToCopyBackward0>)
box = tensor([[5.4300, 0.0000, 0.0000],
        [0.0000, 5.4300, 0.0000],
        [0.0000, 0.0000, 5.4300]], device='cuda:0', dtype=torch.float64,
       grad_fn=<ToCopyBackward0>)
periodic = True, quantities = 'PSD', copy = True

    def compute(
        self,
        points: torch.Tensor,
        box: torch.Tensor,
        periodic: bool,
        quantities: str,
        copy: bool = True,
    ) -> List[torch.Tensor]:
        """
        Compute the neighbor list for the system defined by ``positions``, ``box``, and
        ``periodic``; returning the requested ``quantities``.
    
        ``quantities`` can contain any combination of the following values:
    
        - ``"i"`` to get the index of the first point in the pair
        - ``"j"`` to get the index of the second point in the pair
        - ``"P"`` to get the indexes of the two points in the pair simultaneously
        - ``"S"`` to get the periodic shift of the pair
        - ``"d"`` to get the distance between points in the pair
        - ``"D"`` to get the distance vector between points in the pair
    
        :param points: positions of all points in the system
        :param box: bounding box of the system
        :param periodic: should we use periodic boundary conditions?
        :param quantities: quantities to return, defaults to "ij"
        :param copy: should we copy the returned quantities, defaults to ``True``.
            Setting this to ``False`` might be a bit faster, but the returned tensors
            are view inside this class, and will be invalidated whenever this class is
            garbage collected or used to run a new calculation.
    
        :return: list of :py:class:`torch.Tensor` as indicated by ``quantities``.
        """
    
>       return self._c.compute(
            points=points,
            box=box,
            periodic=periodic,
            quantities=quantities,
            copy=copy,
        )
E       RuntimeError: device cuda:0 is not supported in vesin

../propfolio/.venv/lib/python3.12/site-packages/vesin/torch/_neighbors.py:50: RuntimeError

Total number of tests expected to run: 1
Total number of tests run: 1
Total number of tests passed: 0
Total number of tests failed: 1
Total number of tests failed with errors: 0
Total number of tests skipped: 0
Total number of tests with no result data: 0
Finished running tests!
```

Luthaf:
## Summary

Fix #155

I don't have access to a CUDA GPU, so I would appreciate either @frostedoyster or @orionarcher checking that this does work in practice! 

## Checklist

* [ ] ~Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google).~
* [ ] ~Tests have been added for any new functionality or bug fixes.~

### DIFF
--- a/examples/tutorials/metatensor_tutorial.py
+++ b/examples/tutorials/metatensor_tutorial.py
@@ -3,8 +3,8 @@
 #   <summary>Dependencies</summary>
 # /// script
 # dependencies = [
-#     "metatrain[pet] >=2025.4",
-#     "metatensor-torch >=0.7,<0.8"
+#     "metatrain[pet]>=2025.4",
+#     "metatensor-torch>=0.7,<0.8"
 # ]
 # ///
 # </details>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ test = [
 ]
 mace = ["mace-torch>=0.3.11"]
 mattersim = ["mattersim>=0.1.2"]
-metatensor = ["metatensor-torch >=0.7,<0.8", "metatrain[pet] >=2025.4"]
+metatensor = ["metatensor-torch>=0.7,<0.8", "metatrain[pet]>=2025.4"]
 orb = ["orb-models>=0.5.2"]
 sevenn = ["sevenn>=0.11.0"]
 graphpes = ["graph-pes>=0.0.34", "mace-torch>=0.3.11"]
--- a/torch_sim/models/metatensor.py
+++ b/torch_sim/models/metatensor.py
@@ -223,9 +223,14 @@ class MetatensorModel(torch.nn.Module, ModelInterface):
             )
 
         # Calculate the required neighbor list(s) for all the systems
+
+        # move data to CPU because vesin only supports CPU for now
+        systems = [system.to(device="cpu") for system in systems]
         vesin.torch.metatensor.compute_requested_neighbors(
             systems, system_length_unit="Angstrom", model=self._model
         )
+        # move back to the proper device
+        systems = [system.to(device=self.device) for system in systems]
 
         # Get model output
         model_outputs = self._model(


orionarcher:
Works! Thanks for the quick fix. 

Question though, is there really no way to run vesin on GPU? I thought that was the role of vesin-torch.

Luthaf:
> Question though, is there really no way to run vesin on GPU? I thought that was the role of vesin-torch.

Not currently no =) We are working to create a CUDA and/or HIP implementation to make sure the code can run on GPU, hopefully in the next couple of months. 

If you want a pure torch implementation of neighbor list, that can run on GPU (because it only uses functions from torch), you can try using https://github.com/felixmusil/torch_nl/, but I found that it is slower than vesin even when running on GPU, except when dealing with very large cells and large (>12A) cutoffs. https://github.com/openmm/NNPOps/ does contain some custom CUDA code for neighbor list, but requires a cell larger than 2 cutoff.

`vesin-torch` is only exposing the code from `vesin` in a TorchScript compatible way, and integrating with autograd.

orionarcher:
Thanks for the explanation!

Will merge for now and look forward to the upcoming implementation!

abhijeetgangan:
The torch version of neighbor lists which run on GPU are already included in torchsim including an API for `torch_nl` which accepts the batch as input. https://github.com/Radical-AI/torch-sim/blob/main/examples/scripts/7_Others/7.3_Batched_neighbor_list.py

orionarcher:
If that could be subbed in it would be a nice addition to the metatensor model interface!

Luthaf:
I'm not sure it is actually worth doing, since in all my tests torch_nl is slower than vesin, even when running on GPU. So unless someone has a system where this is not the case, I would rather keep the code simple!

PR #165, opened 2025-04-18
Improves Orb's state_to_atoms_graph calculation

AdeeshKolluru:
## Summary

Makes Orb's `state_to_atoms_graph` calculation slightly more efficient with a single loop across the batch by reverting from `batch_compute_pbc_radius_graph` which internally was already looping across the batch.


orionarcher (approved):
LGTM

Issue #171, opened 2025-04-22
Fractional coordinates converting issue on non orthogonal cell


Seungwoo-Hwang:
Thank you for providing a wonderful project!

Recently, I found errors when running MD with tilted cell structures.
The phenomena that I saw are below:
1. When I run batched MD with the below code, the structure is destroyed suddenly. (See POSCAR_final_state_pymatgen)
2. Even though I run with T=0 nvt MD, some atomic positions are changed.

```py
import os
import torch
import torch_sim as ts
import time
import numpy as np

from sevenn.calculator import SevenNetCalculator
import sevenn
from torch_sim.models import SevenNetModel

from pymatgen.core import Structure

cp = sevenn.util.load_checkpoint('7net-0')
backend = 'e3nn'
model_loaded = cp.build_model(backend)
model_loaded.set_is_batch_data(True)
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
model= model_loaded.to(device)
sevennet_model = SevenNetModel(model=model, device=device)


atoms = []
trajectory_files = []
structure = Structure.from_file('POSCAR_test')
atoms.append(structure)
trajectory_files.append(f"traj.h5md")


prop_calculators = {
    1: {"potential_energy": lambda state: state.energy,
        "kinetic_energy": lambda state: ts.calc_kinetic_energy(
            state.momenta, state.masses
        )
    }
}
reporter = ts.TrajectoryReporter(
    trajectory_files,
    state_frequency=1,  # Save the state every 10 steps
    prop_calculators=prop_calculators,
)

final_state = ts.integrate(
    system=atoms,
    model=sevennet_model,
    n_steps=5,
    timestep=0.002,
    temperature=0,
    integrator=ts.nvt_langevin,
    trajectory_reporter=reporter,
)

structure = final_state.to_structures()[0]
structure.to('POSCAR_final_state_pymatgen', fmt='poscar')
```

And here is the POSCAR_test that I used.
```
P  Ga O  Li
 1.0000000000000000
     8.4961804365324287   -0.0031724345975384   -0.0080141787223568
    -4.2436788574713287    7.3640240420354619   -0.0021261026511822
    -0.2252045807890393    0.1539663100109199   12.6097775384354946
 P   Ga  O   Li 
  12   6  42   6
Direct
  0.7966345194985355  0.7097171658326071  0.0218229037420674
  0.1022368681617419  0.4003667166635160  0.0257948401446204
[...]

Issue #17, opened 2025-03-05
Change default positions to unwrapped


orionarcher:

CompRhys:
Consider unwrapped fractional co-ordinates as it may avoid issues such as we saw here: https://github.com/Radical-AI/torch-sim/pull/199

thomasloux:
Done #294

PR #175, opened 2025-04-25
Updates integrator logic per #171. 

CompRhys:
## Summary

The tests of the function do not use the row vector cell so this suggests this is a bug that it's used in the integrators.

see that tests of function don't use the swap axis: https://github.com/Radical-AI/torch-sim/blob/d2cee6628ff3b4a8da7557772cc12a669a0c09b4/tests/test_transforms.py#L267

## Checklist

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -329,40 +329,16 @@ def test_pbc_wrap_batched_triclinic(device: torch.device) -> None:
     # Stack the cells for batched processing
     cell = torch.stack([cell1, cell2])
 
-    # Define positions (r_row convention)
-    positions = torch.tensor(
-        [
-            [2.5, 2.5, 2.5],  # Atom 0 (batch 0)
-            [2.7, 2.7, 2.7],  # Atom 1 (batch 1)
-        ],
-        dtype=torch.float64,
-        device=device,
-    )
-    batch = torch.tensor([0, 1], device=device)
-
-    # --- Calculate Expected Results Manually (M_row, r_row) --- #
-    def manual_wrap(r_row: torch.Tensor, M_row: torch.Tensor) -> torch.Tensor:
-        inv_M_row_T = torch.linalg.inv(M_row.T)
-        f_row = r_row @ inv_M_row_T
-        wrapped_f_row = f_row - torch.floor(f_row)
-        # Handle boundary cases (optional but good practice)
-        wrapped_f_row = torch.where(
-            torch.isclose(wrapped_f_row, torch.ones_like(wrapped_f_row)),
-            torch.zeros_like(wrapped_f_row),
-            wrapped_f_row,
-        )
-        return wrapped_f_row @ M_row
-
-    expected1 = manual_wrap(positions[0], cell1)
-    expected2 = manual_wrap(positions[1], cell2)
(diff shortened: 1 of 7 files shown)


CompRhys:
~~True error still not showing up with lazy import route, pyi needed to get the language server to work with the lazy imports.~~ reset and moved lazy imports to #179

CompRhys:
#17 is also related to this as if we solve that we can just drop the call to roll the positions back into the cell from the integrators.

janosh:
my preference is to tackle that then and address #17 here instead of merging just a minor patch

CompRhys:
I would still merge this but just wanted to raise that to bump the idea

janosh:
i think i found 2 more bugs in `transforms.py` were the code was still written to work with column vector lattices while the rest of torch-sim now uses row vector convention as of https://github.com/Radical-AI/torch-sim/pull/112


need some eyes on the changes in [6b82750](https://github.com/Radical-AI/torch-sim/pull/175/commits/6b82750f050f2a7b94f738c02078fb3c7b221621) [1e7290e](https://github.com/Radical-AI/torch-sim/pull/175/commits/1e7290e1e472fb0a7499f6c731621460b9f53553) from preferably two reviewers @orionarcher @abhijeetgangan @CompRhys

tests/test_transforms.py
```diff
@@ -99,25 +99,30 @@ def test_pbc_wrap_general_triclinic() -> None:
...
     )
 
     # Position outside triclinic box
-    positions = torch.tensor([[2.5, 2.5, 2.5]])
+    positions = torch.tensor([[2.5, 2.5, 2.5]], dtype=torch.float64)
 
-    # Correct expected wrapped position for this triclinic cell
-    expected = torch.tensor([[2.0, 0.5, 0.2]])
+    # Expected wrapped position calculated using r' = (r @ inv(M_row.T) % 1) @ M_row
+    expected = torch.tensor([[1.875, 0.9875, 0.125]], dtype=torch.float64)
```
CompRhys: These weird values for the expected result seems self-evidently to suggest broken logic to me? These values must be projected onto the lattice vectors which isn't what I would expect the function to do

CompRhys:
```py
import numpy as np
from pymatgen.core import Lattice, Structure

lattice = Lattice(np.array([[2.0, 0.5, 0.0],
                   [0.0, 2.0, 0.0],
                   [0.0, 0.3, 2.0]]).T)

# Create structure with a single H atom at the given position
s = Structure(lattice, ["H"], [[2.5, 2.5, 2.5]], coords_are_cartesian=True)

# Get the wrapped (folded) position in cartesian coordinates
wrapped_pos = s.frac_coords[0] % 1  # fractional coords wrapped to [0,1)
wrapped_cart = lattice.get_cartesian_coords(wrapped_pos)

print(wrapped_cart)
```

CompRhys: this gives the same answer as the old test. I think that also suggests that there was an issue with the old function as I am giving the lattice as column vectors (due to transpose) to a function that expects row vectors.

CompRhys:
@abhijeetgangan this okay to merge?

janosh (approved):
looks great! thanks @CompRhys 👍

Issue #180, opened 2025-04-27
Feature request: tqdm for runners


YutackPark:
I'm a `tqdm` lover. Sometimes this is essential when running huge number of systems but you don't know throughput at all.

I think for `static` and `integrate`, it is easy to implement. For `optimize`, here's the working code (see `added line`). It updates the progress bar when the structure meets convergence criteria.
```python
      from tqdm import tqdm
      pbar = tqdm(total=state.n_batches)  # added line
      while (result := autobatcher.next_batch(state, convergence_tensor))[0] is not None:
          state, converged_states, batch_indices = result
          all_converged_states.extend(converged_states)

          # need to update the trajectory reporter if any states have converged
          if trajectory_reporter and (step == 1 or len(converged_states) > 0):
              trajectory_reporter.load_new_trajectories(
                  filenames=[og_filenames[i] for i in batch_indices]
              )

          for _step in range(steps_between_swaps):
              last_energy = state.energy

              state = update_fn(state)

              if trajectory_reporter:
                  trajectory_reporter.report(state, step, model=model)
              step += 1
              if step > max_steps:
                  # TODO: max steps should be tracked for each structure in the batch
                  warnings.warn(f"Optimize has reached max steps: {step}", stacklevel=2)
                  break

          convergence_tensor = convergence_fn(state, last_energy)
          pbar.update(len([tf for tf in convergence_tensor if tf]))  # added line
```

Another option is using the `trajectory_reporter` but, I think the simplest solution is adding a new argument.

If dev agree with this, I could write PR

CompRhys:
I would support provided there's a verbose flag to turn it off. Also should consider logic to stop it spamming non-tty if running in slurm or something non-interactive (https://git.io/JnBOi).

janosh:
implemented in #181, thanks again @YutackPark 👍

PR #181, opened 2025-04-27
Add `pbar: bool | dict = False` keyword to `optimize`, `integrate`, `static` runners

YutackPark:
## Summary

Include a summary of major changes in bullet points:

* Feature: add `verbose` option for progress bar, if `verbose` is True, tty, and `autobatcher` is not False # update: `verbose` to `pbar`
* Bugfix: `static` runner restores order correctly

## Checklist


YutackPark:
#180 

*Last commit message (`refactor`) is somewhat wrong, it is the commit for the patch.

torch_sim/runners.py
```diff
@@ -399,6 +409,8 @@ def optimize(
                 break
 
         convergence_tensor = convergence_fn(state, last_energy)
+        if pbar:
+            pbar.update(len([tf for tf in convergence_tensor if tf]))
```
orionarcher: Could also do len(converged_states), would be a bit cleaner

torch_sim/runners.py
```diff
@@ -493,12 +509,15 @@ class StaticState(type(state)):
...
+            pbar.update(substate.n_batches)
 
     trajectory_reporter.finish()
 
     if isinstance(batch_iterator, BinningAutoBatcher):
         # reorder properties to match original order of states
         original_indices = list(chain.from_iterable(batch_iterator.index_bins))
-        return [all_props[idx] for idx in original_indices]
+        indexed_props = list(zip(original_indices, all_props, strict=True))
+        return [prop for _, prop in sorted(indexed_props, key=lambda x: x[0])]
```
orionarcher, line 547: Good catch on not restoring the order! Would you mind adding a test to check that the order is returned correctly? You can adapt the `test_integrate_with_autobatcher` to the static runner, which will fail if the order is not correct.

Maybe you could use the autobatcher `restore_original_order` functionality here too?

```python
if isinstance(batch_iterator, BinningAutoBatcher):
    reordered_states = batch_iterator.restore_original_order(final_states)
    return concatenate_states(reordered_states)
```

orionarcher (commented):
Thanks for the fix and enhancements @YutackPark! Just a couple comments.

torch_sim/runners.py
```diff
@@ -106,6 +107,7 @@ def integrate(
     timestep: float,
     trajectory_reporter: TrajectoryReporter | dict | None = None,
     autobatcher: BinningAutoBatcher | bool = False,
+    verbose: bool = False,
```
janosh: how about we make the `verbose` keyword a bit more versatile by accepting both `bool` and `dict`? here's an example from [`matbench_discovery/structure/symmetry.py`](https://github.com/janosh/matbench-discovery/blob/3a0c87dd0b8f57181bd07e585fbd2db88925f4be/matbench_discovery/structure/symmetry.py#L15-L43) of what it could look like:

```py
if pbar:
    pbar_kwargs = pbar if isinstance(pbar, dict) else {}
    pbar_kwargs.setdefault("desc", "Analyzing symmetry")
    iterator = tqdm(iterator, total=len(structures), **pbar_kwargs)
```

janosh (changes requested):

torch_sim/runners.py
YutackPark: I prefer this solution, but `static` returns `props` not `states` making it a little bit complex or inefficient to apply the `restore_original_order` function. 

I have two ideas.

1. Make the `restore_original_order` more flexible

2. Make `static` runner also returns `states`. (like other runner functions, but it is breaking change)

torch_sim/runners.py
YutackPark: Solved with renaming the argument `verbose` to `pbar` as it does nothing but progress bar & more reasonable naming for this fix

torch_sim/runners.py
YutackPark: I tried, but a naive fix causes the progress bar to end before reaching 100% due to remaining `converged_states` at the end (like this line: `all_converged_states.extend(result[1])`). 
Fixing this properly would make the code longer and lose its original intent, so I decided to keep it unchanged.

torch_sim/runners.py
janosh: > Make static runner also returns states. (like other runner functions, but it is breaking change)

@orionarcher and I talked about this while writing `ts.static`. we decided to go with `props` at the time but that may have been a mistake. having a consistent API across related functions (`integrate`, `optimize`, `static`) is probably more important than the convenience of extracting `props` from `states` instead of letting users do that. i need to look at some usage examples to see how much ergonomics are affected but i think i'd go with option 2

torch_sim/runners.py
orionarcher: Ahh good catch, makes sense. Maybe `torch.count_nonzero(convergence_tensor, dim=0).item()` or something like that? Just trying to get away from the for loop.

torch_sim/runners.py
orionarcher: I do think having a consistent API is preferable, however it would be convenient to return the props without forcing the user to return a trajectory to access them. One approach would be to modify the `StaticState` defined inside `static` to also include the props defined in `prop_calculators`, then the returned state will have all of the desired properties. Does that make sense?


```python
    @dataclass
    class StaticState(type(state)):
        energy: torch.Tensor
        forces: torch.Tensor | None
        stress: torch.Tensor | None
```

and then later

```python

        model_outputs = model(substate)

        substate = StaticState(
            **vars(substate),
            energy=model_outputs["energy"],
            forces=model_outputs["forces"] if model.compute_forces else None,
            stress=model_outputs["stress"] if model.compute_stress else None,
        )

        props = trajectory_reporter.report(substate, 0, model=model)
        
        for prop in props:
            substate.prop = # somehow assign the properties to the state before returning it.

        all_props.extend(props)
```

That would allow the use of the `restore_batch_order`, create a consistent api, and allow returning the desired props without the trajectory reporter. If that's too much for this PR, we could also leave as is and create an issue to solve this later.

YutackPark: I think it is straightforward to do so and similar to what I thought initially after seeing other runner functions. Not difficult task at all, I will commit the patch soon!

YutackPark: Just updated. Code is a little bit more complex than my first guess, but I believe this is the best without touching other APIs.
Below is for debugging. I have not fixed & updated unit tests yet. I think it would be better after this change is reviewed. 

```python
  import ase.build
  import torch_sim as ts

  import numpy as np

  from sevenn.calculator import SevenNetCalculator

  systems = [ase.build.bulk("Cu") * (3, 3, 3)] * 100
  rattled = []
  for i, sys in enumerate(systems):
      rng = np.random.RandomState(i)
      rng.normal(scale=0.002, size=sys.positions.shape)
      sys.set_positions(sys.positions + rng.normal(scale=0.001, size=sys.positions.shape))
      rattled.append(sys.copy())

  calc = SevenNetCalculator()

  ref_sys = rattled[63]
  ref_sys.calc = calc
  ref_val = ref_sys.get_potential_energy()

  ts_model = ts.models.SevenNetModel(calc.model)

  # with autobatcher
  final_state = ts.static(
      system=rattled,
      model=ts_model,
      autobatcher=ts.BinningAutoBatcher(ts_model, max_memory_scaler=50000),
      pbar=True
  )
  # Pass, if ordered & potential_energy field is avail
  assert abs(final_state[63].potential_energy.item() - ref_val) < 1e-4

  # w/o autobatcher
  final_state = ts.static(
      system=rattled,
      model=ts_model,
      autobatcher=False,
      pbar=True
  )

  assert abs(final_state[63].potential_energy.item() - ref_val) < 1e-4
  ```

File: torch_sim/runners.py [hunk 4]
```diff
@@ -123,6 +124,8 @@ def integrate(
             tracking trajectory. If a dict, will be passed to the TrajectoryReporter
             constructor.
         autobatcher (BinningAutoBatcher | bool): Optional autobatcher to use
+        pbar (bool | dict[str, str | float | bool], optional): Whether to show progress
+            bar. Defaults to True.
```
orionarcher: Could you note this only works if the autobatcher is provided?

torch_sim/runners.py
```diff
@@ -464,14 +491,34 @@ def static(
...
+        prop_key
+        for prop_key in next(iter(trajectory_reporter.prop_calculators.values()))
+        if prop_key not in ["energy", "forces", "stress"]
+    ]
+    # dynamically extend class fields for extra properties
+    StaticState = make_dataclass(
+        cls_name="StaticState",
+        fields=[(prop_key, torch.Tensor, None) for prop_key in extra_prop_keys],
+        bases=(_StaticState,),
+    )
```
orionarcher: Could name `extra_prop_keys ` -> `prop_keys` and then just add "energy", "forces", and "stress" if they aren't present, that'd get away from the need to define the `_StaticState`.

torch_sim/runners.py
```diff
@@ -490,15 +537,17 @@ class StaticState(type(state)):
         )
 
         props = trajectory_reporter.report(substate, 0, model=model)
-        all_props.extend(props)
+
+        for prop_key in extra_prop_keys:
+            setattr(substate, prop_key, torch.concat([prop[prop_key] for prop in props]))
```
orionarcher: I am realizing this is actually going to be a bit tricky, `torch.concat` is only appropriate for atomwise properties while batchwise properties should be stacked. More in review comment.

orionarcher (commented):
I am realizing this is actually going to be a bit tricky because we need to dynamically determine whether the computed props are batchwise, atomwise, or global. See [here](https://radical-ai.github.io/torch-sim/tutorials/state_tutorial.html#A-Simple-State) for a bit more detail. Solutions are:

1. Dynamically infer batchwise, atomwise, and global with `ts.state.infer_property_scope` and treat each differently when creating the new state. This would allow us to get the desired API but would introduce a bit of overhead, complicate the code, and open up error surface area. 

2. Abort and leave the API as is.

Curious what @janosh and @YutackPark think?

YutackPark:
> Curious what @janosh and @YutackPark think?

I chose option 2. The current complexity is already close to my personal threshold. In my opinion, the following logic:
> Dynamically infer batchwise, atomwise, and global with ts.state.infer_property_scope and treat each differently when creating the new state. This would allow us to get the desired API but would introduce a bit of overhead, complicate the code, and open up error surface area.

should not be placed in `static`. (Although I like this feature: support of the arbitrary property addition.) 
I'm not sure how this issue relates to other runners (integrate, optimize), but if they encounter similar problems, a refactor might be justified.

torch_sim/runners.py
YutackPark: I have changed the docs & defaults to `False` (Idk why I pushed with default `True`).

orionarcher:
> I chose option 2. The current complexity is already close to my personal threshold. In my opinion, the following logic:

Great, makes sense to me. Still a great addition to the codebase. I am happy to merge whenever tests are passing. Could go back to `[661a3dd](https://github.com/Radical-AI/torch-sim/pull/181/commits/661a3dded3727358dab2007588354c7da3f05be7)` and lint and be pretty much there.

Apologies for not seeing how complex this would be sooner.

YutackPark:
@orionarcher 
Reverted changes. Now it passes the test except lint. Few additional notes.

1) Linting fail: torch_sim/runners.py:324:5: C901 `optimize` is too complex (11 > 10)
This is about there exists too many arguments for the `optimize` function, because of addition of `pbar` argument

2) I have checked `test_runners` about `static`. It already has tests ensuring correct orderings. The reason the bug wasn't detected is simply it was lucky. Previous code reorders in an incorrect way, but somehow it passes the test case because the test case was too simple (only shuffles three states). I think it can be compromised as the bug is fixed and, although the test is not perfect but it still doing the right thing.

janosh (approved):
thanks @YutackPark! 👍

PR #183, opened 2025-04-28
Remove metatensor example from CI until dependency resolution issue has solution

CompRhys:
## Summary

Do not run the metatensor example in CI. This is temporary due to it not working with how we're handling dependencies via uv run for the examples in CI. 

See: https://github.com/astral-sh/uv/issues/13173, https://github.com/Radical-AI/torch-sim/issues/178

## Checklist
* [ ] All linting and tests pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,8 +135,8 @@ jobs:
       - name: Find example scripts
         id: set-matrix
         run: |
-          # Find all example scripts but exclude known failing ones
-          EXAMPLES=$(find examples -name "*.py" | jq -R -s -c 'split("\n")[:-1]')
+          # Find all example scripts but exclude the metatensor tutorial
+          EXAMPLES=$(find examples -name "*.py" -not -path "examples/tutorials/metatensor_tutorial.py" | jq -R -s -c 'split("\n")[:-1]')
           echo "examples=$EXAMPLES" >> $GITHUB_OUTPUT
 
   test-examples:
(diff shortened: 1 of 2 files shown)


janosh:
~we can still update our code to be compatible with the latest [metatensor release](https://github.com/metatensor/metatensor/releases/tag/metatensor-learn-v0.3.2), no?~

edit: nvm, hadn't read https://github.com/Radical-AI/torch-sim/issues/178 yet

Luthaf:
So updating the code to run with the latest versions should now be fine! This might require an update to metatrain as well.

It will likely fail again the next time there is a new PyTorch release though, between the PyTorch release and the metatensor-release (typically a couple of days) CI might be broken if using `uv`.

Overall I'll let you decide what you prefer!

janosh:
`uv` is very quick to fix things. by the time the next pytorch release comes around, https://github.com/astral-sh/uv/issues/13173 should be long resolved. so i think we can close this PR @CompRhys?

CompRhys:
This fix didn't work so I guess this can be closed but getting in something that stops the test suite reporting failure when we know that it's not a torch-sim issue seems like something where we should get a hotfix in soon rather than waiting an unknown period for a true one

janosh:
maybe i'm missing something but if we install latest `pytorch` 2.7 and latest `metatensor` in CI, errors should be gone?

CompRhys:
The latest versions are allowed by the pins we currently have so afaict it is already trying to do that

Luthaf:
The failure on main CI and here (`ValueError: path '/var/folders/7g/cnfrzbd555q9lmpsr1byngs80000gn/T/tmp1h3cwb2g' is not a valid checkpoint for the pet architecture`) should now be fixed: there was a change in the implementation of the PET-MAD model that required a change to the checkpoint on HuggingFace. If you restart the CI everything should pass now!

PR #185, opened 2025-04-30
io Optional Dependencies in pyproject.toml

curtischong:
## Summary

* Moved the ase dependency out of the optional [test] dependency to a required dependency because it doesn't seem like a test. Here is an example of it being used in the trajectory file https://github.com/Radical-AI/torch-sim/blob/main/torch_sim/trajectory.py#L866

## Checklist

* [ ] All linting and tests pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 ]
 requires-python = ">=3.11"
 dependencies = [
+    "ase>=3.24",
     "h5py>=3.12.1",
     "numpy>=1.26",
     "tables>=3.10.2",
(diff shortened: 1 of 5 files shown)


CompRhys:
ASE is only used for ASE specific IO and the same pattern is used for all IO to other package formats (https://github.com/Radical-AI/torch-sim/blob/main/torch_sim/io.py). This is also a common pattern used in many material science packages to not have IO to everything else under the sun as a core dependency.

CompRhys:
@curtischong Can you sign the CLA with email associated to github and then can merge this?

PR #187, opened 2025-05-02
Fix Fairchem import possibly failing silently until it reaches setup_imports()

janosh:
fixes https://github.com/Radical-AI/torch-sim/issues/184

### DIFF
--- a/torch_sim/models/fairchem.py
+++ b/torch_sim/models/fairchem.py
@@ -17,7 +17,9 @@ Notes:
 from __future__ import annotations
 
 import copy
+import traceback
 import typing
+import warnings
 from types import MappingProxyType
 from typing import Any
 
@@ -38,7 +40,8 @@ try:
     from fairchem.core.models.model_registry import model_name_to_local_file
     from torch_geometric.data import Batch, Data
 
-except ImportError:
+except ImportError as exc:
+    warnings.warn(f"FairChem import failed: {traceback.format_exc()}", stacklevel=2)
 
     class FairChemModel(torch.nn.Module, ModelInterface):
         """FairChem model wrapper for torch_sim.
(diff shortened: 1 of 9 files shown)


janosh:
@zaporter-work could you check if this solves the `NameError: name 'setup_imports' is not defined` you encountered?

orionarcher:
While fairchem has surfaced the bug, the logic is the same for all the model interfaces, so if we indent everything here, we should do it everywhere. I am not a big fan of the indent everything fix, but I don't have any better ideas.

janosh:
> I am not a big fan of the indent everything fix, but I don't have any better ideas.

you could remove the model here and instead port it upstream into [`fairchem`](https://github.com/facebookresearch/fairchem) which i like a lot and believe should be done with the other models as well (into their respective repos). pinging @zulissimeta and @lbluque to see if you'd be open to having a `torch_sim.models.ModelInterface` subclass in `fairchem`?

lbluque:
> > I am not a big fan of the indent everything fix, but I don't have any better ideas.
> 
> you could remove the model here and instead port it upstream into [`fairchem`](https://github.com/facebookresearch/fairchem) which i like a lot and believe should be done with the other models as well (into their respective repos). pinging @zulissimeta and @lbluque to see if you'd be open to having a `torch_sim.models.ModelInterface` subclass in `fairchem`?

Hi @janosh and @orionarcher 👋  we're in the process of doing a codebase refactor, but this is surely something we'd be interested in looking into the few weeks after.

torch_sim/models/fairchem.py
```diff
@@ -38,355 +38,359 @@
...
-            if key in self._reshaped_props:
-                _pred = _pred.reshape(self._reshaped_props.get(key)).squeeze()
-            results[key] = _pred.detach()
+except ImportError as exc:
 
-        results["energy"] = results["energy"].squeeze(dim=1)
-        if results.get("stress") is not None and len(results["stress"].shape) == 2:
-            results["stress"] = results["stress"].unsqueeze(dim=0)
-        return results
+    class FairChemModel(torch.nn.Module, ModelInterface):
```
orionarcher: What if instead we throw a warning if this class is defined? That could allow for an easier time catching the bug without indenting the whole file.

CompRhys (approved):
LGTM, it doesn't fully address the issue as it will still take a while to raise the error but definitely an improvement.

PR #188, opened 2025-05-11
fix installing conflicting optional dependencies via uv in pyproject.toml

curtischong:
## Summary

When we run `uv lock` (which is ran on `uv sync`), uv will automatically try to resolve all optional dependencies at the same time. However, many of the models rely on different versions of e3nn - meaning that it's impossible to satisfy all optional dependencies simultaneously.

This PR adds a conflicts section inside the pyproject.toml so it knows which optional dependencies will conflict with each other. This enables `uv lock` to properly run without throwing an error.

## Checklist


pyproject.toml
```diff
@@ -146,3 +146,13 @@ addopts = [
...
 testpaths = ["tests"]
+
+[tool.uv]
+# make these dependencies mutually exclusive since they use incompatible e3nn versions
+# see https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies for more details
+conflicts = [
+  [{ extra = "mace" }, { extra = "mattersim" }],
+  [{ extra = "mattersim" }, { extra = "graphpes" }],
+  [{ extra = "sevenn" }, { extra = "graphpes" }],
+  [{ extra = "mace" }, { extra = "sevenn" }],
```
CompRhys, line 157: Is there anyway to automatically get these ?

```sh
(base) ➜  torch-sim git:(test-stress) uv lock
  × No solution found when resolving dependencies for split (python_full_version >= '3.12' and sys_platform == 'win32'):
  ╰─▶ Because torch-sim-atomistic[mattersim] depends on mattersim and mattersim>=1.1.0 depends on e3nn>=0.5.0, we can conclude
      that e3nn<0.5.0, mattersim!=1.0.0, torch-sim-atomistic[mattersim] are incompatible.
      And because mattersim==1.0.0 depends on e3nn==0.5.0 and graph-pes>=0.0.34 depends on e3nn==0.4.4, we can conclude that
      graph-pes>=0.0.34 and torch-sim-atomistic[mattersim] are incompatible.
      And because only the following versions of graph-pes are available:
          graph-pes<=0.0.34
          graph-pes==0.0.35
          graph-pes==0.0.36
          graph-pes==0.1.0
          graph-pes==0.1.1
      and torch-sim-atomistic[graphpes] depends on graph-pes>=0.0.34, we can conclude that torch-sim-atomistic[graphpes] and
      torch-sim-atomistic[mattersim] are incompatible.
      And because your project requires torch-sim-atomistic[graphpes] and torch-sim-atomistic[mattersim], we can conclude that
      your project's requirements are unsatisfiable.
```
 running uv lock only seems to give the first one it encounters

CompRhys (approved):

curtischong:
I tried looking around but couldn't find anything that will auto-generate conflicts for you.

If we find such a feature in the future we should add that. Although, it might be tough, since this error is useful sometimes which makes me suspect that astral may not add it (and would want you to explicitly define a conflicts section so you're aware of the conflict).

The closest thing I found was https://github.com/astral-sh/uv/pull/12005 but the PR doesn't automatically create the conflicts section - it just synthesises every transitive combination between conflicting conficts.

PR #189, opened 2025-05-11
Fix types so pylance's type checker doesn't complain

curtischong:
## Summary

* Feature 1
* Fix 1

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ coverage.xml
 
 # env
 uv.lock
+
+.vscode/
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -152,7 +152,7 @@ class SimState:
         return torch.unique(self.batch).shape[0]
 
     @property
-    def volume(self) -> torch.Tensor:
+    def volume(self) -> torch.Tensor | None:
         """Volume of the system."""
         return torch.det(self.cell) if self.pbc else None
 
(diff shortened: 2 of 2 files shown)


tests/test_state.py
```diff
@@ -26,7 +26,7 @@
 def test_infer_sim_state_property_scope(si_sim_state: SimState) -> None:
     """Test inference of property scope."""
     scope = infer_property_scope(si_sim_state)
-    assert set(scope["global"]) == {"pbc"}
```
CompRhys, line 29: here it is global because things like the PBC cannot be different for things in a batched state. `per_batch` is the misleading one here as that it the `per_graph` level of abstraction. If pylance doesn't like `global` I would agree that changing it is a good thing to do but I would think that a different name, for example `shared`, might be better. I would also consider renaming `per_batch` as I don't think that `per_batch` is clear that it's per substate within a batched simstate.

CompRhys (changes requested):
`per_graph` is the wrong label but not opposed to changing the name here to avoid linting/language server issues.

curtischong:
Closing since the per_graph change is more involved and I want to roll it back but still keep Rhys' comments. I'll put up a smaller initial PR here: https://github.com/Radical-AI/torch-sim/pull/191

I'll prob rename per_batch -> per_graph in its own isolated PR

orionarcher:
Agree with Rhys that if you open another PR I'd favor "shared" or "per-state" over "per-graph". Fine switching away from global if it's causing issues.

Alternate ideas: "atomwise" - "batchwise" - "shared". 

Could also make an Enum so it's not just strings.

PR #190, opened 2025-05-12
Test stresses in the ase consistency test.

CompRhys:
## Summary

Test stresses in ASE. This should expose issues with GPa vs Kbar if they exist.

## Checklist


CompRhys:
~~@YutackPark Could you look at the SevenNet calculator for me and check the stress units? I think internally SevenNet might be returning in units of kbar it still results in an error if I add unit conversions.~~

I think it was a sign convention although still getting non-neglible differences in some examples, these seem to be due to the off diagonal voigt terms being in a different order which doesn't make sense.

The final failure for sevennet stresses is on the benezene example where the cell is given as a zero matrix and pbc is false.

CompRhys:
@Luthaf @frostedoyster With the metatensor models here the stress always appears to be zero in the test cases

YutackPark:
@CompRhys 
Sorry for the confusion, I understand. The strange stress convention of 7net comes from historical reasons, but I have lost the chance to fix it. (If you wonder, 7net follows VASP's OUTCAR stress outputs.) You can check the reference code from the `sevenn.calculator`, which restores the Voigt convention of off-diagonal components.

https://github.com/MDIL-SNU/SevenNet/blob/16f09b94153fce588429a93077024fb706c52ec9/sevenn/calculator.py#L182-L186

CompRhys:
> @CompRhys Sorry for the confusion, I understand. The strange stress convention of 7net comes from historical reasons, but I have lost the chance to fix it. (If you wonder, 7net follows VASP's OUTCAR stress outputs.) You can check the reference code from the `sevenn.calculator`, which restores the Voigt convention of off-diagonal components.
> 
> https://github.com/MDIL-SNU/SevenNet/blob/16f09b94153fce588429a93077024fb706c52ec9/sevenn/calculator.py#L182-L186

yep I saw that and fixed, still not fully sure about the benezene disagreement that we're seeing for sevennet.

Luthaf:
> @Luthaf @frostedoyster With the metatensor models here the stress always appears to be zero in the test cases

We have a paper deadline tomorrow, we'll look into this once it is done!

CompRhys:
> > @Luthaf @frostedoyster With the metatensor models here the stress always appears to be zero in the test cases
> 
> We have a paper deadline tomorrow, we'll look into this once it is done!

thanks! good luck with the final stretch for your deadline.

PR #191, opened 2025-05-12
use `"SimState"` type rather than `Self` in state.py

curtischong:
## Summary

This fixes a few type errors I've been getting via pylance.

I've talked to Janosh about the larger typing issues in the codebase but I currently do not have time to look into those right now. So I want to merge this PR in so others don't think I'm fixing all of the other typing issues.

## Checklist

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -8,7 +8,7 @@ import copy
 import importlib
 import warnings
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Literal, Self
+from typing import TYPE_CHECKING, Literal, cast
 
 import torch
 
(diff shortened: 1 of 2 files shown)


CompRhys (approved):
LGTM, fairchem tests failures can be fixed by pinning the version in the test ci script (see: https://github.com/Radical-AI/torch-sim/pull/174/commits/1a48cb48d1ecbc5368389e751cce9f6edeb6242b)

torch_sim/state.py
```diff
@@ -184,7 +184,7 @@ def row_vector_cell(self, value: torch.Tensor) -> None:
         """
         self.cell = value.transpose(-2, -1)
 
-    def clone(self) -> Self:
+    def clone(self) -> "SimState":
```
CompRhys, line 187: I think typing as Self is correct per the pep? https://peps.python.org/pep-0673/

janosh: agree `Self` is better here (as in higher fidelity). it will correctly resolve to child classes of `SimState` like `DeformState`

https://github.com/Radical-AI/torch-sim/blob/8211e08d36b4ae73407564c78a1503f07f27b26b/tests/test_state.py#L489

curtischong: interesting. pylance gave me errors for it though. maybe ty might not complain

PR #195, opened 2025-05-14
Top-level `SimState`

janosh:
add `from torch_sim.state import SimState` in `torch_sim/__init__.py`. makes `SimState` accessible as

```py
import torch_sim as ts

ts.SimState
```

for better ergonomics. related PR (as in should maybe have been part of) https://github.com/Radical-AI/torch-sim/pull/52


CompRhys (approved):
I don't really have a strong opinion here and this isn't a breaking change as far as I understand as it adds a convenience import so LGTM.

Issue #182, opened 2025-04-27
`static` runner returns properties in incorrect order


YutackPark:
This:
https://github.com/Radical-AI/torch-sim/blob/0e33e6f06aa69e4a649195b69dc203040c9d2e3a/torch_sim/runners.py#L499-L502

should be changed to something correspond to this function:
https://github.com/Radical-AI/torch-sim/blob/0e33e6f06aa69e4a649195b69dc203040c9d2e3a/torch_sim/autobatching.py#L657

I raised PR(#181) to fix this bug. This issue was created because it silently gives the wrong result.

PR #196, opened 2025-05-14
more robust ordering checks for `ts.static`

janosh:


janosh:
@YutackPark if you have time to take a look, let me know if this addresses #182 and in particular the bullet point 2 in https://github.com/Radical-AI/torch-sim/pull/181#issuecomment-2878533282 about `test_static_with_autobatcher_and_reporting` being too simple?

CompRhys (approved):
LGTM.

The issue in #182 was addressed in #181 afaict and so this is just a more robust test. 

The other discussion from #181 about a more consistent interface for integrate, optimize and static remains unresolved.

PR #201, opened 2025-05-22
Add Batched vdot to clean up fire implementation

CompRhys:
Adds a batched vdot to torch_sim.math

### DIFF
--- a/torch_sim/math.py
+++ b/torch_sim/math.py
@@ -987,3 +987,28 @@ def matrix_log_33(
             print(msg)
         # Fall back to scipy implementation
         return matrix_log_scipy(matrix).to(sim_dtype)
+
+
+def batched_vdot(
+    x: torch.Tensor, y: torch.Tensor, batch_indices: torch.Tensor
+) -> torch.Tensor:
+    """Computes batched vdot (sum of element-wise product) for groups of vectors.
+    If is_sum_sq is True, computes sum of x_i * x_i (squared norm components).
+
+    Args:
+        x: Tensor of shape [N_total_entities, D] (e.g., forces, velocities).
+        y: Tensor of shape [N_total_entities, D]. Ignored if is_sum_sq is True.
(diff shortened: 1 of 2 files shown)


CompRhys:
@t-reents I attempted to copy the changes you added but appear to have failed to do so as I do not match your results.

t-reents:
Interesting, at first glance it indeed seems to be very similar. Out of curiosity, do you also get better results than the previous version, but different ones compared to mine, or do you not even see any improvement?

One thing that I noticed while having a quick look, you also included the transformation of the forces according to the current deformation:

https://github.com/Radical-AI/torch-sim/blob/cc94facb1c164c8def7823a464cc7711fa956d97/torch_sim/optimizers.py#L1478-L1480

However, you don't use `forces` in the subsequent steps but rather `state.forces`. For example:

https://github.com/Radical-AI/torch-sim/blob/cc94facb1c164c8def7823a464cc7711fa956d97/torch_sim/optimizers.py#L1507

and

https://github.com/Radical-AI/torch-sim/blob/cc94facb1c164c8def7823a464cc7711fa956d97/torch_sim/optimizers.py#L1527

CompRhys:
I saw results very similar to what's currently on main

CompRhys:
updating to use forces everywhere wasn't a fix sadly

t-reents:
I will have a more detailed look tomorrow, already dedicated some time for it. I'll also double once again check if there are any additional changes on my side that I didn't properly, but in principle it shouldn't be the case.

t-reents:
Already had a quick look. I can reproduce my results (on my branch, so what I shared should be complete). Moreover, when I just checkout your branch and run the script, I get the same results as on my branch, see attached screenshot. Based on this test, it seems that you reproduce my results, doesn't it?

image

CompRhys:
closing in favor of #203!

PR #199, opened 2025-05-16
Fix unit and frechet cell FIRE optimizers not rescaling atom positions after each cell update

janosh:
closes #198 

* `torch_sim/optimizers.py` had a bug discovered by @t-reents in FIRE relaxation where atomic positions were not correctly updated when the cell changed. resolved by scaling atomic positions by the difference between new and previous cell after each cell update
* new relaxation tests for `OsN2` and distorted `FCC Al` structures in `test_optimizers_vs_ase.py`, covering both Frechet and Unit Cell FIRE optimizers
* tightened existing assertion tolerances in `test_optimizers_vs_ase.py` due to improved agreement between `torch-sim` and `ASE` optimizers

in a future PR, it probably makes sense to change the handling of atom positions from absolute to fractional (unwrapped) coordinates which would mean the atom positions don't require manual scaling after cell changes

PR #203, opened 2025-05-28
Fix discrepancies between `FIRE` optimisations in `ASE` and `torch-sim` 

t-reents:
It seems that the remaining deviations could be fixed. This branch is already based on the changes in #201, I'll do the rebase once it is merged.

* I reduced the number of "inner" updates in the testing script to `1`, in order to get the precise number of convergence steps
* In the pos-only case, two small fixes were necessary that led to a wrong (or better to say different) `dt`
* The Frechet Cell optimization also required some additional changes, i.e., correctly scaling the positions with respect to the deformation gradient. See https://gitlab.com/ase/ase/-/blob/master/ase/filters.py?ref_type=heads#L440-441 and https://gitlab.com/ase/ase/-/blob/master/ase/filters.py?ref_type=heads#L405-406
* The previously (#199) introduced rescaling of the positions with respect to the new cell was not necessary (it seems that those were anyway overwritten in ASE: https://gitlab.com/ase/ase/-/blob/master/ase/filters.py?ref_type=heads#L434-441), so I removed it again
* Finally, I adapted the `generate_force_convergence_fn` function to optionally also include the `cell_forces` in the convergence check, which is done in `ASE` and therefore caused the differing numbers of convergence steps

fig1

fig2

fig3

### DIFF
--- a/torch_sim/optimizers.py
+++ b/torch_sim/optimizers.py
@@ -1410,7 +1410,7 @@ VALID_FIRE_CELL_STATES = (UnitCellFireState, FrechetCellFIREState)
 
 
 def _ase_fire_step(  # noqa: C901, PLR0915
-    state: FireState | VALID_FIRE_CELL_STATES,
+    state: FireState | UnitCellFireState | FrechetCellFIREState,
     model: torch.nn.Module,
     *,
     dt_max: torch.Tensor,
--- a/torch_sim/optimizers.py
+++ b/torch_sim/optimizers.py
@@ -1164,13 +1164,13 @@ def frechet_cell_fire(
             batch=state.batch,
             pbc=state.pbc,
             # New attributes
-            velocities=torch.zeros_like(state.positions),
+            velocities=None,
             forces=forces,
             energy=energy,
             stress=stress,
             # Cell attributes
             cell_positions=cell_positions,
-            cell_velocities=torch.zeros((n_batches, 3, 3), device=device, dtype=dtype),
+            cell_velocities=None,
             cell_forces=cell_forces,
             cell_masses=cell_masses,
             # Optimization attributes
@@ -1449,6 +1449,7 @@ def _ase_fire_step(  # noqa: C901, PLR0915
 
     if state.velocities is None:
         state.velocities = torch.zeros_like(state.positions)
+        forces = state.forces
         if is_cell_optimization:
             state.cell_velocities = torch.zeros(
                 (n_batches, 3, 3), device=device, dtype=dtype
@@ -1458,8 +1459,15 @@ def _ase_fire_step(  # noqa: C901, PLR0915
             (n_batches,), alpha_start.item(), device=device, dtype=dtype
         )
 
+        if is_cell_optimization:
+            forces = torch.bmm(
+                state.forces.unsqueeze(1), state.deform_grad()[state.batch]
+            ).squeeze(1)
+        else:
+            forces = state.forces
+
         # 1. Current power (F·v) per batch (atoms + cell)
-        batch_power = tsm.batched_vdot(state.forces, state.velocities, state.batch)
+        batch_power = tsm.batched_vdot(forces, state.velocities, state.batch)
 
         if is_cell_optimization:
             valid_states = (UnitCellFireState, FrechetCellFIREState)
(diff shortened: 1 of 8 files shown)


torch_sim/optimizers.py
```diff
@@ -1554,10 +1556,17 @@ def _ase_fire_step(  # noqa: C901, PLR0915
         old_row_vector_cell = state.row_vector_cell.clone()
 
     dr_scaling_atom = torch.sqrt(dr_scaling_batch)[state.batch].unsqueeze(-1)
+
     dr_atom = torch.where(
         dr_scaling_atom > max_step, max_step * dr_atom / (dr_scaling_atom + eps), dr_atom
     )
-    state.positions = state.positions + dr_atom
+
+    if is_cell_optimization:
```
CompRhys, line 1558: this seems strange to me? does this logic not add the cartesian dr to the fractional space positions?

torch_sim/optimizers.py
```diff
@@ -1590,11 +1599,10 @@ def _ase_fire_step(  # noqa: C901, PLR0915
         transform_matrix_batch = torch.bmm(
             inv_old_cell_batch, current_new_row_vector_cell
         )
-        atom_specific_transform = transform_matrix_batch[state.batch]
-        scaled_positions = torch.bmm(
-            state.positions.unsqueeze(1), atom_specific_transform
-        ).squeeze(1)
-        state.positions = scaled_positions
+
```
CompRhys, line 1589: should lines 1596-1602 not be deleted?

torch_sim/optimizers.py
CompRhys:
```python
        positions = (
            torch.linalg.solve(
                cur_deform_grad[state.batch], (state.positions+dr_atom).unsqueeze(-1)
            ).squeeze(-1)
        )
        state.positions = torch.bmm(
            positions.unsqueeze(1), F_new[state.batch].transpose(-2, -1)
        ).squeeze(1)
```

why would this not be the right thing to do in terms of trying to match your operations here?

t-reents:
I would have to look into the failing test in more detail. As a first step, I took the `rattled_sio2` and manually inspected the whole relaxation trajectory (I did the same for the other debugging as well, i.e., a bunch of print statements comparing all the tensors etc. during relaxation). The final optimized structure as well as the intermediate states seem to agree very well. 

Still, the test is failing, so I'd have to understand what exactly is happening there. Nonetheless, based on my previous comment, those deviations might be related to the logic of the test and not necessarily to actual differences in the optimizations. I'll look into it and give an update here.


**Update:** I have to stop for now, but there is indeed a problem with the logic of the tests/the loop logic in `torch_sim.runners.optimize`. I'll push a fix tomorrow or Monday at the lastest. The failing of the tests for some of the examples is related to a different number of iterations that are performed, while for others it's the reinitialization (the init function is always called again, e.g. resetting the velocities to 0, while ASE keeps the previous state). When adjusting those aspects, I the test does pass locally.

torch_sim/optimizers.py
t-reents: > this seems strange to me? does this logic not add the cartesian dr to the fractional space positions?

No, it doesn't scale between fractional/cartesian coordinates. Based on the comments in ASE, this step is done to consider the positions in the original cell. The `dr_atom` is applied to those coordinates and the new cell deformation is applied afterwards. This is also briefly mentioned in the documentation of the `UnitCellFilter` (https://wiki.fysik.dtu.dk/ase/ase/filters.html#ase.filters.UnitCellFilter):

```
Degrees of freedom are the positions in the original undeformed cell, plus the 
deformation tensor (extra 3 “atoms”). This gives forces consistent with numerical 
derivatives of the potential energy with respect to the cell degreees of freedom.
```

> why would this not be the right thing to do in terms of trying to match your operations here?

Again, to my understanding, you want to consider the positions in the undeformed cell and apply the `dr_atom` to those. That's why I apply the `linalg.solve` first and then add the `dr_atom`, and not the other way around.

t-reents:
The test is now passing, at least locally. As mentioned earlier, one of the issues was the reinitializations between the different steps that are tested. To fix it, I added an if statement for now that checks if the state is already of the correct type.

Moreover, as described in the commit message, the current logic of the autobatching caused a different number of iterations, since the structures are set as converged when the maximum number of swap attempts is reached and not necessarily the maximum number of iterations (which led to different number of steps compared to ASE).

CompRhys:
@t-reents thank you for your careful work here!

PR #211, opened 2025-06-11
Fairchem v2 support

### LINKED ISSUE #193: Fix fairchem model interface given v2 release.
Ci for fairchem model broke after https://github.com/facebookresearch/fairchem/releases/tag/fairchem_core-2.0.0. We will pin in CI to 1.10 stop the error. 

The interface should be updated to work with v2 release.

janosh:
closes #193

Enables running the UMA class of models with `torch-sim`.

Props to the fairchem team. Our internal implementation is now simpler. In particular, I dropped

- deprecated imports: `load_config`, `update_config`, `model_registry`
- no longer needed imports: `available_models`, `local_cache`

## Summary by CodeRabbit

* New Features
  * Updated FairChem integration to UMA predictor with task selection and pretrained model names.
  * Supports energy, forces, and optional stress outputs; exposes device and dtype properties.
  * Improved batching for multiple systems.

* Refactor
  * Streamlined API by removing legacy config/checkpoint parameters; simpler initialization.
  * Example script updated to use the new interface and newer fairchem-core.

* Tests
  * Replaced legacy calculator tests with UMA-based coverage; gated by HuggingFace authentication.

* Chores
  * CI now installs dependencies from PyPI and conditionally authenticates with HuggingFace.


.github/workflows/test.yml
```diff
@@ -98,24 +90,16 @@ jobs:
...
-          fi
-          if [ -f fairchem-repo/packages/requirements.txt ]; then
-            uv pip install -r fairchem-repo/packages/requirements.txt --system
-          fi
-          if [ -f fairchem-repo/packages/requirements-optional.txt ]; then
-            uv pip install -r fairchem-repo/packages/requirements-optional.txt --system
-          fi
-          uv pip install -e fairchem-repo/packages/fairchem-core[dev] --system
+          uv pip install "torch>2" --index-url https://download.pytorch.org/whl/cpu --system
+          uv pip install "torch-scatter" -f https://data.pyg.org/whl/torch-2.6.0+cpu.html --system
```
rayg1234: you shouldnt need torch-scatter and torch-sparse anymore, torch needs to be >=2.6

torch_sim/models/fairchem.py
```diff
@@ -63,12 +56,6 @@ def __init__(self, err: ImportError = exc, *_args: Any, **_kwargs: Any) -> None:
...
     from torch_sim.typing import StateDict
 
-_DTYPE_DICT = {
-    torch.float16: "float16",
-    torch.float32: "float32",
-    torch.float64: "float64",
-}
-
 
 class FairChemModel(torch.nn.Module, ModelInterface):
```
rayg1234: just a note that the task name is only specific to UMA models, might be safer to name this something like FairChemUMAModel for future proof (or make the task name optional if we dont want to specialize just for the UMA model here)

janosh: good catch, thanks for letting me know. i made it optional in [0db53df](https://github.com/Radical-AI/torch-sim/pull/211/commits/0db53df9c3c7d424c9c8ed6e67132ab246933ab3). i think it's best if `FairChemModel` can handle both UMA and earlier models

levineds:
What's the status on this? Does this work?

AdeeshKolluru:
Given the longer timelines of [this](https://github.com/facebookresearch/fairchem/pull/1288) PR merging, I think its the best if we can have two model classes defined for v1 and current. @janosh @CompRhys thoughts?

CompRhys:
Afaik there are a few people using this branch happily and so I think we can just wait until it can land properly rather than creating additional maintenance workload. 

What is the expected timeframe and source of delay?

CompRhys:
Merge for proper credit assignment in advance of #264

PR #212, opened 2025-07-04
Fix memory scaling in `determine_max_batch_size`

t-reents:
The current version results in an infinite loop when `scale_factor < 1.5` due to the rounding. 

This is fixed by increasing the batch size by at least `+1`.

## Summary by CodeRabbit

* **Refactor**
  * Improved batch size scaling logic to ensure consistent increases and prevent potential infinite loops or stagnation.
  * Updated the default scaling factor for batch size determination to provide more efficient batch size progression.
* **Bug Fixes**
  * Added tests to verify batch size calculation avoids infinite loops with small scaling factors.
* **Chores**
  * Updated dependency specifications by removing optional features from package declarations in example scripts.

### DIFF
--- a/docs/_static/draw_pkg_treemap.py
+++ b/docs/_static/draw_pkg_treemap.py
@@ -6,6 +6,7 @@ Run with `uv run docs/_static/draw_pkg_treemap.py`
 # /// script
 # dependencies = [
 #     "pymatviz @ git+https://github.com/janosh/pymatviz",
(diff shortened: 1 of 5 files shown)


janosh (approved):
thanks @t-reents for catching and fixing! 👍

janosh:
CI failures due to `plotly` breaking change https://github.com/plotly/plotly.py/issues/5253#issuecomment-3016615635, workaround in c734c00

bmaranville:
Fixes for pymatviz compatibility with `plotly>=6.2.0` are in https://github.com/janosh/pymatviz/pull/310

PR #219, opened 2025-07-17
Fix concatenation of states in `InFlightAutoBatcher`

t-reents:
The `velocities` and `cell_velocities` are initialized to `None` in the `(FrechetCell)FIREState`. However, when using the `InFlightAutoBatcher` during an optimization, the current and new states are concatenated in `torch_sim.state.concatenate_states`. 

https://github.com/Radical-AI/torch-sim/blob/317985c731170aad578673ebe69a9334f5abe5be/torch_sim/state.py#L902-L909

When trying to merge states that were already processed for a few iterations (i.e., `velocities` are not `None` anymore) and newly initialized ones, an error is raised because the code tries to merge a `Tensor` with a `None`.

Here, I initialize the `(cell_)velocities` as tensors full of `nan` instead, so that one can merge already processed and newly initialized states. During the first initialization, the `fire` methods look for `nan` rows and replace them with zeros.

The error would also have been caught by the existing `test_autobatching` test if the number of iterations between swaps had been set to a smaller value (I changed it here). The reason is simply that within the current 10 iterations all states converge so that a completely new set of states is selected in the next iteration and already processed and newly initialised states never get merged.

You can either change the test as done here or use the following code snippet to reproduce the error with the current version:

<details>
<summary>Code to reproduce</summary>

```py
import torch_sim as ts
import torch
from torch_sim.autobatching import calculate_memory_scaler
from ase.build import bulk
from torch_sim.autobatching import estimate_max_memory_scaler


si_atoms = bulk("Si", "fcc", a=3.26, cubic=True)
si_atoms.rattle(0.05)

cu_atoms = bulk("Cu", "fcc", a=5.26, cubic=True)
cu_atoms.rattle(0.5)

many_cu_atoms = [si_atoms] * 5 + [cu_atoms] * 20

print(f"Using device: {device}")

state = ts.initialize_state(many_cu_atoms, device=device, dtype=torch.float64)

mace = mace_mp(model="small", return_raw_model=True)
mace_model = MaceModel(model=mace, device=device)

fire_init, fire_update = ts.optimizers.fire(mace_model)
fire_state = fire_init(state)

batcher = ts.InFlightAutoBatcher(
    model=mace_model,
    memory_scales_with="n_atoms",
    max_memory_scaler=40,
    max_iterations=10000,  # Optional: maximum convergence attempts per state
)

batcher.load_states(fire_state)

convergence_fn = ts.generate_force_convergence_fn(5e-3, include_cell_forces=False)

all_converged_states, convergence_tensor = [], None
while (result := batcher.next_batch(fire_state, convergence_tensor))[0] is not None:
    fire_state, converged_states = result
    all_converged_states.extend(converged_states)

    for _ in range(3):
        fire_state = fire_update(fire_state)

    convergence_tensor = convergence_fn(fire_state, None)
    print(f"Convergence tensor: {convergence_tensor}")
    print(f"Convergence tensor: {batcher.current_idx}")

else:
    all_converged_states.extend(result[1])

final_states = batcher.restore_original_order(all_converged_states)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of velocity initialization to ensure more robust and consistent behavior during optimization steps.

* **Tests**
  * Updated a test to reduce the number of simulation state updates for improved efficiency.

### DIFF
--- a/torch_sim/optimizers.py
+++ b/torch_sim/optimizers.py
@@ -589,7 +589,9 @@ def fire(
             atomic_numbers=state.atomic_numbers.clone(),
             system_idx=state.system_idx.clone(),
             pbc=state.pbc,
-            velocities=None,
+            velocities=torch.full(
+                state.positions.shape, torch.nan, device=device, dtype=dtype
+            ),
             forces=forces,
             energy=energy,
             # Optimization attributes
@@ -862,13 +864,17 @@ def unit_cell_fire(
             atomic_numbers=state.atomic_numbers.clone(),
             system_idx=state.system_idx.clone(),
             pbc=state.pbc,
-            velocities=None,
+            velocities=torch.full(
+                state.positions.shape, torch.nan, device=device, dtype=dtype
+            ),
             forces=forces,
             energy=energy,
             stress=stress,
             # Cell attributes
             cell_positions=torch.zeros(n_systems, 3, 3, device=device, dtype=dtype),
-            cell_velocities=None,
+            cell_velocities=torch.full(
+                cell_forces.shape, torch.nan, device=device, dtype=dtype
+            ),
             cell_forces=cell_forces,
             cell_masses=cell_masses,
             # Optimization attributes
@@ -1161,13 +1167,17 @@ def frechet_cell_fire(
             atomic_numbers=state.atomic_numbers,
             system_idx=state.system_idx,
             pbc=state.pbc,
-            velocities=None,
+            velocities=torch.full(
+                state.positions.shape, torch.nan, device=device, dtype=dtype
+            ),
             forces=forces,
             energy=energy,
             stress=stress,
             # Cell attributes
             cell_positions=cell_positions,
-            cell_velocities=None,
+            cell_velocities=torch.full(
+                cell_forces.shape, torch.nan, device=device, dtype=dtype
+            ),
             cell_forces=cell_forces,
             cell_masses=cell_masses,
             # Optimization attributes
@@ -1244,15 +1254,19 @@ def _vv_fire_step(  # noqa: C901, PLR0915
     dtype = state.positions.dtype
     deform_grad_new: torch.Tensor | None = None
 
-    if state.velocities is None:
-        state.velocities = torch.zeros_like(state.positions)
+    nan_velocities = state.velocities.isnan().any(dim=1)
+    if nan_velocities.any():
+        state.velocities[nan_velocities] = torch.zeros_like(
+            state.positions[nan_velocities]
+        )
         if is_cell_optimization:
             if not isinstance(state, AnyFireCellState):
                 raise ValueError(
                     f"Cell optimization requires one of {get_args(AnyFireCellState)}."
                 )
-            state.cell_velocities = torch.zeros(
-                (n_systems, 3, 3), device=device, dtype=dtype
+            nan_cell_velocities = state.cell_velocities.isnan().any(dim=(1, 2))
+            state.cell_velocities[nan_cell_velocities] = torch.zeros_like(
+                state.cell_positions[nan_cell_velocities]
             )
 
     alpha_start_system = torch.full(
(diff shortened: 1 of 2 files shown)


kianpu34593:
I just encountered the same issue. This fix is what I was trying to do as well. Glad you jumped on it. I added an improvement that is related to this ( #222 ).

curtischong:
Thank you for the contribution! I'm personally not too familiar with this part of the code, but I'll bring this up with the rest of the team.

curtischong:
I read your code and I think it works, but I'm slightly hesitant to merge it in because it overrides the meaning of NaN (and makes it fail silently if a user were to inadvertently pass in NaN tensors.. I feel like the real problem is that we didn't properly architect it to handle these cases which I'm trying to fix in my free time.

torch_sim/optimizers.py
```diff
@@ -862,13 +864,17 @@ def fire_init(
             atomic_numbers=state.atomic_numbers.clone(),
             system_idx=state.system_idx.clone(),
             pbc=state.pbc,
-            velocities=None,
+            velocities=torch.full(
+                state.positions.shape, torch.nan, device=device, dtype=dtype
+            ),
```
orionarcher, line 869: Why not just initialize to zeros?

t-reents:
**TL;DR:**
- My implementation needs to be adapted anyway, but we should agree on how to do it first
- We have to rethink the logic of the ASE fire step when dealing with states that get merged (i.e., systems are at different iterations) --> Systems at iteration 0 have to treated differently
- Agree on a way to indicate the iteration 0

Thanks for your comments!

>  I'm slightly hesitant to merge it in because it overrides the meaning of NaN (and makes it fail silently if a user were to inadvertently pass in NaN tensors

I totally agree with this and I also wasn't 100% happy with it when implementing it (however, it seemed to be the best option within the current setup). I'm happy to discuss it together to come up with a better/more consistent solution

> It seems like it would be simpler to just change the fire_init to initialize to zeros instead of nan since that's what's being set in the _vv_fire_step.  Do the different step functions initialize things differently? Am I missing something here?

This was the case in the previous version. The initialization as `None` instead of `0` was one of the changes that we did to achieve agreement with the ASE implementation, see #203.
It is necessary to skip the `else` block during the first iteration (see the ASE implementation for reference: https://gitlab.com/ase/ase/-/blob/master/ase/optimize/fire.py?ref_type=heads#L170-205):
https://github.com/Radical-AI/torch-sim/blob/317985c731170aad578673ebe69a9334f5abe5be/torch_sim/optimizers.py#L1476

**In any case**, while writing this explanation I realize that my change is not fully correct as well. If we merge states (some at more advanced iterations, some at the first iteration), we will skip the else block for all systems and therefore don't perform the correct update for systems at more advanced iterations. This being said, the logic has to be split into two branches, one for the systems at iteration 0 and one for the others. 


Just a spontaneous idea without thinking too much about it, so there might be other disadvantages, what about having a `optimization_step` attribute on the different optimization states (i.e., a tensor containing a counter for each system). In that case, one could just check if this counter is 0 for a given system. This could also help to address the following comment:
https://github.com/Radical-AI/torch-sim/blob/317985c731170aad578673ebe69a9334f5abe5be/torch_sim/runners.py#L457-L460

CompRhys:
Is the answer here not to use a zero-dim tensor? @t-reents 

```python
>>> zero_dim_tensor = torch.empty((10, 3, 0), dtype=torch.float32)
>>> zero_dim_tensor
tensor([], size=(10, 3, 0))
```

that makes the most sense to me rather than worrying about a default value. The question would then be does the `zero_dim_tensor` cause any logic issues elsewhere?

t-reents:
@CompRhys Thanks! I was actually thinking (maybe even trying to use it) about `torch.empty` as well, don't remember why I decided not to use it.

**EDIT:** 
> The question would then be does the zero_dim_tensor cause any logic issues elsewhere?

I think that it would fail again when trying to concatenate with other states:
https://github.com/Radical-AI/torch-sim/blob/317985c731170aad578673ebe69a9334f5abe5be/torch_sim/state.py#L902-L909


Moreover, I still think that the issue remains that the current logic wouldn't not work for those "mixed" states, would it? I think this is independent of how we initialize. 
So `_ase_fire_step` probably needs a small refactoring of the logic in any case, if I'm not mistaken.

orionarcher:
> what about having a optimization_step attribute on the different optimization states

I think this makes sense if the optimization has different behavior at different optimization steps, as it sounds like the `ase_fire` optimization does. It's a good solution for handling the branched logic otherwise, having `None` or `empty` be a special marker for the first step is less explicit. 

I wouldn't advocate including an optimization step for other states unless required though.

PR #227, opened 2025-07-26
Classify all variables of a SimState as per-node, per-system, and global features

curtischong:
## Summary

This PR makes handling SimStates much simpler. Rather than guessing if an attribute is per-node/system/global, we just know. My solution is to use a dictionary to store **ALL** of a State's attributes:

```python
node_features: dict[str, torch.Tensor]
system_features: dict[str, torch.Tensor]
global_features: dict[str, torch.Tensor]
```

- Even states like cell/pbc/system_index etc. are stored inside these dictionaries. By not handling exceptions we make iterating through these properties much simpler.

For ease of access to these "standard" properties, I've added custom getters/setters for these properties:
```python
@property
def positions(self) -> torch.Tensor:
    return self.node_features["positions"]

@positions.setter
def positions(self, positions: torch.Tensor) -> None:
    self.node_features["positions"] = positions
```

## Checklist

### DIFF
--- a/examples/tutorials/state_tutorial.py
+++ b/examples/tutorials/state_tutorial.py
@@ -78,17 +78,7 @@ SimState attributes fall into three categories: atomwise, batchwise, and global.
 * Batchwise attributes are tensors with shape (n_systems, ...), this is just `cell` for
   the base SimState. Names are singular.
 * Global attributes have any other shape or type, just `pbc` here. Names are singular.
-
-You can use the `infer_property_scope` function to analyze a state's properties. This
-is mostly used internally but can be useful for debugging.
 """
-
-# %%
-from torch_sim.state import infer_property_scope
-
-scope = infer_property_scope(si_state)
-print(scope)
-
 
 # %% [markdown]
 """
@@ -256,12 +246,7 @@ md_state = MDState(
     energy=torch.zeros((si_state.n_systems,), device=si_state.device),  # Initial 0 energy
 )
 
-print("MDState properties:")
-scope = infer_property_scope(md_state)
-print("Global properties:", scope["global"])
-print("Per-atom properties:", scope["per_atom"])
-print("Per-system properties:", scope["per_system"])
-
+print(md_state)
 
 # %% [markdown]
 """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ test = [
     "pymatgen>=2024.11.3",
     "pytest-cov>=6",
     "pytest>=8",
+    "pytest-xdist>=3.8.0"
 ]
 io = ["ase>=3.24", "phonopy>=2.37.0", "pymatgen>=2024.11.3"]
 mace = ["mace-torch>=0.3.12"]
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -13,7 +13,6 @@ from torch_sim.state import (
     _pop_states,
     _slice_state,
     concatenate_states,
-    infer_property_scope,
     initialize_state,
 )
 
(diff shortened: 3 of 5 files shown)


orionarcher:
HI @curtischong, this is something I thought about a fair amount and I am happy to revisit. I am not convinced I made the right decision to make everything implicit. I have a few thoughts here:

1. I did consider making the atom, batch, and global features explicit in the SimState. The advantage is obvious: it's more explicit and we no longer have to call `infer_property_scope`. The disadvantage is more subtle: it adds unneeded bloat and complicates the definition of every single State that inherits from SimState. In practice `infer_property_scope` is pretty cheap and fails infrequently.
2. If we did want to make the distinction explicit, I'd advocate for just using a tuple of strings instead of a dict. These are immutable and contain the same information. Then we could just rename `infer_property_scope` -> `return_property_scope` and have it return the tuples directly instead of inferring them. Something like this:

```
class SimState:
    positions: torch.Tensor
    masses: torch.Tensor
    cell: torch.Tensor
    pbc: bool  # TODO: do all calculators support mixed pbc?
    atomic_numbers: torch.Tensor
    system_idx: torch.Tensor | None = field(default=None, kw_only=True)
    _atom_features: tuple[str] = ("positions", "masses", "atomic_numbers")
    _system_features: tuple[str] = ("cell", "system_idx")
    _global_features: tuple[str] = ("pbc")
    
    
    @property
    def atom_features(self) -> torch.Tensor:
        return self._atom_features

    @property
    def atom_features(self) -> torch.Tensor:
        return self._atom_features

    @property
    def atom_features(self) -> torch.Tensor:
        return self._atom_features

    def return_property_scope(self):
        return {"global": self.global_features, "per_atom": self.atom_features", "per_system": self.system_features}
```

3. I think adding setters and getters for every attribute is way too bloated. That would have to be done for every single `State`. What does it add?

4. If we think this is the right option, let's use `atom` instead of `node`.

curtischong:
Thank you for your response Orion. The main reason why I'm doing this is because I'm trying to get type safety in torchsim. Having type safety can catch many bugs which is why it's important.

If we do not explicitly define the attributes we cannot guarantee type safely. (e.g. when we call `getattr` here https://github.com/Radical-AI/torch-sim/blob/main/torch_sim/state.py#L100 the types are not enforced.

Like you said, by removing `infer_property_scope`, we no longer hit the edge cases that can crash torch sim. I believe that the orb models are still failing because of something related to infer_property_scope

I agree that manually typing out the getters/setters is ugly as it covers many lines. I'll research to see if I can make it simpler. But having getters/setters does have more benefits:
- We can guarantee type safety when variables are accessed / set
- We can run code when they set attributes (e.g. warn the user if they initialize velocities with NaN)
- The bloat is inside the library, the user doesn't see this.

I agree with the `atoms` > `node` definition if this does go in

orionarcher (commented):
Generally, I am a big fan of typing and would support implementing `ty` for static type analysis. Runtime type checking, however, will add both code complexity and (a tiny bit) of computational cost. 

Even if we decided that was worth it (maybe it is), I am not sure consolidating the variables into three attributes is the best approach. It makes it a bit less readable and removes the assurance that all necessary attributes are defined. It would also make the autocomplete engines less reliable at inferring what attributes are valid. We could instead:
1. leave all the attributes as is, no need for getters
2. run a method in the __post_init__ that adds setters for every attribute that check for shape and type.

curtischong:
I agree. The three variable thing isn't very nice. I think your suggestion works. 

```python
   _atom_features: tuple[str] = ("positions", "masses", "atomic_numbers")
    _system_features: tuple[str] = ("cell", "system_idx")
    _global_features: tuple[str] = ("pbc")
```

curtischong:
closing in favor of https://github.com/Radical-AI/torch-sim/pull/228

PR #228, opened 2025-07-27
Define attribute scopes in SimStates

curtischong:
## Summary

This PR makes handling SimStates much simpler. Rather than guessing if an attribute is per-node/system/global (via `infer_property_scope`, SimState (and its child classes) explicitly define which the scope of each attribute.

#227 was a precursor to this PR

By explicitly defining a property scope to each state we:
1) Add documentation to each attribute - teaching readers which attributes is per-atom/system/global
2) Prevent `n_atoms (1) and n_systems (1) are equal, which means shapes cannot be inferred unambiguously` errors which can happen when systems have a single atom.

In general I think we should prevent these edge cases from occurring because they are uncommon but happen often enough.

I added validation logic inside the `SimState` class to ensure that each child class explicitly defines the property scope for all its attributes.

Note: this PR introduces breaking changes. In particular, it makes all of the classes that derive from SimState be kw_only. This is because the `_system_attributes` of the parent class are defined for all of the attributes:
e.g. `_system_attributes = {"reference_cell"}` so all child classes that inherit from this, put their arguments AFTER the parent's attributes. But since we've assigned a value to `_system_attributes` all future attributes (i.e. the ones in the child class) must be kwonly.

I think this breaking change is fine because by using keywords when defining complex objects like `FireState`, it eliminates the chance of a user from mixing up different `torch.Tensor` constructor params.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced explicit per-atom/per-system/global attribute scopes and added a per-atom tracked attribute "last_permutation".
  * Constructors now return NaN-filled tensors instead of None for omitted model outputs (e.g., forces/stress).

* **Bug Fixes**
  * Enforced that every state attribute is declared in exactly one scope and that tensor attributes cannot be None.

* **Documentation**
  * Tutorials updated to use the new scope-based attribute inspection API.

* **Tests**
  * Added/updated tests for scope inspection and enforcement errors; API rename reflected in tests.


examples/tutorials/state_tutorial.py
```diff
@@ -78,18 +78,8 @@
...
-You can use the `infer_property_scope` function to analyze a state's properties. This
-is mostly used internally but can be useful for debugging.
 """
 
-# %%
-from torch_sim.state import infer_property_scope
-
-scope = infer_property_scope(si_state)
-print(scope)
-
```
orionarcher, line 91: Can we mentnion how to list the state atom, system, and global state properties? We should mention both the tuples and the new generator function.

torch_sim/state.py
```diff
@@ -372,6 +378,33 @@ def __getitem__(self, system_indices: int | list[int] | slice | torch.Tensor) ->
 
         return _slice_state(self, system_indices)
 
+    def __init_subclass__(cls, **kwargs) -> None:
+        """Enforce all of child classes's attributes are specified in _atom_attributes,
+        _system_attributes, or _global_attributes.
+        """
```
orionarcher: We should document the new requirements somewhere

curtischong:
the tests are failing because of the issue pointed out in this PR: https://github.com/Radical-AI/torch-sim/pull/219. I'll revisit this tomorrow

curtischong:
:'( I just got a `ValueError: n_atoms (1) and n_systems (1) are equal, which means shapes cannot be inferred unambiguously.` error. we should probably merge this in

pyproject.toml
```diff
@@ -44,6 +44,7 @@ test = [
     "pymatgen>=2024.11.3",
     "pytest-cov>=6",
     "pytest>=8",
+    "pytest-xdist>=3.8.0",
```
curtischong: this was a nice package to install while testing this PR

curtischong:
@coderabbitai review

curtischong:
@coderabbitai revie

torch_sim/state.py
```diff
@@ -83,6 +85,10 @@ class SimState:
     atomic_numbers: torch.Tensor
     system_idx: torch.Tensor
 
+    _atom_attributes = ("positions", "masses", "atomic_numbers", "system_idx")
+    _system_attributes = ("cell",)
+    _global_attributes = ("pbc",)
```
orionarcher: I know I originally suggested tuples but maybe `set` is the right data structure here? Fine either way just a thought

orionarcher (approved):
Thanks @curtischong I think this is a solid change that reduces potentially dangerous ambiguity and makes the class definitions more readable!

torch_sim/state.py
curtischong: I agree. a set makes more sense. It also makes my validation logic simpler.

torch_sim/optimizers.py
```diff
@@ -33,6 +33,28 @@
 MdFlavor = Literal["vv_fire", "ase_fire"]
 vv_fire_key, ase_fire_key = get_args(MdFlavor)
 
+md_atom_attributes = (*SimState._atom_attributes, "forces", "velocities")  # noqa: SLF001
```
orionarcher: Can we just redefine this in integrators and optimizers instead of importing it in integrators? Feels cleaner as a private method. I don't like creating a dependency between integrators and optimizers

pyproject.toml
orionarcher: xdist is great, unless we are actually including in the testing CI I'd leave it out though.

orionarcher (changes requested):
Just a couple little things I spotted

torch_sim/optimizers.py
```diff
@@ -619,7 +672,7 @@ def fire_init(
     return fire_init, functools.partial(step_func, **step_func_kwargs)
 
 
-@dataclass
+@dataclass(kw_only=True)
```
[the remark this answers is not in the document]
curtischong: I tested it and it's not an issue. I don't think it's an issue because DeformGradMixin is a kw_only dataclass

torch_sim/optimizers.py
```diff
@@ -1261,7 +1321,7 @@ def _vv_fire_step(  # noqa: C901, PLR0915
             state.positions[nan_velocities]
         )
         if is_cell_optimization:
-            if not isinstance(state, AnyFireCellState):
+            if not isinstance(state, get_args(AnyFireCellState)):
```
curtischong, line 1323: without get_args this isinstance check is wrong I believe (since AnyFireCellState is a union type, we need to convert AnyFireCellState to a tuple of its individual types). coderabbit caught this which is why I put it in this PR

torch_sim/state.py
```diff
@@ -498,8 +498,9 @@ class DeformGradMixin:
     """Mixin for states that support deformation gradients."""
 
     reference_cell: torch.Tensor
+    row_vector_cell: torch.Tensor
 
-    _system_attributes: ClassVar[set[str]] = {"reference_cell"}
+    _system_attributes: ClassVar[set[str]] = {"reference_cell", "row_vector_cell"}
```
CompRhys: I took this out because it breaks the logic. `row_vector_cell` is a property not an attribute.

curtischong: oh I see. I thought I accidentally deleted it. I'll revert my downstream commits trying to add it back in

curtischong: the problem is that the `deform_grad()` calls `self.row_vector_cell` below. so we'll probably need to pass it in as a parameter now

curtischong: nvm I missed the fact that row_vector_cell is a simstate property.

orionarcher:
Since this is breaking let's plan to merge this when it's done and then release a v0.3.0 along with the recent bug fix?

curtischong:
oh I am finished with it. I just wanted to make the tests pass. ~~I'll bump the version in this PR though~~ nvm I've already bumped to 0.3.0 in when we renamed batch to system

curtischong:
just so you're aware if you submit a PR review and "request for changes" no other person can approve it and have it merged. So we need your approval to move forward with this!

orionarcher (approved):

orionarcher:
> no other person can approve it and have it merged

Oop my bad I thought it could be overridden 😅

Approved now, LGTM and thank you!

PR #229, opened 2025-08-01
Fix concatenation of states in `InFlightAutoBatcher`

curtischong:
see https://github.com/Radical-AI/torch-sim/pull/219
## Summary

This is actually kinda a serious issue and I'll outline it here in a clear manner.

MD SimStates often track `velocity`. But on the first iteration, the states do NOT have velocity - so they are currently initialized as `none`.

But once the optimizer gets going, these states end up having a `velocity` attribute.

The problem is how we concatenate SimStates. Inside the autobatcher, when some SimStates finish before others, we swap those finished states with fresh states. This means inside the entire SimState, we have some systems with velocity set to `none` (since they were just swapped in and are fresh) and other systems with a set velocity.

When we concatenate these "mixed" SimStates (during the optimization process), we do `torch.concatenate([torch.Tensor, none, none])`. Where the first system's velocity exists (because it's a torch.Tensor, and the last 2 systems do NOT have a velocity - since they were just swapped in by the autobatcher.

PyTmorch cannot concatenate this because we're passing in `none` as an input which is invalid.

@t-reents 's solution works pretty well and is valid (which is why I'm touching it up in this PR). His solution is: "rather than initializing `vector` attributes as `none`, we initiliaze it as `nan` so we can do torch.concatentate between states that are old, and states that have just been swapped in.

* Feature 1
* Fix 1

## Checklist

### DIFF
--- a/examples/scripts/reproduce_err.py
+++ b/examples/scripts/reproduce_err.py
@@ -1,0 +1,53 @@
+import torch
+from ase.build import bulk
+from mace.calculators.foundations_models import mace_mp
+
+import torch_sim as ts
+from torch_sim.models.mace import MaceModel
+
+
+si_atoms = bulk("Si", "fcc", a=3.26, cubic=True)
+si_atoms.rattle(0.05)
+
+cu_atoms = bulk("Cu", "fcc", a=5.26, cubic=True)
+cu_atoms.rattle(0.5)
+
+many_cu_atoms = [si_atoms] * 5 + [cu_atoms] * 20
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+print(f"Using device: {device}")
+
+state = ts.initialize_state(many_cu_atoms, device=device, dtype=torch.float64)
+
+mace = mace_mp(model="small", return_raw_model=True)
+mace_model = MaceModel(model=mace, device=device)
+
+fire_init, fire_update = ts.optimizers.fire(mace_model)
+fire_state = fire_init(state)
+
+batcher = ts.InFlightAutoBatcher(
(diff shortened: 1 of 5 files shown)


examples/scripts/reproduce_err.py
CompRhys, line 1: maybe we should put this in tests and make separate unit and integration dirs. Whilst the example scripts are tested this doesn't particularly make much sense to me in this location

torch_sim/optimizers.py
```diff
@@ -587,9 +587,11 @@ def fire_init(
             masses=state.masses.clone(),
             cell=state.cell.clone(),
             atomic_numbers=state.atomic_numbers.clone(),
-            system_idx=state.system_idx.clone(),
+            system_index=state.system_idx.clone(),
```
CompRhys: `system_index` vs `system_idx` still looks inconsistently applied? see line 863 below

curtischong: yes. this was a "hack" I was trying last night. but it will not work because of other parts of the code. I will have this renamed which makes it easier to see inconsistencies. Thanks for pointing out the inconsistency though

examples/scripts/reproduce_err.py
curtischong: I'll remove this file when this PR is out of draft. it's pure testing code rn

curtischong:
for visibility I have thought of a good way to resolve this problem and will visit it hopefully later tonight. We'll definitely have a good PR by Saturday

curtischong:
closing since this PR is superseded by https://github.com/Radical-AI/torch-sim/pull/232

PR #230, opened 2025-08-02
InitVar dataclass initialization (and subclass checks)

curtischong:
## Summary

In order to properly type `SimState`, I want to remove the `| None` type annotation on `system_idx`. This is solved via the `InitVar` variable `init_system_idx`.

In addition to typing, this PR introduces the `__post_init__` check for subclasses of `SimState`. This paradigm allows us to verify properties of derived classes. In particular, this PR uses `__post_init__` to verify that all `InitVar` classes start with the `init_` prefix (which is required to properly mangle params during concatenation / splitting of `SimStates`.

In the next PR, we will use the `__post_init__` to enforce that all derived `SimState` classes cannot have a `| None` attribute inside (which will break functions like `torch.concatenate` since we cannot concat attributes that are tensors and attributes that are none - See the description of https://github.com/Radical-AI/torch-sim/pull/229 for more info)

This PR is breaking since we modify the constructor for SimState.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved handling of initialization variables and system indices for simulation state objects.
  * Enforced stricter validation rules on subclassing and tensor attribute definitions.
  * Unified and streamlined state object construction to ensure correct initialization and prevent errors with optional tensors.
  * Updated optimizer state classes to require velocity attributes to always be present.

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -6,6 +6,7 @@ operations and conversion to/from various atomistic formats.
 
 import copy
 import importlib
+import typing
 import warnings
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Literal, Self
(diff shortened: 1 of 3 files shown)


torch_sim/state.py
```diff
@@ -107,24 +110,25 @@ def __post_init__(self) -> None:
                 f"masses {shapes[1]}, atomic_numbers {shapes[2]}"
             )
 
-        if self.cell.ndim != 3 and self.system_idx is None:
```
curtischong, line 110: I moved these checks down since it depends on self.system_idx (also to bundle it with the self.cell.shape[0] check)

torch_sim/state.py
```diff
@@ -272,7 +276,7 @@ def clone(self) -> Self:
             else:
                 attrs[attr_name] = copy.deepcopy(attr_value)
 
-        return self.__class__(**attrs)
+        return construct_state(self, attrs)
```
curtischong: since we now have InitVar params in the constructor, creating a class is not as easy as `__class__(**attrs)`. We need to properly handle the `init_` constructor params

curtischong:
closing since I don't like this approach. it doesn't feel right to call initialization params `init_<attribute>` in the constructor

PR #232, opened 2025-08-02
Fix simstate concatenation

curtischong:
## Summary

* Feature 1
* Fix 1

## Checklist


tests/test_autobatching.py
```diff
@@ -489,14 +489,68 @@ def convergence_fn(state: ts.SimState) -> bool:
...
-        # run 10 steps, arbitrary number
+        # run 10 steps (so all states converge before the next batch)
         for _ in range(10):
             state = fire_update(state)
         convergence_tensor = convergence_fn(state)
 
     assert len(all_completed_states) == len(fire_states)
 
 
+def test_in_flight_with_fire_only_converge_some_states(
```
curtischong: I don't like how I just copied over the above test and changed 1 line. Any advice on how make it more DRY?

t-reents: If I see it correctly, you only changed the number of optimization steps before (potentially) updating the batch, right?

In that case, you could just replace the 5/10 steps by a variable and add the `pytest.mark.parametrize` decorator to test both values of this variable with the same test function.

curtischong: fair. I'll add a comment explaining the purposes of both.

curtischong:
I think this PR will not totally fix the issues. This is because inside `_split_state` we have this function:

```python
def split_attr(
    attr_value: torch.Tensor | None, split_sizes: list[int]
) -> list[torch.Tensor | None]:
    return (
        [None] * n_systems
        if attr_value is None
        else torch.split(attr_value, split_sizes)
    )
```

CompRhys (approved):
LGTM. When merging I will insert @t-reents email so he gets shared credit on the PR

curtischong:
I would like to add more tests for the split and concatenate states.

curtischong:
> LGTM. When merging I will insert @t-reents email so he gets shared credit on the PR

we should just merge his PR first. then this can be a follow-up PR which adds the extra checks.

t-reents:
Thanks guys for acknowledging my PR and thanks to @curtischong for your additional work on top of it!

curtischong:
Since a few ppl have already seen this PR it's probably best to add the extra tests in another PR. I'll merge this in.

PR #241, opened 2025-08-30
Basic Pyright checks for state.py

curtischong:
## Summary

* Adds pyright with a bunch of rules checked off. The goal is to solve the actually important type issues (e.g. assuming that a variable is always `torch.Tensor` (and it'll never be `None`). I'll make a few PRs but this is just the first. I tried to get state.py a bit typed just to start a conversation.
* I added `overload`s for `get_attrs_for_scope` so when ppl are using them, they will know the types that are returned. This is NOT a design decision because already; all atom and system attributes must be `torch.Tensor` whereas all global attributes can be Any. This must be enforced because when we split sim_states, we use `torch.split` on those per-atom and per-system attributes.

## Checklist


torch_sim/state.py
```diff
@@ -606,14 +602,25 @@ def state_to_device(
         if isinstance(attr_value, torch.Tensor):
             attrs[attr_name] = attr_value.to(device=device)
 
-    if dtype is not None:
```
curtischong, line 609: we assign dtype, so this can never be None

torch_sim/state.py
```diff
@@ -194,11 +194,7 @@ def n_atoms(self) -> int:
     @property
     def n_atoms_per_system(self) -> torch.Tensor:
         """Number of atoms per system."""
-        return (
-            self.system_idx.bincount()
```
curtischong, line 198: system_idx will never be none

curtischong:
closing because I feel like mypy has much better pytorch support. tinygrad AND pytorch use mypy. I feel like pyright is better for more generic python software but not for this.

PR #242, opened 2025-08-30
mypy type math.py and test_math.py

curtischong:
## Summary

I've typed math.py and test_math.py. This is the first mypy typing PR. In general I think that Mypy is good for catching bugs without forcing us to change the code to fit the typechecker.

The mypy.ini file is the same one that tinygrad uses

## Checklist

## Summary by CodeRabbit

- New Features
  - Added expm_frechet_with_matrix_exp to compute both matrix exponential and its Frechet derivative in one call.
  - Expanded NumPy input support for expm_frechet, expm_frechet_kronform, and expm_cond.

- Refactor
  - Streamlined API: expm_frechet now returns only the derivative and no longer accepts the compute_expm parameter.
  - Updated internal call sites and autograd path to the new API.

- Tests
  - Updated test suite to cover both supported methods via the new API.

- Chores
  - Introduced MyPy configuration for stricter type checking.


torch_sim/math.py
```diff
@@ -1016,7 +1019,7 @@
     if batch_indices.min() < 0:
         raise ValueError("batch_indices must be non-negative")
 
-    output = torch.zeros(batch_indices.max() + 1, dtype=x.dtype, device=x.device)
+    output = torch.zeros(batch_indices.max() + 1, dtype=x.dtype, device=x.device) # type: ignore
     output.scatter_add_(dim=0, index=batch_indices, src=(x * y).sum(dim=1))
```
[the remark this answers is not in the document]
curtischong: this calls .item which synchronizes the thread which is slow

torch_sim/math.py
```diff
@@ -29,11 +29,35 @@ def torch_divmod(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.
...
+        method: str, optional. Choice of algorithm. Should be one of
+            - `SPS` (default)
+            - `blockEnlarge`
+        check_finite: bool, optional. Whether to check that the input matrix contains
+            only finite numbers. Disabling may give a performance gain, but may result
+            in problems (crashes, non-termination) if the inputs do contain
+            infinities or NaNs.
+
+    Returns:
+        ndarray. Frechet derivative of the matrix exponential of A in the direction E.
```
CompRhys: typing is this docstring is off. Suggests it's numpy but return type in code is Tensor

torch_sim/math.py
```diff
@@ -44,34 +68,22 @@ def expm_frechet(  # noqa: C901
...
 
     Returns:
-        If compute_expm is True:
-            expm_A: ndarray. Matrix exponential of A.
-            expm_frechet_AE: ndarray. Frechet derivative of the matrix exponential of A
-                in the direction E.
-        Otherwise:
-            expm_frechet_AE: ndarray. Frechet derivative of the matrix exponential of A
-                in the direction E.
+        expm_A: ndarray. Matrix exponential of A.
```
CompRhys: again ndarray when typed tensor?

torch_sim/math.py
```diff
@@ -450,10 +456,6 @@ def expm_cond(A: torch.Tensor, check_finite: bool = True) -> torch.Tensor:
     if check_finite and not torch.isfinite(A).all():
         raise ValueError("Matrix A contains non-finite values")
 
-    # Convert input to torch tensor if it isn't already
-    if not isinstance(A, torch.Tensor):
-        A = torch.tensor(A, dtype=torch.float64)
-
```
CompRhys: this is fine to have even if typing says it's unreachable code?

torch_sim/math.py
```diff
@@ -504,9 +506,7 @@ def backward(ctx: Any, grad_output: torch.Tensor) -> torch.Tensor:
         (A,) = ctx.saved_tensors
 
         # Compute the Frechet derivative in the direction of grad_output
-        return expm_frechet(
-            A, grad_output, method="SPS", compute_expm=False, check_finite=False
-        )
+        return expm_frechet(A, grad_output, method="SPS", check_finite=False)
```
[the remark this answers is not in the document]
CompRhys: Can we check if this is true and if so add a unit test?

torch_sim/math.py
```diff
@@ -773,7 +773,11 @@ def _matrix_log_case2b(
 
 
 def _matrix_log_case3(
-    T: torch.Tensor, lambda_val: complex, mu: complex, nu: complex, num_tol: float = 1e-16
+    T: torch.Tensor,
+    lambda_val: torch.Tensor,
+    mu: torch.Tensor,
+    nu: torch.Tensor,
+    num_tol: float = 1e-16,
 ) -> torch.Tensor:
```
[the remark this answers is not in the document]
CompRhys: slop comment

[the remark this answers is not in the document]
CompRhys: there was no double comparison in the original code only in your hallucinations?

torch_sim/math.py
curtischong: Yeah I was thinking about this. Maybe we should have the numpy checks further up the pipeline (since math.py feels like more like an internal than an external api)? The idea is that by moving it up we can better trust the types. It is annoying that by removing this we could be introducing a breaking change if ppl didn't adhere to the type of the function.

curtischong: I'll change the types to be | numpy array for now. then we can change this functionality in a future PR if desired. this way the changes are not breaking

torch_sim/math.py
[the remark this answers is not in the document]
curtischong: I tried it and not returning a tuple is fine. https://chatgpt.com/share/68b8ede4-5628-800e-ae38-db7056df865c

Their docs don't mention that it's a tuple: https://docs.pytorch.org/docs/stable/generated/torch.autograd.Function.backward.html?utm_source=chatgpt.com

However, this chatGPT thread was so adamant (despite it running an example of it themselves and noticing that there's no error). It's definitely a problem in their pretraining. https://chatgpt.com/share/68b8ee2b-c424-800e-81b5-621d7e75ee01.

CompRhys (approved):

mypy.ini
CompRhys, line 1: can this not live in pyproject.toml ?

curtischong: if you prefer it to be there I can move it

CompRhys: fewer random files is cleaner

PR #243, opened 2025-08-30
Type other files

curtischong:
## Summary

* Feature 1
* Fix 1

## Checklist


tests/test_io.py
curtischong, line 1: types ignored because it's pointless. that's not what we're testing

torch_sim/neighbors.py
```diff
@@ -174,7 +174,7 @@ def primitive_neighbor_list(  # noqa: C901, PLR0915
                 bin_index_ic[:, c], n_bins_c[c]
             )
         else:
-            bin_index_ic[:, c] = torch.clip(bin_index_ic[:, c], 0, n_bins_c[c] - 1)
+            bin_index_ic[:, c] = torch.clip(bin_index_ic[:, c], 0, n_bins_c[c] - 1) # type: ignore[call-overload]
```
curtischong: these types are ignored because satisfying them would be inefficient. we want to keep the variables as torch tensors, rather than using `.item()` as that would synchronize the computation thread (which is slow). Also I believe that pytorch's own types aren't powerful enough to address this

torch_sim/transforms.py
```diff
@@ -375,8 +375,9 @@ def wrap_positions(
 
     # Convert center to tensor
     if not hasattr(center, "__len__"):
-        center = (center,) * 3
-    center = torch.tensor(center, dtype=positions.dtype, device=device)
+        center_pos = torch.tensor((center,) * 3, dtype=positions.dtype, device=device)
+    else:
+        center_pos = torch.tensor(center, dtype=positions.dtype, device=device)
```
CompRhys: This feels like a vague way to check this? maybe an improvement to do `isinstance(x, float)` and `isinstance(x, tuple) and len(x)==3 and all(isinstance(y, float) for y in x)`?

torch_sim/transforms.py
```diff
@@ -563,6 +564,15 @@ def compute_cell_shifts(
     return cell_shifts
 
 
+def compute_cell_shifts_strict(
```
CompRhys, line 563: why not call this function from inside `compute_cell_shifts`? would be more DRY

CompRhys (approved):
two nits

torch_sim/transforms.py
curtischong: I think `isinstance(x, float)` is sufficient. because center is defined as `tuple[float, float, float] | float`

PR #244, opened 2025-09-09
Type trajectory

curtischong:
## Summary

* Feature 1
* Fix 1

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Write to multiple trajectory files in a single run.
  * Accept multiple filenames and flexible path/sequence inputs.
  * Expose a unified array registry for trajectory data.

* **Improvements**
  * Trajectory writing now accepts NumPy arrays/scalars and PyTorch tensors, converting automatically as needed.
  * Consistent temperature handling on the correct device/dtype during integration.
  * Support for configurable metadata and state/trajectory options when creating trajectories.

* **Tests**
  * Added typing annotations to test data (no runtime changes).

### DIFF
--- a/.coderabbit
+++ b/.coderabbit
@@ -1,0 +1,5 @@
+reviews:
+  high_level_summary: false
+  auto_review:
+    enabled: false
+    drafts: false
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -58,7 +58,7 @@ jobs:
           path: ./docs_build
 
   deploy:
-    if: github.repository_owner == 'Radical-AI' && github.ref == 'refs/heads/main'
+    if: github.repository_owner == 'TorchSim' && github.ref == 'refs/heads/main'
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,89 +6,89 @@
 
 ### 💥 Breaking Changes
 
-* Remove higher level model imports by @CompRhys in https://github.com/Radical-AI/torch-sim/pull/179
+* Remove higher level model imports by @CompRhys in https://github.com/TorchSim/torch-sim/pull/179
 
 ### 🛠 Enhancements
 
-* Add per atom energies and stresses for batched LJ by @abhijeetgangan in https://github.com/Radical-AI/torch-sim/pull/144
-* throw error if autobatcher type is wrong by @orionarcher in https://github.com/Radical-AI/torch-sim/pull/167
+* Add per atom energies and stresses for batched LJ by @abhijeetgangan in https://github.com/TorchSim/torch-sim/pull/144
+* throw error if autobatcher type is wrong by @orionarcher in https://github.com/TorchSim/torch-sim/pull/167
 
 ### 🐛 Bug Fixes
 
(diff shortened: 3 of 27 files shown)


torch_sim/trajectory.py
```diff
@@ -455,7 +465,7 @@ def _initialize_type_map(
 
     def write_arrays(
         self,
-        data: dict[str, np.ndarray | torch.Tensor],
+        data: "Mapping[str, np.ndarray | np.generic | torch.Tensor]",
```
CompRhys, line 468: why is dict not sufficient to type this?

torch_sim/trajectory.py
```diff
@@ -970,7 +981,7 @@ def __len__(self) -> int:
         """
         return self._file.root.data.positions.shape[0]
 
-    def write_ase_trajectory(self, filename: str | pathlib.Path) -> "Trajectory":
+    def write_ase_trajectory(self, filename: str | pathlib.Path) -> "TrajectoryReader":
```
CompRhys, line 984: this feels wrong?

torch_sim/trajectory.py
```diff
@@ -44,7 +44,7 @@
 
 if TYPE_CHECKING:
     from ase import Atoms
-    from ase.io.trajectory import Trajectory
+    from ase.io.trajectory import TrajectoryReader
```
orionarcher, line 47: Why change this?

torch_sim/trajectory.py
curtischong: if you scroll to line 1015, we are opening `Trajectory` in read mode `return Trajectory(filename) # Reopen in read mode`, which is a TrajectoryReader object

torch_sim/trajectory.py
curtischong: view my comment to Rhys below

torch_sim/trajectory.py
curtischong: we run into https://mypy.readthedocs.io/en/stable/common_issues.html#variance with mypy

torch_sim/trajectory.py:784: note: "dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
torch_sim/trajectory.py:784: note: Consider using "Mapping" instead, which is covariant in the value type

curtischong: we also need the np.generic because of this error:

tests/test_trajectory.py:390: error: Argument 1 to "write_arrays" of "TorchSimTrajectory" has incompatible type "dict[str, ndarray[Any, Any] | generic | Tensor]"; expected "Mapping[str, ndarray[Any, Any] | Tensor]"  [arg-type]

which allows us to write a np.bool inside test_trajectory.py

```
scalar_data: dict[str, np.ndarray | np.generic | torch.Tensor] = {
    "float64_scalar": np.float64(1.0),
    "float32_scalar": np.float32(1.0),
    "int64_scalar": np.int64(1),
    "int32_scalar": np.int32(1),
    "bool_scalar": np.bool_(True),  # noqa: FBT003
    "torch_float_scalar": torch.tensor(1.0, dtype=torch.float64),
    "torch_int_scalar": torch.tensor(1, dtype=torch.int64),
    "torch_bool_scalar": torch.tensor(data=True),
}

traj.write_arrays(scalar_data, steps=0)
```

PR #255, opened 2025-09-17
merge vesin dependency declaration

curtischong:
## Summary

The vesin page recommends using vesin[torch] to install torchscript bindings https://luthaf.fr/vesin/latest/index.html.

This also reduces the chance of updating one version but forgetting to update the other

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Consolidated dependencies: replaced separate vesin and vesin-torch entries with a single vesin[torch] dependency (>=0.3.7).
  * Improves dependency management and compatibility with upstream releases.
  * No user-facing changes expected; existing functionality should continue to work as before.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,8 +32,7 @@ dependencies = [
     "tables>=3.10.2",
     "torch>=2",
     "tqdm>=4.67",
-    "vesin-torch>=0.3.8",
-    "vesin>=0.3.8",
+    "vesin[torch]>=0.3.7",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "tables>=3.10.2",
     "torch>=2",
     "tqdm>=4.67",
-    "vesin-torch>=0.3.7",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     "tables>=3.10.2",
     "torch>=2",
     "tqdm>=4.67",
-    "vesin-torch>=0.3.7",
+    "vesin-torch>=0.3.8",
```
CompRhys: Should also update the line below?

CompRhys: why do we need to bump this pin? won't the dependency solver just work for 2.8 now a 2.8 compatible version exists if the end user has a 2.8 pin?

curtischong: I guess the user can explicitly decide to specify a version of vesin that they desire

curtischong: on another note, I looked at https://luthaf.fr/vesin/latest/index.html and they recommend using `vesin[torch]`. Maybe we should update this to be vesin[torch]>=0.3.7 so we don't accidentally update one but not the other?

CompRhys:
merging into single fixup PR

PR #263, opened 2025-09-25
Thermodynamic Integration

### LINKED ISSUE #302: Implement Parallel Tempering
Parallel Tempering is a way to improve the coverage of a distribution, by preventing a system to be trapped in a local minimum. The idea is that it's easier to pass an energy barrier with more energy, in this case provided by a higher temperature.

So, a number N of simulations are run in parallel at temperatures T_1, ..., T_N. Every n_step, a proposal to exchange two adjacent systems with temperatures T_i, T_{i+1} is tested using Metropolis Hasting. In case of success, the systems are exchange, so that the in practice the temperatures are changed and velocities are scaled.

Doing so, all chains for a given *temperature* follow its Boltzmann distribution for the intended temperature.

The implementation in TorchSim would be fairly easy in the first approach. 

1. Choose a system to analyse
2. Create a function that duplicate into N systems
3. Choose the temperatures to consider (often logarithmic)
4. Every n_step test the exchange of adjacent systems. In case of success, of course don't change the index of systems but only the temperatures. Integrators are already able to accept different temperatures for a batched system.
5. Update velocities in case of success
6. (Most important step) save efficiently the trajectories so that it's easy to access the trajectories for all temperatures or a chosen temperature (most of the time the lowest temperature).Note: 
- this approach works both for NVT or NPT simulations.
- I suspect NVT Nosé Hoover may not work properly. Or at least maybe the thermostat (extended variable) need to be rescaled or reset. But I don't have proof for that.
- Actually the swap does not need to be between adjacent, it's often an approximation as the probability of swapping is exponentially low.

References:
- GROMACS: https://manual.gromacs.org/documentation/2024.0/reference-manual/algorithms/replica-exchange.html
-  K. Hukushima and K. Nemoto, “Exchange Monte Carlo Method and Application to Spin Glass Simulations,” J. Phys. Soc. Jpn., 65 1604–1608 (1996). -> Apply for Monte Carlo simulations, no thermostat
- Y. Sugita and Y. Okamoto, “Replica-exchange molecular dynamics method for protein folding,” Chem. Phys. Lett., 314 141–151 (1999). -> MD simulation but thermostat only relevant for perfect gaz, as far as I understood.

thomasloux:
## Summary

Adding a big feature to compute free energy for crystal phases.
1. Forward TI with Einstein reference:
   - Compute Einstein model frequencies from NVT simulation of reference model
   - Run multiple forward TI trajectories from Einstein model to target model
   - Compute free energy difference using Jarzynski equality
  Reference:
  - Jarzynski equality: <https://doi.org/10.1103/PhysRevLett.78.2690>
2. Forward-backward TI:
   - Run forward TI from model_a to model_b
   - Equilibrate at model_b
   - Run backward TI from model_b to model_a (with reverse=True)
   - Compute free energy using Jarzynski equality and adiabatic switching method
 Reference:
  - de Koning, Maurice, and A. Antonelli.
  "Adiabatic switching applied to realistic crystalline solids: Vacancy-formation
  free energy in copper." Physical Review B 55.2 (1997): 735.

Inspiration and sources:
- https://calorine.materialsmodeling.org/get_started/free_energy_tutorial.html
- Freitas, Rodrigo, Mark Asta, and Maurice De Koning.
  "Nonequilibrium free-energy calculation of solids using LAMMPS."
  Computational Materials Science 112 (2016): 333-341.

Note: I need to add the pipeline to see the results for different simulation lengths, for different temperatures and the actual evaluation to find the temperature of transition of titanium from HCP to BCC. I try to reproduce https://calorine.materialsmodeling.org/get_started/free_energy_tutorial.html
The code is a somewhat direct application of LAMMPS implementation.

What is lacking?
- It's only either Helmholtz free energy, or Gibbs free energy as Pressure=0, but normally I simply need to add the PV term in the computation.
- The computation of the cell parameters at varying temperature. In my current implementation, not in this first draft PR, I use IsotropicMTKNPT, but current TorchSim NPT should be fine as I think they are also isotropic.
- Tests: will add them later.
- A good documentation and full working example

What can evolve?
- The current reimplementation of MDState and Langevin integrator to add the energy_difference does not seem optimal I think that we should be able to add arbitrary information in MDState and possible save everything that is returned by an MLIP in MDState.

## Checklist

* [ ] I have read and agree to the Contributor's Certification in [CONTRIBUTING.md](https://github.com/TorchSim/torch-sim/pull/CONTRIBUTING.md).


orionarcher:
Thanks for this contribution! Convert out of draft when you want a review, happy to take a look.

thomasloux:
I have some incoming commits that needs Velocity / Cell rescaling to be implemented. Waiting for #326 to be merged

Issue #262, opened 2025-09-24
NVT Nose Hoover does not work for batched systems


thomasloux:
I'm looking at the nvt_nose_hoover method and I see this intriguing comment:
\# For now, sum the per-system DOF as chain expects a single int
\# This is a limitation that should be addressed in the chain implementation
total_dof = int(dof_per_system.sum().item())

This seems quite weird if one wants to run NVT MD simulation with a batched system. Actually looking at the NoseHooverChain, nothing indicates that it handles a batch system.

Running script 3.5 and defining a batch system, the code actually fails.

```
"""NVT simulation with MACE and Nose-Hoover thermostat."""

# /// script
# dependencies = [
#     "mace-torch>=0.3.12",
# ]
# ///

import os

import torch
from ase.build import bulk

import torch_sim as ts
from torch_sim.integrators.nvt import nvt_nose_hoover, nvt_nose_hoover_invariant
from orb_models.forcefield import pretrained

from torch_sim.models.orb import OrbModel
from torch_sim.quantities import calc_kT
from torch_sim.units import MetalUnits as Units
from torch_sim.state import concatenate_states


# Set device and data type
device = "cuda" if torch.cuda.is_available() else "cpu"
dtype = torch.float32

# Number of steps to run
N_steps = 20

# Create diamond cubic Silicon
si_dc = bulk("Si", "diamond", a=5.43, cubic=True).repeat((2, 2, 2))

orbff = pretrained.orb_v3_conservative_inf_omat(
        device=device,
        precision="float32-highest",
        compile=False
    )
orb_model = OrbModel(model=orbff, dtype=torch.get_default_dtype())

state = ts.io.atoms_to_state(si_dc, device=device, dtype=dtype)
state = concatenate_states([state, state])  # Create a batch of 2 systems

# Run initial inference
results = orb_model(state)

dt = 0.002 * Units.time  # Timestep (ps)
kT = (
    torch.tensor(1000, device=device, dtype=dtype) * Units.temperature
)  # Initial temperature (K)


nvt_init, nvt_update = nvt_nose_hoover(model=orb_model, kT=kT, dt=dt)
state = nvt_init(state=state, kT=kT, seed=1)

for step in range(N_steps):
    if step % 10 == 0:
        temp = (
            calc_kT(
                masses=state.masses, momenta=state.momenta, system_idx=state.system_idx
            )
            / Units.temperature
        )
        invariant = nvt_nose_hoover_invariant(state, kT=kT)
[...]

Issue #247, opened 2025-09-12
Half supercell error when using OrbModel with batched inputs


thomasloux:
OrbModel wraps orb-models. An optimization used is half_supercell. While I'm not sure what is exactly performed, the orb repo only allows this optimization with volume of a cell > 1000 (https://github.com/orbital-materials/orb-models/blob/4c3bbdd11fbb08c95455be70aaf7057a20fd0a2d/orb_models/forcefield/featurization_utilities.py#L331). This is checked in OrbModel but it uses the maximum value, while it should ensure all crystals have a volume greater than 1000.

Minimum code to reproduce the error.

```
from ase.build import bulk
from torch_sim.models.orb import OrbModel
from torch_sim.state import initialize_state
import torch_sim as ts
from torch_sim.integrators import nvt
import torch
# ORB model imports
from orb_models.forcefield import pretrained
from orb_models.forcefield.calculator import ORBCalculator

device = "cpu"

input = bulk("Ti", "hcp", a=2.95, c=4.68) * 4
input2 = bulk("Ti", "hcp", a=2.95, c=4.68) * 3
input = initialize_state([input, input2], dtype=torch.float32, device=device)

cells = input.cell
# compute volume
print(torch.det(cells)) # tensor([2257.3582,  952.3230])

orbff = pretrained.orb_v3_conservative_inf_omat(
    device=device,
    precision="float32-highest",
    compile=False
)
orb_model = OrbModel(model=orbff, dtype=torch.get_default_dtype())

temperature = 300.0  # Kelvin
timestep = 0.001
pressure = -1
#  Run batched NPT MD
results = ts.integrate(
    system=input,
    model=orb_model,
    integrator=nvt.nvt_langevin,
    n_steps=10,
    temperature=temperature,
    timestep=timestep,
    )

```

orionarcher:
Thanks for raising @thomasloux, we'll try to get this fixed soon.

EDIT: Ah, just saw the PR, thank you!

thomasloux:
Fixed in #254

PR #265, opened 2025-09-29
Feature/batched nvt nose hoover

thomasloux:
## Summary

Add support for batch systems for NVT Nosé Hoover #262 
Add tests: batched systems, identical results for batched systems and independent single run (with same initial state, most importantly the momenta).

I've also carefully checked that the behaviour is the same as the previous non batched implementation.

## Checklist

### DIFF
--- a/tests/test_integrators.py
+++ b/tests/test_integrators.py
@@ -8,6 +8,8 @@ from torch_sim.integrators import (
     npt_langevin,
     nve,
     nvt_langevin,
+    nvt_nose_hoover,
+    nvt_nose_hoover_invariant,
 )
 from torch_sim.models.lennard_jones import LennardJonesModel
 from torch_sim.quantities import calc_kT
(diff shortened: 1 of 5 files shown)


thomasloux:
@orionarcher You can review, normally there are not big changes ! Let me know if you want me to modify some parts

abhijeetgangan:
@thomasloux Thanks for the PR. I will take a look at it once we merge the API changes.

torch_sim/integrators/nvt.py
```diff
@@ -6,7 +6,7 @@
 
 import torch
 
-from torch_sim.integrators.md import (
+from torch_sim.integrators.md_batched import (
```
abhijeetgangan: Wrong import?

thomasloux: you are right, typo from development when I'm wanted to check the consistency of batched vs unbatched. Just sent a fix

torch_sim/integrators/nvt.py
```diff
@@ -377,14 +377,14 @@ def nvt_nose_hoover_init(
...
-        n_atoms_per_system = torch.bincount(state.system_idx)
+        n_atoms_per_system = state.n_atoms_per_system
         dof_per_system = (
             n_atoms_per_system * state.positions.shape[-1]
         )  # n_atoms * n_dimensions
 
         # For now, sum the per-system DOF as chain expects a single int
         # This is a limitation that should be addressed in the chain implementation
-        total_dof = int(dof_per_system.sum().item())
+        # total_dof = int(dof_per_system.sum().item())
```
orionarcher: is this limitation still true?

torch_sim/integrators/nvt.py
```diff
@@ -430,10 +430,10 @@ def nvt_nose_hoover_update(
         chain = state.chain
 
         # Update chain masses based on target temperature
-        chain = chain_fns.update_mass(chain, kT)
+        # chain = chain_fns.update_mass(chain, kT)
```
orionarcher: Maybe remove commented code before merge

torch_sim/models/orb.py
```diff
@@ -416,7 +416,7 @@ def forward(self, state: ts.SimState | StateDict) -> dict[str, torch.Tensor]:
             if not model_has_direct_heads and prop == "stress":
                 continue
             _property = "energy" if prop == "free_energy" else prop
-            results[prop] = predictions[_property].squeeze()
+            results[prop] = predictions[_property]
```
orionarcher, line 425: Since I believe this is fixing #247 could you mention the issue in this PR?

orionarcher: nvm I just merged #247 instead.

orionarcher: It'd be helpful to understand why the current tests aren't capturing this issue. If you could write a test that fails without this change that'd be really helpful.

torch_sim/integrators/nvt.py
thomasloux: this shouldn't have been commented to be honest

torch_sim/models/orb.py
thomasloux: hey, I added a new model output shape test with only 1 system. There was a test for a batch input only.

thomasloux:
By the way, I probably broke the npt nosé hoover. So I need to adapt in the same way the integrator.

tests/models/conftest.py
```diff
@@ -228,3 +228,85 @@ def test_model_output_validation(
     # Rename the function to include the test name
     test_model_output_validation.__name__ = f"test_{model_fixture_name}_output_validation"
     return test_model_output_validation
+
+
+def make_validate_single_system_model_outputs_test(
```
CompRhys: can this single case just be added to the main model interface test factory and then can avoid changes to all the model files?

thomasloux: you're right, i've updated it!
it's way better like that.

thomasloux:
Hey @abhijeetgangan, I've merged the API change. Tests are passing. Note: we may want to add a test for we hardcode the value of a simulation with all integrators that we have to check that we don't change the results among versions.

abhijeetgangan:
Thanks a lot @thomasloux . I will go through it this weekend. For the tests; I agree that we need another check other than the stability checks. We could have numerical values of MD averages. I would also like to have integration tests comparing the values with LAMMPS. Could be a script that runs in CI with bunch of assertions.

thomasloux:
Added a correction for #266. Need a careful review for that to make sure that I did not forget something. Decided to add a condition rather then expanding xi, p_xi and Q in a similar way than LAMMPS

thomasloux:
This is ready to merge

tests/test_integrators.py
```diff
@@ -295,6 +295,399 @@ def test_nvt_langevin_multi_kt(
...
+        energies.append(state.energy)
+        temperatures.append(temp / MetalUnits.temperature)
+        invariants.append(ts.nvt_nose_hoover_invariant(state, kT))
+
+    # Convert temperatures list to tensor
+    temperatures_tensor = torch.stack(temperatures)
+    temperatures_list = [t.tolist() for t in temperatures_tensor.T]
+    assert torch.allclose(
+        temperatures_tensor[-1],
+        torch.tensor([299.9910, 299.6800], dtype=dtype),
```
abhijeetgangan, line 332: These specific temperature values correspond to some seed?

abhijeetgangan:
~~Just one comment about the specific temperature values in NPT and NVT tests. Otherwise good to merge.~~ This is to keep expected behavior over versions and changes to integrators.

PR #270, opened 2025-10-01
Fairchem legacy

CompRhys:
## Summary

Introduce legacy fairchem model class to allow continued use of EGIP and GEMNet family models whilst they're not yet implemented in fairchem v2.

## Checklist

Before a pull request can be merged, the following items must be checked:

* [ ] Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google).
* [ ] Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.
* [ ] Tests have been added for any new functionality or bug fixes.
* [ ] I have read and agree to the Contributor's Certification in [CONTRIBUTING.md](CONTRIBUTING.md).

We highly recommended installing the pre-commit hooks running in CI locally to speedup the development process. Simply run `pip install pre-commit && pre-commit install` to install the hooks which will check your code before each commit.

PR #278, opened 2025-10-07
Fixing model loading logic (names and cache dir) for fairchem models

niklashoelter:
Previously: model (str | Path | None) was passed as to fairchem's get_predict_unit - although this is expecting a model name and thereby throws an error even if a correct model path is set.

According to Fairchem, when loading a cached model, the `cache_dir` parameter needs to be set in addition to a valid `model_name`. This is fixed here.

## Summary

According to the fairchem's `pretrained_mlip.get_predict_unit` logic, I made the model_name a required parameter when initalizing a FairChemModel object and added a second, optional `model_cache_dir` string parameter that is passed to the `get_predict_unit function` if set.

As this is my first contribution, hope that I did not forget anything!

## Checklist

### DIFF
--- a/torch_sim/models/fairchem.py
+++ b/torch_sim/models/fairchem.py
@@ -84,7 +84,7 @@ class FairChemModel(ModelInterface):
         """Initialize the FairChem model.
 
         Args:
-            model (str | Path): Path to model checkpoint file
+            model (str | Path | None): Path to model checkpoint file
             neighbor_list_fn (Callable | None): Function to compute neighbor lists
                 (not currently supported)
             model_name (str | None): Name of pretrained model to load


CompRhys:
Thanks for contributing!

I would need a bit more time to dig into what's happened here but this may be because of https://github.com/TorchSim/torch-sim/pull/211 and https://github.com/TorchSim/torch-sim/pull/270 where `FairChemModel` now corresponds to the v2 fairchem library. Older fairchem models can be accessed in the farichem_legacy.py / `FairChemV1Model`. Does just swapping to using `FairChemV1Model` get the behavior you want?

niklashoelter:
@CompRhys thanks for your quick answer!
I did not check with the v1 fairchem models as I am relying on the UMA ones. The simple fix from my branch however worked for me.

CompRhys:
I think the mistake is that we need something along the lines of https://github.com/facebookresearch/fairchem/blob/dcd772e7921969643dc8ff71f2430326e2afad57/src/fairchem/core/calculate/ase_calculator.py#L132-L149. We would still want to allow users to load checkpoints from paths even if we expect majority to reuse models released by meta.

niklashoelter:
Tried to implement it as you suggested while keeping the cache dir option which is None by default.

CompRhys:
Made a few tweaks for better back compatibility and to get the tests to work.

CompRhys:
Fairchem tests fail because I believe your branch won't have access to the HF_TOKEN we have in the repo secrets. GraphPES errors are known to be flaky.

CompRhys:
Merging and will fix fairchem test issues in another PR.

PR #287, opened 2025-10-15
Create py.typed

Andrew-S-Rosen:
I added a `py.typed` marker so that libraries using torch-sim as a dependency can get everything type checked. https://peps.python.org/pep-0561/


abhijeetgangan:
@Andrew-S-Rosen Thanks for the suggestion and contribution! The mace tests failing should be fixed. Unfortunately, the fairchem tests fail because of missing hugging face token which will be fixed soon. Will merge once other tests pass.

CompRhys:
I will merge this and then resolve the build issues and fairchem test issues in another PR

PR #289, opened 2025-10-16
MACE: transfer atomic numbers to CPU before converting to numpy.

t-reents:
I get the following error when trying to use the MACE model for structural optimization (or MD):

```py
  File "/home/reents_t/project/dev-torch-sim/git/torch-sim/torch_sim/models/mace.py", line 232, in setup_from_system_idx
    atomic_numbers_to_indices(atomic_numbers.numpy(), z_table=self.z_table),
                              ^^^^^^^^^^^^^^^^^^^^^^
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

This PR fixes this error by transferring the `atomic_numbers` to the CPU first (in case the data is on the GPU, which should be the case), before converting the `tensor` to `numpy`.

I just used the example scripts or tried to follow the instructions in the `README`: https://github.com/TorchSim/torch-sim/tree/v0.4.0?tab=readme-ov-file#running-batched-relaxation

I'm not sure whether this was really a problem or whether I just did something else in a wrong way, as I'd expect that this would have been encountered before. Nonetheless, also comparing with other models, this change seems to be reasonable.

If this is not a bug, just close this PR :smile:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,6 +96,10 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v6
 
+      - name: Install HDF5 on macOS
+        if: runner.os == 'macOS'
+        run: brew install hdf5
+
       - name: Install legacy fairchem repository and dependencies
         if: ${{ matrix.model.name == 'fairchem-legacy' }}
         run: |


thomasloux:
Relevant bug, I can reproduce. 
I think it is because the pytests are run on CPU and not GPU so that the tests currently don't cover these bugs

By the way, the README example needs to be updated to new API (I added that in #290 )

t-reents:
Thanks @thomasloux for confirming!

>I think it is because the pytests are run on CPU and not GPU

Yes, I was also mainly referring to users, as I would have expected that someone has encountered this before

CompRhys:
The mace and elastic test failures here are blocking merge but in principal this is good. I can look over weekend

t-reents:
Thanks @CompRhys 
Just checked it again and the tests are passing locally. According to the message, it seems that something goes wrong during the loading process, so it might not really be related to the changes introduced in this PR.

Thanks for having a look at it!

CompRhys:
Likely uv downloads a new MACE release which is breaking for us somehow

abhijeetgangan:
The mace tests are only failing on mac OS because of the hdf5. I think there might be a easy fix. In case it is blocking we can skip them for mac.

Issue #114, opened 2025-04-05
Support constraints in the relaxation


zdcao121:
Currently, it seems the `torch-sim` does not support setting [constraints](https://wiki.fysik.dtu.dk/ase/ase/constraints.html) like `ase` in the relaxation or MD.

abhijeetgangan:
@zdcao121 Which constraints do you think would be good addition?

zdcao121:
@abhijeetgangan  fix space group symmetry in the relaxation, like [FixSymmetry](https://wiki.fysik.dtu.dk/ase/ase/constraints.html#ase.constraints.FixSymmetry) in ase.

t-reents:
Personally, I'd also find an equivalent to [FixAtoms](https://wiki.fysik.dtu.dk/ase/ase/constraints.html#the-fixatoms-class) in ASE quite useful, i.e., fixing the positions of certain atoms in a structure during relaxation.

efako:
FIxAtoms - like constraint would indeed be super useful!

RokasEl:
➕ on supporting FixAtoms. Would be super useful!

thomasloux:
Done in #294

PR #294, opened 2025-10-17
Features/constraints

thomasloux:
## Summary

First draft for constraints. Inspiration from ASE.
This is a rather big feature, I prefer to post a first mvp so that we can discuss some implementation and make sure constraints are compatible with Integrators and Optimizers.

### Codes implementing constraints
- JAX-MD: no
- ASE: yes
- LAMMPS: yes, but difficult to read (C++) and copy.
- OpenMM: yes, but only bonds lengths or angles as far as I understand
For the rest I did not check this much.

### Implementation:
- Constraints: FixAtoms, FixCom
Change in the code:
- Add an optional constraints variable in SimState
- Turn every subclass dataclass(kw_only=True) to prevent errors (positional argument before non pos argument)
- Adapt MD only, via setter (set_momenta, set_positions)

### What's working perfectly
- Contraints tests implemented
- All tests pass
- FixAtoms really fixes the atoms concerned

### Future work
- Adapt optimizer
- Make sure that FixCom is working well. The Com drift is small, but I would expect an even smaller value. ASE seems to have a smaller value from first tests. I did not yet reproduce their tests as it is based on an optimization task.
- Implement other constraints
- make sure everything is compatible with Integrators and Optimizers. Essentially we should apply RATTLE algorithm.
- Adapt the calculation of temperature to take into account the reduction of degrees of freedom.
- For FixCom,  one wants the center of mass to be fixed in unwrap coordinate. So of course, we need to check after unwrapping. And as a result, I now affect set_momenta, then apply the wrapping on the systems. Actually ASE does not wrap during the simulations, but the wrapping is applied from each calculator forward pass. I think that it is the case for most of TorchSim model implementations. So we may want to stop wrapping during the simulations. Of course if always possible to wrap afterwards.

### Implementation discussions
- I wanted first to have an hidden variable _positions and have a implicit setter and getter. But this is very painful to adapt with _atom_attributes and the copy of state sometimes performed directly accessing the variables using vars(state).  At least it is the case if I set the __init__ function to accept an `positions` argument and not a `_positions` argument. Actually I think that the setter is clearer to indicate that the constraints are imposed.
- I decided to add constraints as a global attribute, but prevent list of constraint and looping over the constraints.
- The disavantage: It's rather difficult to write the constraints and define the constraints over the batch system. For FixAtoms you need to provide the index of the atoms in the batch system. Probably we want to define actually for each system and then batch the constraint. For FixCom, we need to compute the COM in an efficient, which is not super easy to read. It will be even worse if one wants to fix the com of a subgroup in a batch systems. 

## Checklist

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -300,6 +300,31 @@ class SimState:
 
         self._constraints = constraints
 
+    def set_cell(
+        self,
+        cell: torch.Tensor,
+        scale_atoms: bool = False,  # noqa: FBT001, FBT002
+    ) -> None:
+        """Set the unit cell of the system, optionally scaling atomic positions.
+        Torch version of ASE Atoms.set_cell.
+
+        Args:
+            cell (torch.Tensor): New unit cell with shape (n_systems, 3, 3)
+            scale_atoms (bool, optional): Whether to scale atomic positions according to
+                the change in cell. Defaults to False.
+        """
+        if cell.shape != self.cell.shape:
+            raise ValueError(
+                f"New cell must have shape {self.cell.shape}, got {cell.shape}"
+            )
+        if scale_atoms:
+            M = torch.linalg.solve(self.cell.mT, cell.mT)
+            self.positions = torch.bmm(
+                self.positions.unsqueeze(1), M[self.system_idx]
+            ).squeeze(1)
+
+        self.cell = cell
+
     def get_number_of_degrees_of_freedom(self) -> torch.Tensor:
         """Calculate degrees of freedom accounting for constraints.
 
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -650,3 +650,34 @@ def test_state_to_device_no_side_effects(si_sim_state: SimState) -> None:
             "New state doesn't have correct device!"
         )
         assert si_sim_state is not new_state_gpu, "New state is not a different object!"
+
+
+def test_state_set_cell(ti_sim_state: SimState) -> None:
+    """Test the set_cell method of SimState."""
+    new_cell = (
+        torch.diag_embed(
+            torch.tensor(
+                [3.0, 4.0, 5.0], device=ti_sim_state.device, dtype=ti_sim_state.dtype
+            )
+        )
+        @ ti_sim_state.cell
+    )
+    ase_atoms = ti_sim_state.to_atoms()[0]
+    ti_sim_state.set_cell(new_cell, scale_atoms=True)
+    ase_atoms.set_cell(new_cell[0].T.cpu().numpy(), scale_atoms=True)
+    assert torch.allclose(
+        ti_sim_state.positions.cpu(), torch.from_numpy(ase_atoms.positions)
+    )
+
+    M = torch.tensor(
+        [[[1.0, 0.2, 0.0], [0.1, 1.5, 0.0], [0.0, 0.0, 2.0]]],
+        device=DEVICE,
+        dtype=ti_sim_state.dtype,
+    )
+    new_cell = M @ ti_sim_state.cell
+    ase_atoms = ti_sim_state.to_atoms()[0]
+    ti_sim_state.set_cell(new_cell, scale_atoms=True)
+    ase_atoms.set_cell(new_cell[0].T.cpu().numpy(), scale_atoms=True)
+    assert torch.allclose(
+        ti_sim_state.positions.cpu(), torch.from_numpy(ase_atoms.positions)
+    )


curtischong (commented):
This looks good!

I like how the abstract methods like adjust_positions are flexible enough for constraining specific indices/atoms and for constraining specific axes of movement (which'll probably be added later)

torch_sim/integrators/npt.py
```diff
@@ -19,7 +19,7 @@
 from torch_sim.typing import StateDict
 
 
-@dataclass
+@dataclass(kw_only=True)
```
curtischong, line 25: We'll need to tag this PR with breaking bc of this. I think this is a reasonable change though

torch_sim/constraints.py
```diff
@@ -0,0 +1,406 @@
...
+            forces: Forces to be adjusted
+        """
+
+    def copy(self) -> FixConstraint:
+        """Create a copy of this constraint.
+
+        Returns:
+            A new instance of this constraint with the same parameters
+        """
+        return type(self)(**self.__dict__)
```
orionarcher: I'd get rid of this, imo `copy.copy` should do the job

torch_sim/constraints.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def todict(self) -> dict[str, Any]:
+        """Convert constraint to dictionary representation.
+
+        Returns:
+            Dictionary representation of the constraint
+        """
+        return {"name": self.__class__.__name__, "kwargs": self.__dict__.copy()}
```
orionarcher: I'd get rid of this too. `asdict` is a built-in for `dataclasses` and users can add a name themselves if they want.

torch_sim/constraints.py
```diff
@@ -0,0 +1,406 @@
...
+from typing import TYPE_CHECKING, Any
+
+import torch
+
+
+if TYPE_CHECKING:
+    from torch_sim.state import SimState
+
+
+class FixConstraint(ABC):
```
orionarcher: If this is going to be the base class for all constraints I'd favor naming it `Constraint`

torch_sim/integrators/npt.py
orionarcher: Yeah I'm fine making state initialization kw only, you'd be crazy to do it with args

torch_sim/transforms.py
```diff
@@ -1157,3 +1157,67 @@ def safe_mask(
     """
     masked = torch.where(mask, operand, torch.zeros_like(operand))
     return torch.where(mask, fn(masked), torch.full_like(operand, placeholder))
+
+
+def unwrap_positions(
+    pos: torch.Tensor, box: torch.Tensor, system_idx: torch.Tensor
+) -> torch.Tensor:
```
orionarcher, line 1182: This is just used in the tests so I'd suggest moving it into the testing file and making it a private method. Unwrapping coordinates is quite tricky to do right, and I don't want to imply we've done it perfectly by adding it to the public API.

thomasloux: for now I can do that, but I consider that this implement should be right, at least assuming that the displacement at each step is small enough

thomasloux:
@orionarcher thanks for the feedback. I'm starting to adapt for `_split_state`, `_pop_states`, `_slice_state`, and concatenate_states. `_filter_attrs_by_mask` seems to be slightly too general. Actually as it is accepting both an `atom_mask`and `system_mask`, although actually you probably expect only to depend on `system_mask`. I would suggest to determine the atom_mask in the function. Even if it is slightly less efficient.

EDIT: or find a way to be sure that atom_mask always correspond to the system_mask

thomasloux:
Hey, I've changed quite a lot the API of the constraints. Actually I've almost reproduced the equivalent of atom_attribute or system_attribute at the level of constraints. The code is still a draft, they are probably better way reorganize the operation, especially for manipulation of states (pop, split, concatenante) which are probably difficult to read and too long.

- Now support states manipulations
- I removed wrap
- Add a fix reference in FixCom (set at the first call)
- Test for constraints for optimizers
- Modify calc_kt to support new degrees of freedom

Regarding SystemConstraint, I need to implement a default slice(None) so that the constraint can function in the case the final user directly sets constraint (state.constraint = [FixCom()]). We could accept not to support this.

Also I realize the general_attribute is only supported for now for pbc, which should change soon. We should have constraint in its own category as it's not even propagate in similar ways as the rest of the attributes. This would also allow to have a private _constraint variable, and support the previous statement not to set directly constraint.

More than open for suggestion how to best add constraints.

By the way, in the meantime, we may want to have a dedicate branch so that people can try (a lot of demand for optimization #114). It's probably better to do before having extensive tests and proofs that our optimizers and integrators are compatible with the constraints. 

Regarding calc_kt, I suggest to make it a MDState method (like in ASE). So that the user does not forget to add degrees of freedom.

I did not add for now a check of incompatible constraints for now, not sure it is relevant. For instance you may want in the future to fix the length of all atoms in a chain (think of a polymer). Then the index of an atom will appear multiple times.

janosh:
@thomasloux great work on this! i'd like to start using symmetry constrained relaxation and would like to help push this over the finish line. if you don't mind i'll push a few minor bug fixes and some more tests?

thomasloux:
Hey, no problem for me. Actually I need to modify the current implementation of FixCom. I also need to add a warning if someone uses that on NPT integrators, because I'm almost sure the current implementation is not ready (as least for the simple reason that the constraint should be rescaled with the cell).

torch_sim/constraints.py
```diff
@@ -400,21 +400,55 @@ def count_degrees_of_freedom(
...
-    """Issue warnings if constraints might overlap in problematic ways.
+def validate_constraints(  # noqa: C901
+    constraints: list[Constraint], state: SimState | None = None
+) -> None:
+    """Validate constraints for potential issues and incompatibilities.
 
-    This function checks for potential issues like multiple constraints
-    acting on the same atoms, which could lead to unexpected behavior.
+    This function checks for:
+    1. Overlapping atom indices across multiple constraints
```
thomasloux, line 554: For 2. AtomIndexedConstraints spanning multiple systems:
So actually the current FixAtoms is implemented so that it is supposed to be only one FixAtoms contraints for a batch system. So it's expected to act on multiple systems.

for 1. it's not so clear that it's a problem to have the same atoms affected by multiple constraints. Another precise example of that, take a water molecule, you often want to constraints H-bonds. Then the oxygen atom will be concerned by 2 constraints.

thomasloux:
@janosh I've made the changes I wanted to do. Now everything  is working as I wanted. Let me know if you have questions or remark on my implementation

curtischong:
I'll review this today/tomorrow. Sorry I've been a bit busy lately

orionarcher:
I'm working on a PR to this PR that will make the logic for combining and separating constraints a bit more modular. Will hopefully finish later today.

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+
+        Fix atoms using a boolean mask:
+        >>> mask = torch.tensor([True, True, True, False, False])
+        >>> constraint = FixAtoms(mask=mask)
+    """
+
+    def get_removed_dof(self, state: SimState) -> torch.Tensor:
+        """Get number of removed degrees of freedom.
+
+        Each fixed atom removes 3 degrees of freedom (x, y, z motion).
```
curtischong, line 354: It'd be nice if people that implemented this function wouldn't have to think about systems at all. they just think about how many dofs it removes per system

orionarcher:
Please see my PR on @thomasloux repo for a minor overhaul to how constraints are handled when states are split and merged: [https://github.com/thomasloux/torch-sim/pull/1/files](https://github.com/thomasloux/torch-sim/pull/1/files)

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+        Args:
+            indices: Indices of atoms to constrain. Can be a tensor or list of integers.
+
+        Raises:
+            ValueError: If both indices and mask are provided, or if indices have
+                       wrong shape/type
+        """
+        if indices is None:
+            # Empty constraint
+            self.indices = torch.empty(0, dtype=torch.long)
```
orionarcher: I'd advocate we call this atom_idx to match the syntax of System Constraint

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+        This method should modify forces in-place to ensure the constraint
+        is satisfied.
+
+        Args:
+            state: Current simulation state
+            forces: Forces to be adjusted
+        """
+
+
+class AtomIndexedConstraint(Constraint):
```
orionarcher: Maybe call this AtomConstraint to match SystemConstraint?

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+
+
+class SystemConstraint(Constraint):
+    """Base class for constraints that act on specific system indices.
+
+    This class provides common functionality for constraints that operate
+    on a subset of systems, identified by their indices.
+    """
+
+    def __init__(self, system_idx: torch.Tensor | list[int] | None = None) -> None:
```
orionarcher: Why allow None as an input / why allow empty constraints at all?

If a state does not have any constraints, I think it's better for `constraints=[]` to be the only possiblility instead of `constraints=[SystemConstraint()]`, `constraints=[AtomConstraint()]`, and `constraints=[AtomConstraint(), SystemConstraint()]` all to be equivalent.

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+            forces[index_to_consider] * state.masses[index_to_consider].unsqueeze(-1),
+        )
+        lmd /= system_square_mass.unsqueeze(-1)
+        forces_change = torch.zeros(state.n_systems, 3, dtype=dtype)
+        forces_change[self.system_idx] = lmd
+        forces -= forces_change[state.system_idx] * state.masses.unsqueeze(-1)
+
+    def __repr__(self) -> str:
+        """String representation of the constraint."""
+        return "FixCom()"
```
orionarcher: This should list the indices, right?

tests/test_constraints.py
```diff
@@ -0,0 +1,676 @@
...
+    assert torch.all(first_system.constraints[0].indices == torch.tensor([0, 1]))
+    assert torch.all(second_system.constraints[0].indices == torch.tensor([]))
+    assert torch.all(
+        concatenated_state.constraints[0].indices == torch.tensor([0, 1, 32, 33])
+    )
+
+    # Verify FixCom constraint system masks
+    assert torch.all(
+        concatenated_state.constraints[1].system_idx == torch.tensor([0, 1, 2])
+    )
```
orionarcher, line 192: Why do we expect constraint to be on all three systems here when we set `FixCom()` above? In my mind that would mean no CoM constraint.

torch_sim/state.py
```diff
@@ -226,6 +239,52 @@ def row_vector_cell(self, value: torch.Tensor) -> None:
...
+        """
+        # check it is a list
+        if isinstance(constraints, Constraint):
+            constraints = [constraints]
+        for constraint in constraints:
+            if (isinstance(constraint, SystemConstraint)) and (
+                not constraint.initialized
+            ):
+                constraint.system_idx = torch.arange(self.n_systems, device=self.device)
+                constraint.initialized = True
```
orionarcher: Can we get around the special treatment of SystemConstraint here? Do we need this `initialized` flag? What if we just eagerly initialized the constraints as soon as they are created?

In my mind it would be nice if `constraints.py` defined an API and `state.py` used that without needing any knowledge of the internals.

torch_sim/state.py
```diff
@@ -673,6 +750,35 @@ def _filter_attrs_by_mask(
...
+                    constraint.system_idx = torch.tensor(
+                        [
+                            system_idx_map[idx.item()]
+                            for idx in constraint.system_idx
+                            if system_mask[idx]
+                        ],
+                        device=constraint.system_idx.device,
+                        dtype=constraint.system_idx.dtype,
+                    )
+
```
orionarcher: This handling is hard for me to follow, I made an attempt to clean it up here: https://github.com/thomasloux/torch-sim/pull/1

Same is true for `_split_state` and `concatenate_states` but I won't comment again.

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+    def __repr__(self) -> str:
+        """String representation of the constraint."""
+        if len(self.indices) <= 10:
+            indices_str = self.indices.tolist()
+        else:
+            indices_str = f"{self.indices[:5].tolist()}...{self.indices[-5:].tolist()}"
+        return f"FixAtoms(indices={indices_str})"
+
+
+class FixCom(SystemConstraint):
```
curtischong, line 398: I think this is only valid for non PBC systems right? Maybe we should throw a warning if they're doing it on a state with pbc?

torch_sim/models/einstein.py
```diff
@@ -0,0 +1,285 @@
+"""Einstein model where each atom is treated as an independent 3D harmonic oscillator.
+
+Contrary to other models, the model energies depend on an absolute reference position,
+so the model can only be used on systems that the model was initialized with.
+As a analytical model, it can provide its Helmholtz free energy and can also generate
```
curtischong: nit: "an analytical"

torch_sim/models/einstein.py
```diff
@@ -0,0 +1,285 @@
...
+        The Boltzmann distribution for a harmonic oscillator leads to Gaussian
+        distributions
+        for both positions and velocities.
+        """
+        N = self.x0.shape[0]
+        kB = units.BaseConstant.k_B / units.UnitConversion.eV_to_J
+        beta = 1.0 / (kB * temperature)  # Inverse temperature in 1/eV
+
+        # Sample positions from a normal distribution around equilibrium positions
+        stddev = torch.sqrt(1.0 / (self.masses * (self.frequencies**2) * beta)).unsqueeze(
```
curtischong: maybe we can replace `self.masses * (self.frequencies**2)` with self.spring_constants

torch_sim/optimizers/fire.py
```diff
@@ -211,13 +212,15 @@ def _vv_fire_step[T: "FireState | CellFireState"](  # noqa: PLR0915
     state.velocities += 0.5 * atom_wise_dt * state.forces / state.masses.unsqueeze(-1)
 
     # Position update
-    state.positions = state.positions + atom_wise_dt * state.velocities
+    # state.positions = state.positions + atom_wise_dt * state.velocities
```
curtischong: we should probably remove these comments

torch_sim/optimizers/fire.py
```diff
@@ -211,13 +212,15 @@ def _vv_fire_step[T: "FireState | CellFireState"](  # noqa: PLR0915
...
+    state.set_positions(state.positions + atom_wise_dt * state.velocities)
 
     # Cell position updates are handled in the velocity update step above
 
     # Get new forces and energy
     model_output = model(state)
-    state.forces = model_output["forces"]
+    # state.forces = model_output["forces"]
```
curtischong: same here

torch_sim/optimizers/fire.py
```diff
@@ -454,16 +457,19 @@ def _ase_fire_step[T: "FireState | CellFireState"](  # noqa: C901, PLR0915
...
+            ).squeeze(1)
+        )
     else:
-        state.positions = state.positions + dr_atom
+        state.set_positions(state.positions + dr_atom)
 
     # Get new forces, energy, and stress
     model_output = model(state)
-    state.forces = model_output["forces"]
+    # state.forces = model_output["forces"]
```
curtischong: this one too

torch_sim/models/einstein.py
```diff
@@ -0,0 +1,285 @@
...
+        equilibrium_position = torch.as_tensor(
+            equilibrium_position, device=self._device, dtype=self._dtype
+        )
+        frequencies = torch.as_tensor(
+            frequencies, device=self._device, dtype=self._dtype
+        )  # [N, 3]
+
+        if frequencies.shape[0] != equilibrium_position.shape[0]:
+            raise ValueError("frequencies shape must match equilibrium_position shape")
+        if frequencies.min() < 0:
```
curtischong: The frequencies cannot be 0 since the atoms have to oscillate

torch_sim/optimizers/gradient_descent.py
```diff
@@ -107,7 +108,8 @@ def gradient_descent_step(
     atom_lr = pos_lr[state.system_idx].unsqueeze(-1)
 
     # Update atomic positions
-    state.positions = state.positions + atom_lr * state.forces
+    # state.positions = state.positions + atom_lr * state.forces
```
curtischong: another :p

torch_sim/optimizers/state.py
```diff
@@ -23,6 +23,39 @@ class OptimState(SimState):
...
+        self.masses = masses
+        self.cell = cell
+        self.pbc = pbc
+        self.atomic_numbers = atomic_numbers
+        self.system_idx = system_idx
+        self._constraints = _constraints
+        self.energy = energy
+        self.set_forces(forces)
+        self.stress = stress
+        SimState.__post_init__(self)
```
curtischong: I don't think this is needed

torch_sim/optimizers/state.py
```diff
@@ -23,6 +23,39 @@ class OptimState(SimState):
...
+        stress: torch.Tensor | None = None,
+        masses: torch.Tensor,
+        cell: torch.Tensor,
+        pbc: torch.Tensor,
+        atomic_numbers: torch.Tensor,
+        system_idx: torch.Tensor,
+        _constraints: list | None = None,
+    ) -> None:
+        """Initialize optimization state."""
+        self.positions = positions
```
curtischong: shouldn'y we do self.set_positions as well?

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+    def adjust_positions(self, state: SimState, new_positions: torch.Tensor) -> None:
+        """Adjust positions to maintain center of mass position.
+
+        Args:
+            state: Current simulation state
+            new_positions: Proposed positions to be adjusted in-place
+        """
+        dtype = state.positions.dtype
+        n_systems = (
+            state.n_systems if self.system_idx == slice(None) else len(self.system_idx)
```
curtischong: It's a bit unfortunate how `self.system_idx` is a different meaning from SimState.system_idx

curtischong: I do think it's fine since ik we were thinking of changing system_idx to a sparse tensor due to performance

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+
+        Fixing center of mass removes 3 degrees of freedom (x, y, z translation).
+
+        Args:
+            state: Simulation state
+
+        Returns:
+            Always returns 3 (center of mass translation degrees of freedom)
+        """
+        if self.system_idx != slice(None):
```
curtischong: this `!= slice(None)` is a bit hard to read. Maybe we need to change our assumptions (a default value?) of what self.system_idx can be? Or maybe by addressing Orion's earlier comment we can avoid this?

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+    # Subtract DOF removed by constraints
+    if constraints is not None:
+        for constraint in constraints:
+            total_dof -= constraint.get_removed_dof(state)
+
+    return max(0, total_dof)  # Ensure non-negative
+
+
+def validate_constraints(
+    constraints: list[Constraint], state: SimState | None = None
```
curtischong: I think we should remove the | None here. Because I think we are always validating the constraints for a state

torch_sim/constraints.py
```diff
@@ -0,0 +1,470 @@
...
+        if len(unique_indices) < len(all_indices):
+            warnings.warn(
+                "Multiple constraints are acting on the same atoms. "
+                "This may lead to unexpected behavior.",
+                UserWarning,
+                stacklevel=3,
+            )
+
+    # Warn about COM constraint with fixed atoms
+    if has_com_constraint and indexed_constraints:
```
curtischong, line 603: This is great work

torch_sim/state.py
```diff
@@ -235,7 +294,16 @@ def get_number_of_degrees_of_freedom(self) -> torch.Tensor:
...
-        return 3 * self.n_atoms_per_system
+        dof_per_system = 3 * self.n_atoms_per_system
+
+        # Subtract DOF removed by constraints
+        if self.constraints is not None:
+            for constraint in self.constraints:
+                removed_dof = constraint.get_removed_dof(self)
+                dof_per_system -= removed_dof
+
+        # Ensure non-negative DOF
```
curtischong, line 345: Maybe we should warn if the degrees of freedom is below 0? Actually, maybe we can put it inside validate_constraints? I can see a case where the dof is 0 but if it's below 0 I think it's more often an error?

torch_sim/state.py
```diff
@@ -649,8 +718,16 @@ def _filter_attrs_by_mask(
     Returns:
         dict: Filtered attributes with appropriate handling for each scope
     """
+    # atoms_mask = torch.isin(state.system_idx, torch.nonzero(system_mask).squeeze())
```
curtischong: nit: unused comment

torch_sim/transforms.py
curtischong: ik this is a helper function, but if we're using it to verify the correctness of other code, it'd be nice to have a test for this

curtischong (commented):
There are definitely things I'll have to rereview after reading Orion's PR. But thank you for this PR! It's currently quite strong and brings us a good step towards implementing constraints

torch_sim/constraints.py
thomasloux: It's the case isn't it? Otherwise, I'm not sure to understand what you mean

torch_sim/constraints.py
thomasloux: I've removed wrap during MD. So state.positions are always unwrapped when applying pbc. it's also like this that it is implemented in ASE. So I don't think it's a problem

torch_sim/models/einstein.py
thomasloux: This file is not supposed to be in this PR sorry

torch_sim/models/einstein.py
thomasloux: This file is not supposed to be in this PR sorry. But this remark is relevant yes

torch_sim/optimizers/state.py
thomasloux: indeed, changed to __post__init__ to enforce force constraints

torch_sim/optimizers/state.py
thomasloux: Actually set_positions at the moment fix the positions according to the current positions so it's not needed. If we need to apply set_positions, it should even be for SimState and not only Optimizer.
Actually tests passed without the set_forces. So maybe it's not relevant, but it seems more rigorous to apply them.

torch_sim/constraints.py
thomasloux: agreed I'm not a fan of this var name

torch_sim/constraints.py
thomasloux: I've removed it

torch_sim/constraints.py
thomasloux: Changed

torch_sim/state.py
thomasloux: actually even dof 0 is a problem. There is no motion then

torch_sim/state.py
```diff
@@ -664,6 +724,18 @@ def _filter_attrs_by_mask(
...
+        constraint.select_constraint(atom_mask, system_mask)
+        for constraint in copy.deepcopy(state.constraints)
+    ]
+    # Remove any None constraints resulting from selection
+    filtered_attrs["_constraints"] = [
+        constraint
+        for constraint in filtered_attrs["_constraints"]
+        if constraint is not None
+    ]
+
```
orionarcher, line 776: Can we do this in a single comprehension with an if statement and walrus operator, like below?

thomasloux:
I already changed that to atom_idx and system_idx!

torch_sim/constraints.py
```diff
@@ -0,0 +1,581 @@
...
+        Args:
+            indices: Indices of atoms to constrain. Can be a tensor or list of integers.
+
+        Raises:
+            ValueError: If both indices and mask are provided, or if indices have
+                       wrong shape/type
+        """
+        # Convert to tensor if needed
+        if not isinstance(indices, torch.Tensor):
+            indices = torch.tensor(indices)
```
orionarcher: AtomConstraint.atom_idx was renamed but that was not propagated to the initialization. Can we rename the arg to `atom_idx` here too?

torch_sim/constraints.py
curtischong: right now this code needs to know about state.system_idx. I was thinking purely from an interface point of view, it would be nice if users that implemented constraints don't need to think about different systems (i.e. not concern themselves with state.system_idx.

However I realize that it's unlikely since we want our tensor operations to be fast and in a batched matter, so we will always need to consider system_idx in implementations.

orionarcher: Yeah I don't think there is really any way around needing to be aware of systems, nor is it a big burden on constraint developers imo

torch_sim/state.py
```diff
@@ -238,6 +247,46 @@ def row_vector_cell(self, value: torch.Tensor) -> None:
         """
         self.cell = value.mT
 
+    def set_positions(self, new_positions: torch.Tensor) -> None:
```
curtischong: since we want people to use this new function, maybe we should add a warning if they don't use set_positions?

```
@property
def positions(self) -> torch.Tensor:
    """Unit cell following the row vector convention."""
    logger.warn("Directly setting state.positions is deprecated. use set_positions instead")
    return self.positions
```

curtischong: Or we can make the set_positions function the @property of `def positions` instead so it will automatically handle constraints and people don't ever need to know about the set_positions function. We may need to prefix positions with an underscore since it the variable might shadow the `@property` definition but I'm not sure

torch_sim/state.py
curtischong: I think dof 0 might be valid. let's say we apply the dof constraint on a bunch of systems. if a system only has 1 atom, then it will have a dof of 0. I think we should still let this happen since other systems might not have a 0 dof.

This might sound nonsensical, but such systems could be generated using generative models - which could make systems of only 1 atom. (I'm saying this because I've experienced issues with an mlip, where it would fail for systems with 0 edges and it was a bit annoying because my run would crash)

torch_sim/constraints.py
```diff
@@ -0,0 +1,581 @@
...
+            raise ValueError(
+                f"Indices must be integers or boolean mask, not dtype={indices.dtype}"
+            )
+
+        # Check for duplicates
+        if len(torch.unique(indices)) < len(indices):
+            raise ValueError(
+                "The indices array contains duplicates. "
+                "Perhaps you want to specify a mask instead, but "
+                "forgot the mask= keyword."
```
curtischong: not sure what this mask= keyword is. but maybe this is leftover from older code. I think we can end this warning at "Perhaps you want to specify a mask instead"

torch_sim/constraints.py
```diff
@@ -0,0 +1,581 @@
...
+        Raises:
+            ValueError: If both indices and mask are provided, or if indices have
+                        wrong shape/type
+        """
+        # Convert to tensor if needed
+        system_idx = torch.as_tensor(system_idx)
+
+        # Ensure we have the right shape and type
+        system_idx = torch.atleast_1d(system_idx)
+        if system_idx.ndim > 1:
```
curtischong: I think we actually want this if statement to be `if system_idx.ndim != 1:` (so we always have a consistent dimension for system_idx?). You'd also need to change the message in ValueError if this is the case

curtischong (approved):
I think this is great! There are a few things I think could be improved but in general (e.g. allowing 0 dof in state.py), but it seems like we have constraints now.

torch_sim/state.py
curtischong: I think this is a small but important thing to do because we don't want people to accidentally use the wrong functions to set positions/forces.

torch_sim/integrators/npt.py
```diff
@@ -435,7 +429,8 @@ def _npt_langevin_velocity_step(
 
     # Update momenta (velocities * masses) with all contributions
     new_velocities = c_1 + c_2 + c_3
-    state.momenta = new_velocities * state.masses.unsqueeze(-1)
+    # Apply constraints. Is it correct to apply constraints here?
+    state.set_momenta(new_velocities * state.masses.unsqueeze(-1))
```
orionarcher: This is a bit of a worrying comment, is it correct?

torch_sim/state.py
orionarcher: Ahh yeah this is important. I agree that having `set_positions` and `positions` is a bit confusing. 

I would suggest we instead call this `constrain_positions` which has no arguments and applies the constraints to the existing positions. I don't think setting positions should necessarily incur the application of constraints, and having the naming mixed up is confusing. A separate operation strikes me as the cleanest solution. We could call it `apply_positions constraints` too.

Whatever we do, let's make sure to propagate it to `forces` and `momenta` as well.

So the flow would look like:
```
state.positions = new_positions
state.constrain_positions()
```

torch_sim/constraints.py
thomasloux: Actually this is handle by torch.atleast_1d

torch_sim/state.py
thomasloux: Actually this `set_positions` mostly makes sense for under-the-hood algorithms. I don't think you can add @property on something that is a already a variable of the class, otherwise you just replace it. So your approach would work with a hidden `_positions` but this start to change a lot the current logic. By the way, you don't always want to apply constraints.
Example: in MD, it's redundant to apply on both forces and momenta. It's probably not wrong, though it may not be this simple for more complicated Constraint like FixBonds, but you do two time the work.

torch_sim/constraints.py
thomasloux: I actually added support for mask. These init functions were copied from ASE a bit too quickly.

torch_sim/integrators/npt.py
thomasloux: It was a reference for apply constraints for NPT in general, but we already added a warning in case you add constraint on a NPT simulation, which I'm sure is wrong at the moment

torch_sim/constraints.py
curtischong: I mentioned != 1 because it'd be more explicit: we want exactly a 1D tensor

thomasloux: agreed on the if statement, but not on error message, although it's a really small detail

torch_sim/state.py
orionarcher: Yeah it makes sense to me that @property isn't a good fit. Still, I don't think it's clear just from the name how `set_positions` differs from direct `positions` assignment.

I think
```
state.positions = new_positions
state.constrain_positions()
```
or
```
state.set_positions_w_constraints(new_positions)
```
is preferable to 
```
state.set_positions(new_positions)
```
which is not clearly different from
```
state.positions = new_positions
```

Somthing less wordy or more precise than `set_positions_w_constraints` is fine, maybe `set_constrained_positions`.

orionarcher: Ok upon reflection my proposed solution would not work. I suggest instead:
1. Removing the `set_positions`, `set_momenta`, and `set_forces` functions from the states
2. Instead creating `constrain_positions`, `constrain_momenta`, and `constrain_forces` functions in `constraints.py`, these would have the following signature and use:

```python
def constrain_positions(positions: torch.Tensor, constraints: list[Constraint], reference_state: ts.SimState) -> torch.Tensor:
        for constraint in (constraints or []):
            constraint.adjust_positions(reference_state, positions)

# then used by
constrain_positions(new_positions, state.constraints, state)
state.positions = new_positions

# of course with a slight adjustment we could also do
constrain_positions(new_positions, state)
state.positions = new_positions
```

I think removing the `constrain_X` logic from the SimStates is preferable because it makes it more reusable across the package and isolates the contraint logic in `constraints.py`.

thomasloux: Potentially I would prefer a more verbose
`state.set_positions(new_positions, apply_constraints=True)`. Only reason is having 1 line when you want to set a new position.
But I don't have a strong opinion on any of the different implementations we discussed

orionarcher: Yeah I agree having a one liner is nice. What about: 
```python
from constraints import set_constrained_positions

set_constrained_positions(state, new_positions)
```
I prefer not having two ways to do the same thing and I like having all the constraints logic in constraints. Could call it something else, I'd be ok with simply `set_positions` too.

orionarcher:
Fantastic job in getting this to the finish line @thomasloux. The only unresolved issue is the question of position/momentum/force assignment with constraints. Happy to be an extra pair of hands on that if desired

torch_sim/state.py
thomasloux: I've changed my mind. One liner like ASE is a good option when you have a private variable _positions and force final user to use `set_positions`. In our case, it's not possible. So I think it's better as you said to keep in 2 lines
```
state.positions = positions
state.constrain_positions()
```
The code is more verbose but more self explainable as well.
I think I prefer this form as one understand that constraints come from state and modify internal positions, rather than  `constrain_positions` which I think is a bit weird either with the argument/signature `constrain_positions(new_positions, state.constraints, state)` or `constrain_positions(new_positions, state)`. In one case, it's a bit weird to provide state (but it's needed for some constraints (eg FixAtoms). The second case, the signature is the same as state.constraint_positions().

orionarcher: Sounds good to me! I could also imagine there could be special states that need their own constraint implementations, so I am happy to leave it on the state. It's a pretty short function, anyhow.

thomasloux:
@orionarcher, I decided to switch to `state.set_constrained_positions(new_positions)` because:
- you need the current state.positions to modify the new_positions
- which property to modify (positions, momenta, forces) is linked to a class (SimState: positions, MDState: momenta, OptimState: forces). So it is simpler to keep on the class level rather than in the constraints.py file.
- It's an explicit one line way to modify a property and apply constraints.

PR #295, opened 2025-10-18
Skip FairChem tests on forks due to HF secret.

CompRhys:
## Summary

Will skip fairchem tests on forks to allow forks to hit merge criteria. This is not a true fix as it skips the tests but it prevents the failures which make end users more likely to ignore other test issues.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,6 +119,7 @@ jobs:
           uv pip install -e ".[test,${{ matrix.model.name }}]" --resolution=${{ matrix.version.resolution }} --system
 
       - name: Run ${{ matrix.model.test_path }} tests
+        if: ${{ !contains(matrix.model.name, 'fairchem') || github.event.pull_request.head.repo.fork == false }}
         env:
           HF_TOKEN: ${{ secrets.HUGGING_FACE_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


CompRhys:
related: https://github.com/facebookresearch/fairchem/issues/1570

.github/workflows/test.yml
```diff
@@ -119,6 +119,7 @@ jobs:
           uv pip install -e ".[test,${{ matrix.model.name }}]" --resolution=${{ matrix.version.resolution }} --system
 
       - name: Run ${{ matrix.model.test_path }} tests
+        if: ${{ !contains(matrix.model.name, 'fairchem') || github.event.pull_request.head.repo.fork == false }}
```
CompRhys, line 122: The tests show as passing with this change rather than not showing due to being skipped. This isn't ideal.

orionarcher: How about we set an env variable with the repo in the action and then skip in `pytest` if the repo isn't TorchSim?

Something like:

```yaml
      - name: Run tests
        env:
          REPO: ${{ github.repository }}
```

```python

@pytest.mark.skipif(
    os.getenv("REPO") != os.getenv("TorchSim/torch-sim"), # or something
    reason="Skipping tests for forked PR"
)
def test_fairchem():
```

CompRhys: functionally that's the same as this? they will just be skipped in the action and the action will be green? All the fairchem tests rely on HF for the models they use

CompRhys: show as passing here is about Github UX not pytest

orionarcher: Ah yeah good point, my mistake.

PR #297, opened 2025-10-19
Fix #293: State to device side effects

### LINKED ISSUE #293: state_to_device does not create a new state contrary to its description
state_to_device in ts.state is supposed to create a new SimState with all tensors moved to the specified device and with the specified data type.

But in order to do it, vars(state) is used which actually is a pointer to the internal variables of state. Change the indexes of the dictionary you modify the input. And of course attrs["positions"] = attrs["positions"].to(dtype=dtype), actually changes the device of the input.

So 
1. It has a side effect of modifying implicitly it's input. 
2. As the variables are not cloned, the "new" state has all its variables linked to the input state.

This function is used in 3 places in torchsim.
1. SimState.to (so side effect is ok)
2. in concatenate_state, for which actually .to should be used, but it does change the behavior in the end
3. initialize_state: for which is not really clear if one wants to modify the input.

What I suggest is either:
- Create a really new state. needs copy vars(state) and clone tensors.
- Just return state, and not create a new state, referencing the input entirely. Then changing the description

sacredvoid:
## Summary

**Fix**: Resolve SimState side effects in device/dtype conversion
**Problem**: SimState.to(), concatenate_states(), and initialize_state() were modifying input states instead of creating new instances.

### Solution:
Fixed _state_to_device() to use copy.copy(vars(state)) instead of direct reference
Updated functions to create new instances without side effects
Key Change:
Impact: Fixes Issue #293, eliminates unexpected state mutations, maintains backward compatibility.
Testing: Added test_fix_demo.py to verify all operations preserve input states unchanged.
## Checklist

### DIFF
--- a/tests/models/conftest.py
+++ b/tests/models/conftest.py
@@ -123,7 +123,7 @@ def make_model_calculator_consistency_test(
     return test_model_calculator_consistency
 
 
-def make_validate_model_outputs_test(
+def make_validate_model_outputs_test(  # noqa: PLR0915
     model_fixture_name: str,
     device: torch.device = DEVICE,
     dtype: torch.dtype = torch.float64,
@@ -135,7 +135,7 @@ def make_validate_model_outputs_test(
         model_fixture_name: Name of the model fixture to validate
     """
 
-    def test_model_output_validation(request: pytest.FixtureRequest) -> None:
+    def test_model_output_validation(request: pytest.FixtureRequest) -> None:  # noqa: PLR0915
         """Test that a model implementation follows the ModelInterface contract."""
         # Get the model fixture dynamically
         model: ModelInterface = request.getfixturevalue(model_fixture_name)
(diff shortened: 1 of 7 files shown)


sacredvoid:
Hi @orionarcher and @thomasloux, saw your message on the slack channel and wanted to make my contribution, let me know if any edits are required!

thomasloux:
by the way, not sure if we need this kind of tests in the repo

torch_sim/state.py
```diff
@@ -573,7 +573,9 @@ def state_to_device[T: SimState](
     if dtype is None:
         dtype = state.dtype
 
-    attrs = vars(state)
+    # Use copy.copy to create a shallow copy of the attributes dict
+    # This avoids modifying the input state's internal dictionary
+    attrs = copy.copy(vars(state))
```
orionarcher: Let's not copy here. The behavior of .to() below is such that it will create a copy if the device is different.

orionarcher:
I'd agree with @thomasloux that we can get rid of most of the tests. I'd just test `.to()` and not the downstream behavior.

sacredvoid:
@orionarcher hey, i've updated the PR with requested changes, let me know if it's good to go, thank you

thomasloux:
Hey, I would remove the tests and modify the description. Now, the IS to change the input state. I would also directly return state to prevent create a new dataclass object with all its variables pointing to another dataclass object

Issue #291, opened 2025-10-16
Convert NPTLangevinState to inherit from MDState and not StateSim


thomasloux:
Context:
NPTLangevinState currently inherits from StateSim. There are not bugs, but it is inconsistent with the other states (example: NVT/NPTNoseHooverState which inherits from MDState).

This will be easy to fix

PR #299, opened 2025-10-20
NPTLangevinState inherits from MDState

hholb:
## Summary

Resolves: #291

This updates the `NPTLangevinState` to inherit from `MDState` instead of `SimState`. Updated the state calculations to use momenta instead of velocities to match other MD integrators.

## Checklist

### DIFF
--- a/torch_sim/integrators/npt.py
+++ b/torch_sim/integrators/npt.py
@@ -427,8 +427,7 @@ def _npt_langevin_velocity_step(
     b = 1 / (1 + (alpha_atoms * dt_atoms) / M_2)
 
     # Velocity contribution with damping
-    velocities = state.momenta / state.masses.unsqueeze(-1)
-    c_1 = a * velocities
+    c_1 = a * state.velocities
 
     # Force contribution (average of initial and final forces)
     c_2 = dt_atoms.unsqueeze(-1) * ((a * forces) + state.forces) / M_2


torch_sim/integrators/npt.py
```diff
@@ -381,9 +374,8 @@ def _npt_langevin_position_step(
     noise_term = noise_prefactor.unsqueeze(-1) * noise
 
     # Velocity and force contributions with random noise
-    c_3 = (
-        state.velocities + dt_atoms.unsqueeze(-1) * state.forces / M_2 + noise_term / M_2
-    )
+    velocities = state.momenta / state.masses.unsqueeze(-1)
```
CompRhys: this should be available as a property of the MDState base class so should use that meaning that we don't need to change anything here.

torch_sim/integrators/npt.py
```diff
@@ -66,11 +65,10 @@ class NPTLangevinState(SimState):
     cell_velocities: torch.Tensor
     cell_masses: torch.Tensor
 
-    _atom_attributes = SimState._atom_attributes | {  # noqa: SLF001
+    _atom_attributes = MDState._atom_attributes | {  # noqa: SLF001
```
CompRhys: forces are already in the _atom_attributes of the MDState so we can remove this line.

torch_sim/integrators/npt.py
```diff
@@ -79,11 +77,6 @@ class NPTLangevinState(SimState):
         "energy",
```
CompRhys: energy is one of the MDState attributes so we can also drop that here.

CompRhys (changes requested):
Thanks for the fixes! A few nit comments that should be fixed before merging.

torch_sim/integrators/npt.py
hholb: Makes life easier! Thanks!

torch_sim/integrators/npt.py
hholb: 👍

torch_sim/integrators/npt.py
hholb: 👍

hholb:
Thanks for the review! Tweaks made!

torch_sim/integrators/npt.py
```diff
@@ -438,13 +427,14 @@ def _npt_langevin_velocity_step(
     b = 1 / (1 + (alpha_atoms * dt_atoms) / M_2)
 
     # Velocity contribution with damping
-    c_1 = a * state.velocities
+    velocities = state.momenta / state.masses.unsqueeze(-1)
+    c_1 = a * velocities
```
CompRhys:
```suggestion
    c_1 = a * state.velocities
```

hholb: Ah good catch, thanks!

PR #305, opened 2025-10-25
Deprecate pbc_wrap_general

curtischong:
## Summary

Originally I wanted to delete it because it's not used in our codebase and we already have wrap_positions which works.

I also wanted to switch pbc_wrap_general's internal implementation to just use wrap_positions but couldn't because pbc_wrap_general wraps cells of arbitrary dimension (not just 3x3). So we're just going to deprecate this for now and delete it at a later date.

I think it's fine dropping support for this function later because an unused helper function that supports periodicity for nxn cell matrices that are bigger than 3x3 probably shouldn't be in torchsim.

## Checklist

### DIFF
--- a/torch_sim/transforms.py
+++ b/torch_sim/transforms.py
@@ -7,10 +7,10 @@ general PBC wrapping.
 
 from collections.abc import Callable, Iterable
 from functools import wraps
-from typing_extensions import deprecated
 
 import torch
 from torch.types import _dtype
+from typing_extensions import deprecated
 
 
 def get_fractional_coordinates(


tests/test_transforms.py
```diff
@@ -318,15 +318,12 @@ def test_pbc_wrap_batched_triclinic() -> None:
...
 
-    # Stack the cells for batched processing
-    cell = torch.stack([cell1, cell2])
-
     # Apply wrapping
     wrapped = ft.pbc_wrap_batched(positions, cell=cell, system_idx=batch)
 
-    # Calculate expected result for first atom (using original algorithm for verification)
-    expected1 = ft.pbc_wrap_general(positions[0:1], cell1)
-    expected2 = ft.pbc_wrap_general(positions[1:2], cell2)
```
curtischong, line 329: changing this test to wrap_positions removes our last dependence on the wrap_positions function in our codebase

PR #306, opened 2025-10-25
convert cell_to_cellpar from ase's numpy implementation to pytorch

### LINKED ISSUE #304: Reimplement cell_to_cellpar in torchsim
cell_to_cellpar is used in Orb: https://github.com/TorchSim/torch-sim/blob/main/torch_sim/models/orb.py#L185

Currently TorchSim transfers the cell to numpy to use the function before transferring it back to torch.

Just need to turn this numpy code into torch: https://ase-lib.org/_modules/ase/geometry/cell.html

wwymak:
## Summary
 resolves #304 

## Checklist

* [ x ] Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.

### DIFF
--- a/docs/user/overview.md
+++ b/docs/user/overview.md
@@ -27,6 +27,6 @@ Learn more in [Understanding Reporting](../tutorials/reporting_tutorial.ipynb)
 
 ## High-level vs Low-Level
 
-Under the hood, TorchSim takes a modular functional approach to atomistic simulation. Each integrator or optimizer function, such as `nvt_langevin,` takes in a model and parameters and returns `init` and `update` functions that act on a unique `State.` The state inherits from `SimState` and tracks the fixed and fluctuating parameters of the simulation, such as the `momenta` for NVT or the timestep for FIRE. The runner functions take this basic structure and wrap it in a convenient interface with autobatching and reporting.
+Under the hood, TorchSim takes a modular functional approach to atomistic simulation. Each integrator or optimizer has associated `init` and `update` functions that initialize and update a unique `State.` The state inherits from `SimState` and tracks the fixed and fluctuating parameters of the simulation, such as the `momenta` for NVT or the timestep for FIRE. The runner functions take this basic structure and wrap it in a convenient interface with autobatching and reporting.
 
 Learn more in [Fundamentals of TorchSim](../tutorials/low_level_tutorial.ipynb) and [Implementing New Methods](../tutorials/hybrid_swap_tutorial.ipynb)
--- a/examples/tutorials/low_level_tutorial.py
+++ b/examples/tutorials/low_level_tutorial.py
@@ -117,10 +117,10 @@ print(f"Stress is a systemwise property with shape: {model_outputs['stress'].sha
 """
 ## Optimizers and Integrators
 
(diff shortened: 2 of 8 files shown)


orionarcher (approved):
LGTM, test is good. Maybe get final sign off from @thomasloux

tests/models/test_orb.py
```diff
@@ -74,3 +78,14 @@ def orbv3_direct_20_omat_calculator() -> ORBCalculator:
 test_validate_direct_model_outputs = make_validate_model_outputs_test(
     model_fixture_name="orbv3_direct_20_omat_model",
 )
+
+
+def test_cell_to_cellpar(si_sim_state: SimState) -> None:
```
thomasloux: You can't use this one as it's cubic cell. So you wouldn't identify the possible issue of convention.

torch_sim/models/orb.py
```diff
@@ -181,9 +213,7 @@ def state_to_atom_graphs(  # noqa: PLR0915
         num_edges.append(len(edges[0]))
 
         # Calculate lattice parameters
-        lattice_per_system = torch.from_numpy(
-            cell_to_cellpar(cell_per_system.squeeze(0).cpu().numpy())
-        )
+        lattice_per_system = cell_to_cellpar(cell_per_system.squeeze(0))
```
thomasloux, line 214: Ok I didn't see that the cell was extracted explicitly with the convention row_vector. When I say that the convention is not the same is that if A=Ase.atoms.cell[:] and B=SimState.cell, then A=B.T (transpose form). So in here, either you assume that the input is row and then you should have the same behaviour as ASE.cell_to_cellpar or you change that and then you have to change the input.
Given the current state of code, I suggest to use the same cell_to_cellpar convention as ASE, which is to assume that the input cell is a row_vector. Then modify the test accordingly

wwymak: Hi, I have changed the function to use the row convention and also updated the test.

torch_sim/models/orb.py
```diff
@@ -59,6 +58,39 @@ def __init__(self, err: ImportError = exc, *_args: Any, **_kwargs: Any) -> None:
...
+        Tensor with [a, b, c, alpha, beta, gamma].
+    """
+    lengths = torch.linalg.norm(cell, dim=1).squeeze()
+    angles = []
+    for i in range(3):
+        j = i - 1
+        k = i - 2
+        ll = lengths[j] * lengths[k]
+        if ll.item() > 1e-16:
+            cell_j = cell[j].squeeze()
```
thomasloux: Any particular reason to use squeeze() ?

thomasloux (commented):
ok

torch_sim/models/orb.py
wwymak: it's mostly to make sure it's 0 dimension vector for the calc. (I think I had an issue where there was an extra dimension before without this, ie [[x, y, z]], but that might have been when I was using the column vector convention)

thomasloux:
Tested locally, small modification to remove unnecessary torch.tensor and use torch.stack to preserve the device of the tensor. Otherwise it would not work on cuda (not catched by test that are only run on cpu)

PR #320, opened 2025-10-28
Mixed Pbc Support

curtischong:
## Summary

Updates torchsim to use the latest version of Vesin which supports mixed PBC neighborlist calculations.

## Checklist

### DIFF
--- a/examples/tutorials/diff_sim.py
+++ b/examples/tutorials/diff_sim.py
@@ -381,7 +381,7 @@ def simulation(
     # Randomly initialize the system.
     # Fix seed for reproducible random positions
     torch.manual_seed(seed)
-    R = torch.rand(N, 2) * box_size
+    R = torch.rand(N, 3) * box_size
 
     # Minimize to the nearest minimum.
     init_fn, apply_fn = gradient_descent(model, lr=0.1)
@@ -424,7 +424,7 @@ diameters = torch.linspace(0.4, 1.0, 10)
 seeds = torch.arange(1, 6)
 box_size_tensor = torch.zeros(len(diameters), len(seeds))
 raft_energy_tensor = torch.zeros(len(diameters), len(seeds))
-bubble_positions_tensor = torch.zeros(len(diameters), len(seeds), N, 2)
+bubble_positions_tensor = torch.zeros(len(diameters), len(seeds), N, 3)
 for i, d in enumerate(diameters):
     for j, s in enumerate(seeds):
         box_size, raft_energy, bubble_positions = simulation(d, s)


curtischong:
got more time tonight will continue then

torch_sim/io.py
```diff
@@ -362,18 +369,11 @@ def phonopy_to_state(
         torch.arange(len(phonopy_atoms_list), device=device), atoms_per_system
     )
 
-    """
-    NOTE: PhonopyAtoms does not have pbc attribute for Supercells assume True
-    Verify consistent pbc
-    if not all(all(at.pbc) == all(phonopy_atoms_lst[0].pbc) for at in phonopy_atoms_lst):
-        raise ValueError("All systems must have the same periodic boundary conditions")
-    """
-
```
curtischong, line 371: I'm unfamiliar with this "supercell" thing. but I deleted it because I think that phononpy always has periodic boundary conditions (https://github.com/phonopy/phonopy/blob/develop/phonopy/structure/atoms.py#L140)? Maybe I just revert this change and we can talk it out in a different PR

torch_sim/models/soft_sphere.py
```diff
@@ -714,7 +715,7 @@ def unbatched_forward(  # noqa: PLR0915
                 cell=cell,
                 pbc=self.pbc,
                 cutoff=self.cutoff,
-                sorti=False,
```
curtischong, line 717: yep. this is a spelling mistake

examples/scripts/7_Others/7.3_Batched_neighbor_list.py
```diff
@@ -18,8 +18,7 @@
 cutoff = torch.tensor(4.0, dtype=pos.dtype)
 self_interaction = False
 
-# Fix: Ensure pbc has the correct shape [n_systems, 3]
-pbc_tensor = torch.tensor([[pbc] * 3] * len(atoms_list), dtype=torch.bool)
+pbc_tensor = torch.tensor(pbc).repeat(state.n_systems, 1)
```
curtischong, line 22: this repeat is needed since torch_nl_linked_cell accepts (num_systems, 3) for the pbc. it doens't just accept a tensor of shape (3,)

tests/test_trajectory.py
```diff
@@ -93,7 +93,7 @@ def test_write_state_single(
     assert trajectory.get_array("positions").shape == (1, 10, 3)
     assert trajectory.get_array("atomic_numbers").shape == (1, 10)
     assert trajectory.get_array("cell").shape == (1, 3, 3)
-    assert trajectory.get_array("pbc").shape == (1,)
+    assert trajectory.get_array("pbc").shape == (1, 3)
```
curtischong: I'm not 100% sure this is the correct test. should it be shape (1,3) or just shape (3,)

torch_sim/integrators/md.py
```diff
@@ -154,10 +172,14 @@ def position_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
     """
     new_positions = state.positions + state.velocities * dt
 
-    if state.pbc:
+    # TODO(curtis): we should only wrap along pbc axes
+    if state.pbc.any():
```
curtischong: lol wait we're going to remove this wrapping. I'll leave wrapping in this PR, but we should delete it in another for discoverability (so ppl can see "removed wrapping" in the change log)

torch_sim/models/metatomic.py
```diff
@@ -217,7 +213,7 @@ def forward(self, state: ts.SimState | StateDict) -> dict[str, torch.Tensor]:  #
                     positions=system_positions,
                     types=system_atomic_numbers,
                     cell=system_cell,
-                    pbc=system_pbc,
```
curtischong, line 220: the simstate.pbc is a torch.Tensor by default now! so we don't need this

torch_sim/neighbors.py
```diff
@@ -11,7 +11,7 @@
 @torch.jit.script
 def primitive_neighbor_list(  # noqa: C901, PLR0915
     quantities: str,
-    pbc: tuple[bool, bool, bool],
```
curtischong, line 14: this is a breaking change. but I'm fine with this change because this will make pbc match all of the other params. if this param was `bool` originally, I'd probably want to make it a `torch.Tensor | bool` instead.

Another reason why I don't want to do `torch.Tensor | bool`  is because I think that torch.jit.script will be slightly confused by union types (and I'll need to manually specify overrides)

torch_sim/transforms.py
```diff
@@ -149,14 +157,20 @@ def pbc_wrap_general(
     frac_coords = positions @ torch.linalg.inv(lattice_vectors).T
 
     # Wrap to reference cell [0,1) using modulo
-    wrapped_frac = frac_coords % 1.0
+    wrapped_frac = frac_coords.clone()
+    wrapped_frac[:, pbc[0]] = frac_coords[:, pbc[0]] % 1.0
+    wrapped_frac[:, pbc[1]] = frac_coords[:, pbc[1]] % 1.0
+    wrapped_frac[:, pbc[2]] = frac_coords[:, pbc[2]] % 1.0
```
curtischong: I tried  `wrapped_frac[:, pbc] = frac_coords[:, pbc] % 1.0` and it doesn't work :'(

curtischong:
there is an annoying bug with torch.jit that I'm trying to solve in a test (our nl function takes too long bc we are timing the compilation and the invocation of the function I believe) but the main API for this is pretty much done. So you can review if you want

torch_sim/transforms.py
thomasloux: Are you sure it doesn't work?
It's a deprecated method, we may not want to adapt it and use it.
If we do want to keep it, pbc needs to be a (d,) shape and so this method doesn't work (I'm actually surprise it would work here. It probably only does because you provide a (n_systems, 3) pbc tensor.

thomasloux: So yes actually the answer is my last observation. But as you don't have a system_idx, you probably need to squeeze to account for (1, d) shape and if n_systems>1 raise an error

torch_sim/transforms.py
```diff
@@ -202,7 +222,10 @@ def pbc_wrap_batched(
     frac_coords = torch.bmm(B_per_atom, positions.unsqueeze(2)).squeeze(2)
 
     # Wrap to reference cell [0,1) using modulo
-    wrapped_frac = frac_coords % 1.0
+    wrapped_frac = frac_coords.clone()
+    wrapped_frac[:, pbc[0]] = frac_coords[:, pbc[0]] % 1.0
```
thomasloux: Same remark, but now account for systems_idx:

wrapped_frac[pbc[system_idx]] should work

torch_sim/transforms.py
```diff
@@ -216,19 +239,22 @@ def minimum_image_displacement(
...
     Returns:
         torch.Tensor: Minimum image displacement vectors with same shape as input.
     """
-    if cell is None or not pbc:
+    if isinstance(pbc, bool):
+        pbc = torch.tensor([pbc] * 3, dtype=torch.bool, device=dr.device)
+    if cell is None or not pbc.any():
         return dr
 
     # Convert to fractional coordinates
```
thomasloux, line 247: You need to adapt the rest of the function to account for mixed periodicity. 
Chatgpt suggests this, which I think is correct:
`    dr_frac -= torch.where(pbc, torch.round(dr_frac), torch.zeros_like(dr_frac))`

torch_sim/neighbors.py
```diff
@@ -778,9 +783,9 @@ def torch_nl_linked_cell(
         positions (torch.Tensor [n_atom, 3]):
             A tensor containing the positions of atoms wrapped inside
             their respective unit cells.
-        cell (torch.Tensor [3*n_structure, 3]): Unit cell vectors according to
+        cell (torch.Tensor [3*num_systems, 3]): Unit cell vectors according to
```
thomasloux: I would rather use n_systems to be consistent with SimState

torch_sim/neighbors.py
thomasloux: Seems reasonable

torch_sim/io.py
thomasloux: Phonopy only makes sense for periodic crystal, although I can imagine that some people may want to that for some kind of 2D crystals with a small z-axis thickness. Sounds reasonable to me to keep that.

tests/test_trajectory.py
thomasloux: I think pbc should be system_attributes so I would argue for (1, 3)

torch_sim/io.py
curtischong: yeah it only makes sense for periodic crystals. But I'll revert this bc I feel like it's better discussed in a diff PR

tests/test_trajectory.py
curtischong: yeah I saw that we're also probably going to change trajectory reporter soon as well to make it less confusing

torch_sim/transforms.py
curtischong: The `d` shape is true. since it's deprecated I'm in favor of not updating this function

torch_sim/transforms.py
curtischong: I definitely missed this.

torch_sim/transforms.py
curtischong: Since pbc is currently a global attribute I think the current change is okay.

curtischong:
@orionarcher or @abhijeetgangan do you recall why pbc is a global attribute? I'm actually fine with either option (it being a global/system attribute) because I think when ppl are doing simulations, they would choose to fix periodicity for all of the systems in their batch.

I do see the case for it being a system attribute. Just wanted your opinion!

tests/test_trajectory.py
curtischong: I don't think performance-wise making pbc a system attribute will do anything. Since I can see the case for torchsim taking in arbitrary states and manipulating them I am also leaning towards making pbc a system attribute

abhijeetgangan:
@curtischong In principle you could have different settings for each system in the batch, even mixed PBC for that matter. Only cases it would break would be where the simulations is meant for PBC systems only like phonons or elastic constants etc. We could certainly add error handling case by case but I think this could be future addition and we can keep and issue open for it.

IMO this PR is already a big feature compared to the latter as I see it having more applications than people running different PBC systems in a batch.

thomasloux:
> IMO this PR is already a big feature compared to the latter as I see it having more applications than people running different PBC systems in a batch.

I agree, we should do that separately. Thanks for this PR @curtischong

torch_sim/transforms.py
thomasloux: ok but I don't think this code is correct. if pbc is tensor of shape (3,), all lines do the same. then normally wrapped_frac[:, pbc] should work. if pbc is shape (1, 3) then wrapped_frac[:, pbc[0]], but then I don't understand how the test can pass

curtischong: I think I was testing on stale code. I've updated it to use `wrapped_frac[:, pbc] = frac_coords[:, pbc] % 1.0`

tests/test_trajectory.py
curtischong: okay. yeah this is completely wrong. In this PR I'll keep pbc a global attribute and make it (3,). that should be the last one. Man I was quite tired when I first wrote this PR.

curtischong:
@thomasloux I have to rewrite how trajectory stores pbc tensors later tonight (it doesn't understand global attributes that are not constants). but other than that, we should be good.

torch_sim/trajectory.py
curtischong, line 1: The tests pass, but I had a hard time following this logic. This file is overdue for an overhaul.

torch_sim/state.py
```diff
@@ -91,6 +93,11 @@ def system_idx(self) -> torch.Tensor:
             """A getter for system_idx that tells type checkers it's always defined."""
             return self.system_idx
 
+        @property
+        def pbc(self) -> torch.Tensor:
+            """A getter for pbc that tells type checkers it's always defined."""
+            return self.pbc
```
orionarcher, line 99: I have mixed feelings about adding nonfunctional code just to satisfy type checkers. Why aren't they picking up on the attribute?

torch_sim/state.py
```diff
@@ -125,6 +129,13 @@ def __post_init__(self) -> None:
                 f"masses {shapes[1]}, atomic_numbers {shapes[2]}"
             )
 
+        if isinstance(self.pbc, bool):
+            self.pbc = [self.pbc] * 3
+        if not isinstance(self.pbc, torch.Tensor):
+            self.pbc = torch.tensor(
+                self.pbc, dtype=torch.bool, device=self.positions.device
```
orionarcher: nit but should be able to use `self.device` instead of `self.positions.device`

torch_sim/trajectory.py
orionarcher: I'd like to see trajectory aligned with the [H5MD](https://www.nongnu.org/h5md/) standard at some point, thats the next overhaul I have planned

torch_sim/trajectory.py
```diff
@@ -828,11 +834,11 @@ def return_prop(self: Self, prop: str, frame: int) -> np.ndarray:
...
                 start, stop = 0, 1
-            return self.get_array(prop, start=start, stop=stop)[0]
+            return self.get_array(prop, start=start, stop=stop)
 
-        arrays["cell"] = np.expand_dims(return_prop(self, "cell", frame), axis=0)
-        arrays["atomic_numbers"] = return_prop(self, "atomic_numbers", frame)
-        arrays["masses"] = return_prop(self, "masses", frame)
+        arrays["cell"] = np.expand_dims(return_prop(self, "cell", frame), axis=0)[0]
+        arrays["atomic_numbers"] = return_prop(self, "atomic_numbers", frame)[0]
+        arrays["masses"] = return_prop(self, "masses", frame)[0]
```
orionarcher: why move the index here?

orionarcher (commented):
Broadly looking good to me, a few comments

orionarcher:
> IMO this PR is already a big feature compared to the latter as I see it having more applications than people running different PBC systems in a batch.

Given that we may soon have batched neighbor list calculations it's unclear to me if we want to support mixed-PBC. It's nice in principle but I would bet it's very rarely needed in practice and may complicate neighbor list calculations.

tests/test_trajectory.py
orionarcher: Agree let's keep as (3,) for now

torch_sim/state.py
curtischong: This is because we define the attribute as torch.Tensor | bool so when the dataclass is initialized, users know they can pass those types in. This attribute is converted to a torch.Tensor inside the __post_init__. This typehint here is to let users know that when they access this attribute it's always a torch.Tensor.

I tried many different things to make this type accurate. One idea I had was to make pbc just a torch.Tensor, and then manually define a custom __init__ for the dataclass. Unfortunately, child classes will have to **manually** call this init - which is just really messy and bloats up all of the code for children childclasses. I believe that this is the best way to use the nice "auto init" functions that dataclasses provides while having accurate types

torch_sim/trajectory.py
curtischong: I updated the return_prop function (line 831) to not auto return the [0] index. This was done because `return_prop(self, "pbc", frame)` is now a global attribute that is a tensor (so we don't want to just return the first index, but the entire tensor).

So for all of the attributes, the functionality is the exact same - it's actually just a change to make PBC to work. (so we can return a tensor for global attribuets - not just constants - which was what it was before)

curtischong:
I think catalytic applications would benefit from mixed PBC (without having to manually create a vacuum).

Mixed PBC does complicate the writing of kernels. Here is an example of what Vesin had to do when implementing mixed PBC https://github.com/Luthaf/vesin/pull/76/files#diff-38d490d3ab0a91479a981cd188ef933c68d7cbb24bc90ab3d66595f7da4e0da6R133
I do think that this PR does seem larger than it actually is because I think mic displacement vectors were changed to match ASE's. 

Here was my implementation of mixed PBC in vesin (albeit with the wrong mic displacement vectors) and it's quite small
https://github.com/Luthaf/vesin/pull/75/files#diff-38d490d3ab0a91479a981cd188ef933c68d7cbb24bc90ab3d66595f7da4e0da6R74

We should ask the nvidia ppl to see how hard it might be to add mixed PBC (can they use the same algorithms?). It might be a really small change like the diff in my PR

torch_sim/state.py
thomasloux: I actually still think, we could do the custom *init* only for SimState, and assume, as it is the case at the moment, that any Subclass of SimState (MDState...) will be created from a init_fn and another SimState, so that pbc will indeed be a torch.Tensor.

torch_sim/trajectory.py
```diff
@@ -643,10 +643,15 @@ def get_array(
         if name not in self.array_registry:
             raise ValueError(f"Array {name} not found in registry")
 
-        return self._file.root.data.__getitem__(name).read(
+        data = self._file.root.data.__getitem__(name).read(
             start=start, stop=stop, step=step
         )
 
+        if name == "pbc":
```
thomasloux: Why did you choose to squeeze here ?

torch_sim/trajectory.py
thomasloux: are you sure about that? I think it's just because get_array(name, n_frame, n_frame+1) normally return (1, size). You modify it to squeeze for pbc, so indeed you need to modify here.  But I think you shouldn't squeeze pbc

torch_sim/state.py
curtischong: Nope. Child dataclasses do not call the inits of parent dataclasses (because they don't know which params to pass in - annoying I know). See https://github.com/TorchSim/torch-sim/pull/335.

This issue is reported here https://bugs.python.org/issue43835

I should add a comment in the code pointing to that issue

thomasloux: I agree, but my point is that it's not a problem. You can assume it *will* be a torch.Tensor because it will be initiated from a SimState

thomasloux: If I take a NVT Langevin, the init function is 

```
def nvt_langevin_init(
    state: SimState | StateDict,
    model: ModelInterface,
    *,
    kT: float | torch.Tensor,
    seed: int | None = None,
    **_kwargs: Any,
) -> MDState:
    """Docs..."""
    if not isinstance(state, SimState):
        state = SimState(**state)

    model_output = model(state)

    momenta = getattr(
        state,
        "momenta",
        calculate_momenta(state.positions, state.masses, state.system_idx, kT, seed),
    )

    return MDState(
        positions=state.positions,
        momenta=momenta,
        energy=model_output["energy"],
        forces=model_output["forces"],
        masses=state.masses,
        cell=state.cell,
        pbc=state.pbc, # So pbc is taken from a SimState, so pbc *is* torch.tensor
        system_idx=state.system_idx,
        atomic_numbers=state.atomic_numbers,
        constraints=state.constraints,
    )
```

thomasloux: My code design assumes that no user should create directly a MDState, which can be called into question. But it has the advantage of being a simpler code

torch_sim/trajectory.py
curtischong: I need to stop vibecoding this fix. My apologies

torch_sim/state.py
curtischong: ahhh I was confusing nvt_langevin_init with the __init__ of MDState. Can you put an issue up? so we can discuss it further (bc it'll prob be in a diff PR). Personally I'm fine with either pattern (using the overrides like I was doing in this PR vs the nvt_langevin_init function)

torch_sim/trajectory.py
```diff
@@ -518,6 +518,20 @@ def write_arrays(
 
         self.flush()
 
+    def write_global_array(self, name: str, array: np.ndarray | torch.Tensor) -> None:
```
curtischong, line 515: I needed to write this function just for pbc. because otherwise the trajectory format will add an extra dimension (bc `write_arrays` thinks there is a system_idx dimension)

torch_sim/trajectory.py
```diff
@@ -830,15 +839,17 @@ def _get_state_arrays(self, frame: int) -> dict[str, np.ndarray]:
         arrays["positions"] = self.get_array("positions", start=frame, stop=frame + 1)[0]
 
         def return_prop(self: Self, prop: str, frame: int) -> np.ndarray:
+            if prop == "pbc":
```
curtischong, line 836: yes I hate this extra if statement as well. this trajectory file format also doesn't egneralize well to global attributes from child simstates. we need to settle on a good simstate definition for attributes soon

thomasloux (commented):
opened a PR just before you on that. Is the change on radian really needed? If so i'll change my PR and merge it

curtischong:
I was mainly trying to fix the tests. was still playing around. if you got it to work without the radian it's fine. the angle is the actual issue. I'll probably revert the orb changes and rebase off main to have your fix

examples/tutorials/diff_sim.py
```diff
@@ -369,20 +373,25 @@ def simulation(
 ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     # Create the simulation environment.
     box_size = box_size_at_packing_fraction(diameter, packing_fraction)
-    cell = torch.eye(2) * box_size
+    cell = torch.eye(3) * box_size
```
thomasloux, line 376: I'm not against diff_sim being in 3D, but was there any reason to be in 2D in the first place?

tests/test_trajectory.py
```diff
@@ -509,7 +510,7 @@ def test_write_ase_trajectory(
         np.testing.assert_allclose(
             atoms.get_atomic_numbers(), random_state.atomic_numbers.numpy()
         )
-        assert atoms.pbc.all() == random_state.pbc
+        np.testing.assert_array_equal(atoms.pbc, random_state.pbc.numpy())
```
thomasloux, line 513: maybe you should transfer first in cpu() in case we ever run this on gpu?

curtischong: I think this is fine because the random_state is automatically made on cpu by default. Maybe we can do it in a separate PR if people run these tests and see it fail? (trying to keep this one slim)

examples/tutorials/diff_sim.py
curtischong: ¯\\_(ツ)_/¯

PR #326, opened 2025-10-30
Add CSVR / V-Rescale thermostat and anisotropic C rescale barostat 

thomasloux:
## Summary

Add Canonical sampling through velocity rescaling (CSVR) or v-rescale as called on Gromacs.

Note: removed the optimization to sample from Gamma distribution which I doubt is useful on a high level framework like pytorch.

## Checklist

### DIFF
--- a/torch_sim/integrators/npt.py
+++ b/torch_sim/integrators/npt.py
@@ -367,9 +367,9 @@ def _npt_langevin_position_step(
     state.positions = c_1 + c_2.unsqueeze(-1) * c_3
 
     # Apply periodic boundary conditions if needed
-    if state.pbc:
+    if state.pbc.any():
         state.positions = ts.transforms.pbc_wrap_batched(
-            state.positions, state.cell, state.system_idx
+            state.positions, state.cell, state.system_idx, state.pbc
         )
 
     return state
@@ -1030,9 +1030,9 @@ def _npt_nose_hoover_exp_iL1(  # noqa: N802
     new_positions = state.positions + new_positions
 
     # Apply periodic boundary conditions if needed
-    if state.pbc:
+    if state.pbc.any():
         return ts.transforms.pbc_wrap_batched(
-            new_positions, state.current_cell, state.system_idx
+            new_positions, state.current_cell, state.system_idx, pbc=state.pbc
         )
     return new_positions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,67 @@
 <!-- markdownlint-disable -->
 # Changelog
 
+## v0.4.1
+
+Thank you to everyone who contributed to this release! This release includes important bug fixes, new features, and API improvements. @thomasloux, @curtischong, @CompRhys, @orionarcher, @WillEngler, @samanvya10, @hn-yu, @wendymak8, @chuin-wei, @pragnya17, and many others made valuable contributions. 🚀
+
+### 💥 Breaking Changes
+* Standardize argument order to (state, model) by @pragnya17 in [#341](https://github.com/TorchSim/torch-sim/pull/341)
+* Deprecate pbc_wrap_general by @curtischong in [#305](https://github.com/TorchSim/torch-sim/pull/305)
+
+### 🎉 New Features
+* Mixed PBC Support by @curtischong in [#320](https://github.com/TorchSim/torch-sim/pull/320)
+* Feature: Batched NVT Nose-Hoover by @thomasloux in [#265](https://github.com/TorchSim/torch-sim/pull/265)
+* Add degrees of freedom (dof) in state and temperature computation by @thomasloux in [#328](https://github.com/TorchSim/torch-sim/pull/328)
+* Add sources npt langevin by @thomasloux in [#298](https://github.com/TorchSim/torch-sim/pull/298)
+
+### 🐛 Bug Fixes
+* MACE: transfer atomic numbers to CPU before converting to numpy by @t-reents in [#289](https://github.com/TorchSim/torch-sim/pull/289)
+* Fixed max atoms memory estimation by @nh-univie in [#279](https://github.com/TorchSim/torch-sim/pull/279)
+* Fixing model loading logic (names and cache dir) for fairchem models by @nh-univie in [#278](https://github.com/TorchSim/torch-sim/pull/278)
+* Fix fairchem and neighbors tests by @WillEngler in [#317](https://github.com/TorchSim/torch-sim/pull/317)
+* Fix #293: State to device side effects by @samanvya10 in [#297](https://github.com/TorchSim/torch-sim/pull/297)
+* Fix graph-pes key issue by @jla-gardner in [#303](https://github.com/TorchSim/torch-sim/pull/303)
+* Fix calculate static state after relax by @curtischong in [#338](https://github.com/TorchSim/torch-sim/pull/338)
+* Misc fixes by @orionarcher in [#336](https://github.com/TorchSim/torch-sim/pull/336)
+* Fix cell to cellpar by @thomasloux in [#342](https://github.com/TorchSim/torch-sim/pull/342)
+* Fix failing Docs build by @CompRhys in [#296](https://github.com/TorchSim/torch-sim/pull/296)
+* Fix: correct speedup plot image path in README by @Joaqland in [#333](https://github.com/TorchSim/torch-sim/pull/333)
+
+### 🛠 Enhancements
+* Put SimState Init logic into __post_init__ (and enforce kw_args=true for SimState children) by @curtischong in [#335](https://github.com/TorchSim/torch-sim/pull/335)
+* Replace vars(state) with state.attributes by @orionarcher in [#329](https://github.com/TorchSim/torch-sim/pull/329)
+* Rename and reorder some variables by @orionarcher in [#316](https://github.com/TorchSim/torch-sim/pull/316)
+* Add SevenNet path, str types for model arg by @YutackPark in [#322](https://github.com/TorchSim/torch-sim/pull/322)
+* Add version attribute by @thomasloux in [#311](https://github.com/TorchSim/torch-sim/pull/311)
+* Convert cell_to_cellpar from ase's numpy implementation to pytorch by @wendymak8 in [#306](https://github.com/TorchSim/torch-sim/pull/306)
+* Autobatch OOM handling by @chuin-wei in [#337](https://github.com/TorchSim/torch-sim/pull/337)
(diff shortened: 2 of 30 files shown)


thomasloux:
image

And is working well. Temperature set is indeed 200K, script modified from https://github.com/TorchSim/torch-sim/blob/main/examples/scripts/3_Dynamics/3.11_Lennard_Jones_NPT_Langevin.py

thomasloux:
Note: change the name to follow GROMACS, I prefer this name

thomasloux:
I need to check very carefully for the number of degrees of freedom used. I think the V-rescale implementation in SimpleMD removes the 3 dof linked to the COM.

thomasloux:
Observation on some simple implementations:
- V-rescale seems to work great
- C-Rescale seems interestingly more difficult to use. Note that I haven't tried to use an estimation of the isothermal compressibility. In my tests it was rather unstable depending on the used parameters.

If you want to experiment with C-rescale:
```
"""Lennard-Jones simulation in NPT ensemble using Langevin thermostat."""

# /// script
# dependencies = ["scipy>=1.15"]
# ///
import itertools
import os

import torch

import torch_sim as ts
from torch_sim.units import MetalUnits as Units
from torch_sim.units import UnitConversion


# Set up the device and data type
dtype = torch.float32

a_len = 5.26  # Lattice constant

# Generate base FCC unit cell positions (scaled by lattice constant)
base_positions = torch.tensor(
    [
        [0.0, 0.0, 0.0],  # Corner
        [0.0, 0.5, 0.5],  # Face centers
        [0.5, 0.0, 0.5],
        [0.5, 0.5, 0.0],
    ],
    device=device,
    dtype=dtype,
)

# Create 4x4x4 supercell of FCC Argon manually
positions = []
for i, j, k in itertools.product(range(4), range(4), range(4)):
    for base_pos in base_positions:
        # Add unit cell position + offset for supercell
        pos = base_pos + torch.tensor([i, j, k], device=device, dtype=dtype)
        positions.append(pos)

# Stack the positions into a tensor
positions = torch.stack(positions)

# Scale by lattice constant
positions = positions * a_len

# Create the cell tensor
cell = torch.tensor(
    [[4 * a_len, 0, 0], [0, 4 * a_len, 0], [0, 0, 4 * a_len]], device=device, dtype=dtype
)

# Create the atomic numbers tensor (Argon = 18)
atomic_numbers = torch.full((positions.shape[0],), 18, device=device, dtype=torch.int)
# Create the masses tensor (Argon = 39.948 amu)
masses = torch.full((positions.shape[0],), 39.948, device=device, dtype=dtype)

# Initialize the Lennard-Jones model
# Parameters:
#  - sigma: distance at which potential is zero (3.405 Å for Ar)
#  - epsilon: depth of potential well (0.0104 eV for Ar)
#  - cutoff: distance beyond which interactions are ignored (typically 2.5*sigma)
model = LennardJonesModel(
    use_neighbor_list=False,
    sigma=3.405,
    epsilon=0.0104,
    cutoff=2.5 * 3.405,
    device=device,
    dtype=dtype,
    compute_forces=True,
    compute_stress=True,
)
model.compile()
state = ts.SimState(
    positions=positions,
    masses=masses,
    cell=cell.unsqueeze(0),
    atomic_numbers=atomic_numbers,
    pbc=True,
)
# state = ts.concatenate_states([state.clone() for _ in range(2)])  # Duplicate for 2 systems

n_steps = 2_500
dt = torch.tensor(0.001, dtype=dtype) * ts.units.MetalUnits.time
kT = torch.tensor(100.0, dtype=dtype) * ts.units.MetalUnits.temperature
external_pressure = torch.tensor(10_0.0, dtype=dtype) * ts.units.MetalUnits.pressure
# tau_p = torch.tensor(0.1, dtype=dtype)
tau_p = 100 * dt
tau = 100 * dt
isothermal_compressibility = torch.tensor(5e-6, dtype=dtype) / ts.units.MetalUnits.pressure

# Initialize integrator using new direct API
state = ts.npt_crescale_init(
    state=state,
    model=model,
    dt=dt,
    kT=kT,
    tau_p=tau_p,
    isothermal_compressibility=isothermal_compressibility,
    seed=42,
)

# Run dynamics for several steps
energies = []
temperatures = []
pressures = []
volumes = []
for _step in range(n_steps):
    state = ts.npt_crescale_step(
        state=state,
        model=model,
        dt=dt,
        kT=kT,
        external_pressure=external_pressure,
        tau=tau
    )
    if _step % 100 == 0:
        pressure = ts.get_pressure(
            model(state)["stress"],
            ts.calc_kinetic_energy(
                masses=state.masses, momenta=state.momenta, system_idx=state.system_idx
            ),
            torch.linalg.det(state.cell),
        )

        # Calculate instantaneous temperature from kinetic energy
        temp = ts.calc_kT(
            masses=state.masses, momenta=state.momenta, system_idx=state.system_idx
        )
        energies.append(state.energy)
        temperatures.append(temp / ts.units.MetalUnits.temperature)
        pressures.append(pressure / ts.units.MetalUnits.pressure)
        volumes.append(torch.linalg.det(state.cell))

# Convert temperatures list to tensor
temperatures_tensor = torch.stack(temperatures)
temperatures_list = [t.tolist() for t in temperatures_tensor.T]

energies_tensor = torch.stack(energies)
energies_list = [t.tolist() for t in energies_tensor.T]

pressures_tensor = torch.stack(pressures)
pressures_list = [t.tolist() for t in pressures_tensor.T]
```

torch_sim/integrators/npt.py
```diff
@@ -1602,3 +1602,330 @@ def npt_nose_hoover_invariant(
...
+
+    # c projections
+    box[:, 2, 0] = torch.sum(a * c, dim=1) / box[:, 0, 0]
+    box[:, 2, 1] = (torch.sum(b * c, dim=1) - box[:, 2, 0] * box[:, 1, 0]) / box[:, 1, 1]
+    box[:, 2, 2] = torch.sqrt(
+        torch.sum(c * c, dim=1) - box[:, 2, 0] ** 2 - box[:, 2, 1] ** 2
+    )
+
+    # Upper-triangular entries are 0 by initialization
+    return box
```
abhijeetgangan: Shoudn't the box be upper triangular as per the other parts of code base? unless you are taking care of it elsewhere.

thomasloux:
Current implementation with an exterior isotropic pressure only.
Could be easily extended, the equations are in paper and are rather simple.

The isotropic barostat should probably be implemented as well

torch_sim/integrators/npt.py
```diff
@@ -1602,3 +1602,330 @@ def npt_nose_hoover_invariant(
...
+    tau_p: torch.Tensor  # shape: [n_systems]
+
+    _system_attributes = MDState._system_attributes | {  # noqa: SLF001
+        "stress",
+        "isothermal_compressibility",
+        "tau_p",
+    }
+
+
+def _compute_instantaneous_internal_pressure(
```
abhijeetgangan: This looks like a general definition of stress with the velocity term. Would be nice to move it to `quantities.py` since it might be useful in other places?

torch_sim/integrators/nvt.py
```diff
@@ -458,3 +458,185 @@ def nvt_nose_hoover_invariant(
...
+### Number of DOF?######################################################################################
+def _vrescale_update(
+    state: MDState,
+    tau: float | torch.Tensor,
+    kT: float | torch.Tensor,
+    dt: float | torch.Tensor,
+) -> MDState:
+    """Update the momentum by a scaling factor as described by Eq.A7 Bussi et al.
+
+    Note that we don't implement the optimize code from Bussi, which won't be useful
```
abhijeetgangan, line 507: This is in reference to the noise term?

torch_sim/integrators/npt.py
```diff
@@ -1602,3 +1602,330 @@ def npt_nose_hoover_invariant(
     e_tot += torch.square(cell_momentum) / (2 * state.cell_mass)
 
     return e_tot
+
+
+###################
+# Implement full anisotropic NPT with cell rescaling barostat
```
abhijeetgangan: Better suited in the doc string?

abhijeetgangan:
Looks great. Left a few comments. Choosing the correct value if isothermal compressibility was also an issue in previous version of NPT in ase so we could leave that to the user for now.

torch_sim/integrators/npt.py
thomasloux: Forgot to remove that

torch_sim/integrators/nvt.py
thomasloux: yes

torch_sim/integrators/npt.py
thomasloux: done

torch_sim/integrators/npt.py
thomasloux: good catch thanks

thomasloux:
Hey, I've been playing a lot with this integrator. So my feeling is that the anisotropic NPT is quite unstable, so one should choose a relatively high tau_p for it to work. I've merely run the simulations for some 100s picoseconds, which is quite short. But usually for other integrators, the convergence is quite fast.
The thing is that I'm not able to say wether it does work if tau_p is large enough, or if unstabilities still arise but later. I let you and future users provide more insight on this.
I'm quite confident that the equations are correctly implemented, but it's not impossible that I missed something. I'll provide reach out to some of the authors to better understand what's happening.

thomasloux:
@orionarcher @curtischong Hey, could you have a look on these integrators. I'd like to merge them. Out of cautiousness, we could add a message saying it's still in beta or something else if you want.
More documents should be added to give feedbacks on these integrators, but this will only be possible with more feedback from users/contributors.
Add also still add a good testing of the correct sampling of integrators, for instance using physical_validation. Started doing it but didn't have time to fully do it: https://github.com/shirtsgroup/physical_validation
I'll add an issue with part of my code for physical_validation

abhijeetgangan:
The default `tau_p` is a bit large but I think that's fine. I wonder if `1000 / 2000 * dt` could also work. Wasn't `rotate_gram_schmidt` supposed to be upper triangular? 

Otherwise LGTM.

thomasloux:
I've changed it back because I needs to be like that to update positions and momenta. I transpose it when updating the cell though, so the cell stays upper triangular

abhijeetgangan:
Great! I think to avoid this we might want to change it to lower triangular as other codes use that convention. It's becoming challenging to keep track as the code base grows.

This PR can be merged.

orionarcher (approved):
Didn't closely analyze the physics but LGTM

PR #328, opened 2025-10-31
Add degrees of freedom (dof) in state and temperature computation

thomasloux:
## Summary

Current implementation always assumes 3*num_atoms degrees of freedom, but it's not true for Nosé Hoover if center of mass momentum is 0, which is done in calculate_momenta.

Now the temperature should be compute directly state.calc_temperature().

Notice however that the temperature unit is assumed to be Metal.

This PR will be necessary for Constraints and V-rescale integrator and has to be merged before.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,4 +11,4 @@ Before a pull request can be merged, the following items must be checked:
 * [ ] Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.
 * [ ] Tests have been added for any new functionality or bug fixes.
 
-We highly recommended installing the `prek` hooks running in CI locally to speedup the development process. Simply run `pip install prek && prek install` to install the hooks which will check your code before each commit.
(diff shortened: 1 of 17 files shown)


thomasloux:
I haven't modify the neighbor list, I don't think the fail is because of the PR

torch_sim/quantities.py
curtischong, line 69: Not related, but maybe we should put `calc_temperature` into MDState in the future so we don't need to rely on people correctly passing in the params to this function

curtischong (approved):
I'd consider renaming it to `get_number_of_degrees_of_freedom` to be consistent with ase https://ase-lib.org/ase/atoms.html#ase.Atoms.get_number_of_degrees_of_freedom. It would also be more searchable (for people unfamiliar with the code) and be more interpretable for people because dof might be an unknown acronym to people

torch_sim/quantities.py
thomasloux: already done ! I don't mind having it additionally here as if it was a utils.py file. But agreed that if people miss use it we may change it

thomasloux:
I'm not a fan of this verbose method name, but I prefer to align with ASE convention as much as possible

Issue #334, opened 2025-11-02
SimState child classes are not calling SimState Init


curtischong:
This is a severe issue and is a bug I introduced when I was working on the SimState Init this summer.

I thought child dataclass classes of parent dataclasses will always call the parent init on the child init. They don't.

I tried several different approaches (moving from an __init__ paradigm in SimState towards a __post_init__ paradigm) but it had many typing drawbacks and would fail some typecheckers. The current __init__ paradigm is the most typesafe. So I think we will keep it.

I have a few ideas but the simplest and most straightforward is to just have all the child states of simstate call the parent init. I will also probably add some type verification inside the `__init_subclass__` to verify that it is callign the simstate init.

The reason why htis is a big problem is because if we are not calling the simstate init, some important variables are checked/coerced into the right types. (e.g. system_index could be a tensor or none, but after the init it's always a tensor)

I have a wif fix. But will finish it up next morning:

https://github.com/curtischong/torch-sim/pull/10/files#r2484087837

thomasloux:
Didn't know either about init of child classes.
Why don't you want to put all the checks in 'post__init'?
Additionally at the moment, no one should directly create any subclass. This is handled in integrator / optimizer init starting from a SimState. Although we should take care of this edge case, because the check would only be run at the start of any pipeline

curtischong:
We can init some fields (system_idx) as `torch.Tensor | None`. If we define the system_idx field as `torch.Tensor | None` in the dataclass (so we can pass in a tensor or none during the init time), during inference time, the type of system_idx will appear as `torch.Tensor | None` to type checkers everytime we use it

thomasloux:
It would still advocate for using 'post_init' to check for shape and actually force user to provide a Tensor 'system_idx' for subclass of SimState. It would be easier to write new subclass, knowing that init function will be used in all cases from a SimState

curtischong:
yeah I tried a few things and I think moving back to postinit is the best way. It was very frustrating. But #335 is pretty much the fix I feel

PR #335, opened 2025-11-02
Put SimState Init logic into __post_init__ (and enforce kw_args=true for SimState children)

curtischong:
## Summary

Child dataclasses do not automatically call the parents init. So instead, I changed SimState to remove the init and just use post init (so child simstates will automatically call the post init, rather than having to explicitly cal the __init__). The key insight in making this work was to have a typechecking block force typecheckers to think that the value is not `| None` when we get `system_idx`

Note: this PR is breaking because we needed to add `kw_only` to all of the child dataclasses. This is because I needed to add `= field(default=None)` to the system_idx field in `SimState`

This fixes https://github.com/TorchSim/torch-sim/issues/334

## Checklist

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -88,7 +88,7 @@ class SimState:
 
         @property
         def system_idx(self) -> torch.Tensor:
-            """A getter for simstate that tells typecheckers that it's always defined."""
+            """A getter for system_idx that tells type checkers it's always defined."""
             return self.system_idx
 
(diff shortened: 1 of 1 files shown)


thomasloux:
Need to have a more careful review but looks good to me. Actually I've already suggest using kw_args=False when adding constraints, as this variable is also a default value []

orionarcher (approved):
LGTM

thomasloux (approved):

PR #338, opened 2025-11-06
Fix calculate static state after relax

curtischong:
## Summary

Before this change, we'd get this error

```
TypeError: torch_sim.runners.static.<locals>.StaticState() got multiple values for keyword argument 'energy'
```
This is because we were creating a static_state but passing in fields that already existed (because we optimized the state before calculating the static property). Since we used **vars() to initialize the static state, we'd pass properties like energies twice into the constructor for StaticState.

To fix this PR I also had to make StaticState's attributes be dependent on only SimState's attributes.

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -579,8 +579,8 @@ def static(
         forces: torch.Tensor
         stress: torch.Tensor
 
-        _atom_attributes = state._atom_attributes | {"forces"}  # noqa: SLF001
-        _system_attributes = state._system_attributes | {  # noqa: SLF001
+        _atom_attributes = SimState._atom_attributes | {"forces"}  # noqa: SLF001
+        _system_attributes = SimState._system_attributes | {  # noqa: SLF001
             "energy",
             "stress",
         }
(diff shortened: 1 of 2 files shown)


torch_sim/runners.py
```diff
@@ -579,8 +577,8 @@ class StaticState(SimState):
         forces: torch.Tensor
         stress: torch.Tensor
 
-        _atom_attributes = state._atom_attributes | {"forces"}  # noqa: SLF001
-        _system_attributes = state._system_attributes | {  # noqa: SLF001
+        _atom_attributes = SimState._atom_attributes | {"forces"}  # noqa: SLF001
```
curtischong, line 582: I don't like changing the atom attributes from state to SimState. But I'm okay with it for now since torchsim currently only supports calculating energies, forces, and stresses.

When we support magnetic moments (and arbitrary properties in general) we will need to rethink how to create this

curtischong: The reason why I needed to set these atom attributes to be derived from SimState is because if I left it as `state`, then inside the trajectory reporter, it would think that StaticState would have extra attributes (that we did not initialize when we created it) the extra attribute in particular is `velocity`.

So by explicitly defining the _atom_attributes to be from SimState, we only say that this StaticState has a fixed limited set of attributes

orionarcher: I plan to add mag moms soon so I'd prefer we pick a solution here that will be generalizable.

curtischong: Yeah I need to think through that some more as well. On another note, I'd prefer to do that in a different PR so this one can be constrained to fixing just this bug

Issue #331, opened 2025-10-31
Add support for AMD GPUs (consumer/datacenter)


amacbride:
I'm working on a patch to support AMD consumer and datacenter GPUs (gfx1102/gfx942). I'm building a regression test suite and will be testing on a few CDNA3/4 and RDNA3/4 cards.

If anyone has specific suggestions about what to include, please comment here and I will incorporate them.

orionarcher:
Thanks for working on this! Per our discussion on slack it sounds like the main change is to the default neighbor list?

> I'm building a regression test suite and will be testing on a few CDNA3/4 and RDNA3/4 cards.

Is this something that would be integrated into TorchSim CI? Curious what you have in mind.

amacbride:
[torchsim-rocm-support.patch](https://github.com/user-attachments/files/23293027/torchsim-rocm-support.patch)

amacbride:
> Thanks for working on this! Per our discussion on slack it sounds like the main change is to the default neighbor list?
> 
> > I'm building a regression test suite and will be testing on a few CDNA3/4 and RDNA3/4 cards.
> 
> Is this something that would be integrated into TorchSim CI? Curious what you have in mind.

Yes, that's probably the right place for it. I want to check both correctness and performance.

amacbride:
I'm going to continue testing on more-exotic cards, but the test suite passed on the AMD Radeon RX 7600. There was a warning about one MACE test, but that was already a float64 model being run as float32, so a 6% difference isn't entirely surprising.

  - Test: test_mace_consistency[ti_sim_state] - the titanium (Ti) test case fails (14/15 pass)                                                                                                                                                                                                                             
  - Error: Force comparison off by small margin                                                                                                                                                                                                                                                                                  
    - Absolute difference: 1.065e-05 vs tolerance 1.0e-05 (6.5% over tolerance)                                                                                                                                                                                                                                                 
    - Relative difference: 1.2 (120%) vs tolerance 1e-05                                                                                                                                                                                                                                                                        
    - 2 out of 6 force components mismatched for this one case

abhijeetgangan:
This should be moved to discussion along with the AMD neighborlist kernel request. I have had other people request this as well.

The problem will always be acceleration of the models as they are most of the time the bottleneck. A intermediate solution is to have triton kernels.

PR #347, opened 2025-11-20
Add support for AMD GPUs (consumer/datacenter) #331

amacbride:
## Summary

* Enable AMD GPUs by making vesin optional and using standard_nl as a fallback
* patched neighbors.py

## Checklist

* [ x ] Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.

### DIFF
--- a/tests/models/test_mace.py
+++ b/tests/models/test_mace.py
@@ -25,7 +25,7 @@ except (ImportError, ValueError):
 
 raw_mace_mp = mace_mp(model=MaceUrls.mace_mp_small, return_raw_model=True)
 raw_mace_off = mace_off(model=MaceUrls.mace_off_small, return_raw_model=True)
-DTYPE = torch.float32
+DTYPE = torch.float64
 
 
 @pytest.fixture
@@ -47,27 +47,6 @@ def ts_mace_model() -> MaceModel:
     )
 
 
-@pytest.fixture
-def ase_mace_calculator_float64() -> MACECalculator:
-    return mace_mp(
-        model=MaceUrls.mace_mp_small,
-        device="cpu",
-        default_dtype="float64",
-        dispersion=False,
-    )
-
-
-@pytest.fixture
-def ts_mace_model_float64() -> MaceModel:
-    return MaceModel(
-        model=raw_mace_mp,
-        device=DEVICE,
-        dtype=torch.float64,
-        compute_forces=True,
-        compute_stress=True,
-    )
-
-
 test_mace_consistency = make_model_calculator_consistency_test(
     test_name="mace",
     model_fixture_name="ts_mace_model",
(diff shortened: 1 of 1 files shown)


torch_sim/neighbors.py
```diff
@@ -498,19 +508,19 @@ def standard_nl(
     return mapping, shifts
 
 
-@torch.jit.script
 def vesin_nl_ts(
```
orionarcher: I'd advocate that we retain `vesin_nl_ts` unchanged but add a new function `vesin_w_fallback` that has the switching logic between `vesin_nl_ts` and `standard_nl`, then make that the default.

tests/test_vesin_fallback.py
```diff
@@ -0,0 +1,78 @@
+"""Test vesin fallback functionality for ROCm/AMD GPU support."""
```
orionarcher: I'd advocate we put these tests in test_neighbors.py instead of in their own file.

tests/test_vesin_fallback.py
```diff
@@ -0,0 +1,78 @@
...
+
+import pytest
+import torch
+
+from torch_sim import neighbors
+
+
+def test_vesin_availability_flag() -> None:
+    """Test that VESIN_AVAILABLE flag is correctly set."""
```
orionarcher: As I understand it, none of these tests will actually test the switching if vesin is available, which it will be in CI.

Could we write a test where we monkeypatch the imports for vesin to None so that the `VESIN_AVAILABLE` flag will be set to `False`?

torch_sim/neighbors.py
amacbride: Sounds good, will make that change.

amacbride: (I decided to call the function `torchsim_nl` so that it could remain unchanged if it is decided to change implementations in the future.)

amacbride:
This should be close to finished. I incorporated @orionarcher 's feedback and slightly modified the titanium test to run in float64 instead of float32 -- boundary issues were causing the test to fail originally; it passes now.

tests/models/test_mace.py
```diff
@@ -47,15 +47,47 @@ def ts_mace_model() -> MaceModel:
...
+
+@pytest.fixture
+def ts_mace_model_float64() -> MaceModel:
+    return MaceModel(
+        model=raw_mace_mp,
+        device=DEVICE,
+        dtype=torch.float64,
+        compute_forces=True,
+        compute_stress=True,
+    )
```
orionarcher: Should we make float64 the default here? Might be better than having two versions of both calculators floating around.

torch_sim/neighbors.py
```diff
@@ -508,95 +508,165 @@ def standard_nl(
...
+        shifts = S.to(dtype=dtype, device=device)
+
+        if sort_id:
+            idx = torch.argsort(mapping[0])
+            mapping = mapping[:, idx]
+            shifts = shifts[idx, :]
+
+        return mapping, shifts
+
+    def vesin_nl(
```
orionarcher, line 586: It looks like vesin_nl is defined twice? Once with jit compilation and once without?

orionarcher (approved):
Thanks for the quick turnaround!

Just a couple small comments. I trust you to finish this off.

tests/models/test_mace.py
amacbride: Fine with me -- I didn't want to change the default unilaterally.

PR #354, opened 2025-11-23
[DRAFT] Arbitrary Attribute Support

curtischong:
## Summary

This is the first draft for supporting arbitrary attributes inside of SimState. The tests don't pass because this is half-complete. We may make some changes which will affect this code.

The main idea is to introduce 3 new dictionaries inside SimState to track all of the extra attributes each state may want to store.

When we define a new model, the output of this model needs to specify the kind of each attribute it outputs (if it's an atom-wise, system-wise, or global attribute) - this allows us to add models that work on any set of properties (not just energy, forces, and stress)

Note: I want the constraints PR https://github.com/TorchSim/torch-sim/pull/294 to go in before this one because these 2 PRs change fundamental parts of torchsim and I'd rather rebase this one ontop of the constraints PR.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -142,8 +142,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HUGGING_FACE_TOKEN }}
         run: |
           if [[ "${{ matrix.model.name }}" == *"fairchem"* ]]; then
-            uv pip install "huggingface_hub[cli]" --system
-            hf auth login --token "$HF_TOKEN"
+            uv pip install huggingface_hub --system
           fi
           pytest -vv -ra -rs --cov=torch_sim --cov-report=xml ${{ matrix.model.test_path }}
 
@@ -193,7 +192,6 @@ jobs:
           HF_TOKEN: ${{ secrets.HUGGING_FACE_TOKEN }}
         run: |
           if [[ "${{ matrix.example }}" == *"fairchem"* ]]; then
-            uv pip install "huggingface_hub[cli]" --system
-            hf auth login --token "$HF_TOKEN"
+            uv pip install huggingface_hub --system
           fi
           uv run --with . ${{ matrix.example }}
--- a/tests/models/test_fairchem_legacy.py
+++ b/tests/models/test_fairchem_legacy.py
@@ -2,7 +2,9 @@ import os
 import traceback
 
 import pytest
+import torch
 
+import torch_sim as ts
 from tests.conftest import DEVICE
 from tests.models.conftest import (
     consistency_test_simstate_fixtures,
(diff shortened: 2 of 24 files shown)


torch_sim/models/interface.py
```diff
@@ -151,27 +168,32 @@ def forward(self, state: SimState | StateDict, **kwargs) -> dict[str, torch.Tens
...
-            stress = output.get("stress", None)
+            energy = output["system_attributes"]["energy"]
+            forces = output["atom_attributes"]["forces"]
+            stress = output["system_attributes"].get("stress")
             ```
         """
 
 
+# TODO: we should put this logic inside __init_subclass__ of Modelinterface to
+# automatically validate the model outputs when the model is subclassed.
```
curtischong, line 189: I think we should do this in another PR because this Pr is already kinda long, and automatically enforcing validate_model_outputs (by putting this logic inside __init_subclass__) will require a refactor of all models (which will make this PR longer). Since we are backwards compatible with older models that just returns energy/forces/stress it's okay to not do this enforcement now.

torch_sim/state.py
```diff
@@ -907,21 +1021,53 @@ def concatenate_states[T: SimState](  # noqa: C901
...
+            concatenated_per_system_extra[prop] = torch.cat(tensors, dim=0)
+        else:  # Non-tensor attributes, take first one (they should all be identical)
+            concatenated_per_system_extra[prop] = tensors[0]
+
     # Concatenate system indices
     concatenated["system_idx"] = torch.cat(new_system_indices)
 
     # Create a new instance of the same class
-    return state_class(**concatenated)
+    new_state = state_class(**concatenated)
```
curtischong, line 1051: this line of code is the main reason why I put up this feature as a draft. If we just pass in these extra attributes (e.g. energy) into the constructor of the state_class, we will error out since the init function doesn't expect it. So to compensate, I am just setting these extra attributes *after* we've initialized the state_class.

This works - but is obviously a bit longer (since we need to distinctly specify get_attrs_for_scope for "only_extra" and "only_default". It'd be nice if we didn't need to do this and the constructor just understood which attributes should go into extra or not. maybe the extra dictionary isn't even the right way to do this.

Thoughts are very welcome here!

torch_sim/state.py
```diff
@@ -629,17 +703,27 @@ def get_attrs_for_scope(
...
+            if attribute_kind in ["only_extra", "all"]:
+                attr_names |= state._extra_global_attributes.keys()  # noqa: SLF001
         case _:
             raise ValueError(f"Unknown scope: {scope!r}")
     for attr_name in attr_names:
-        yield attr_name, getattr(state, attr_name)
+        yield attr_name, state.get(attr_name)
 
 
 def _filter_attrs_by_mask(
```
curtischong, line 729: This function is what's causing the tests to fail. I have yet to migrate get_attrs_for_scope to set the attributes differently for default vs extra attributes.

orionarcher:
Can you say a little more about the used case here? Are you sure we need arbitrary properties?

Reporters support arbitrary properties, so we can get them out as long as they are a function of the model and state. If integrators/optimizers need them, it will be easier to build on them if they are defined in code.

If we do need things like charge and spin, it might be sufficient to just add charge and spin to the SimState as optional attributes. Until the diversity of properties supported by models becomes overwhelming, that should work fine. Perhaps I'll regret saying this, but there are only _so many_ atomic properties that models can support. If we feel it becomes overwhelming, we can figure out a more abstractable solution down the road.

With that in mind, I'd suggest either opening a new PR that adds charge/spin to SimState or just adding it to the model interface for MACE and Fairchem.

curtischong:
@orionarcher Yeah I wanted to offer a solution that was more general but I'm fine with just using https://github.com/TorchSim/torch-sim/pull/373 for now. I kinda don't like the solution in this PR because making a general solution feels a bit clunky.

Issue #356, opened 2025-11-24
Cannot append to exisiting trajectory via `ts.integrate`/`ts.optimize`


danielzuegner:
When trying to append to existing trajectories, I construct my reporter like so:

```python
reporter = TrajectoryReporter(
    filenames=relax_filenames,
    state_frequency=1,
    metadata=self.metadata,
    trajectory_kwargs=dict(mode="a"),
)
```

Here, the trajectories inside `relax_filenames` already contain partial trajectories. Now, if I continue from these trajectories into `ts.integrate`, I get an error in `_validate_arrays()`:

```
  File "/opt/miniconda/envs/env/lib/python3.12/site-packages/torch_sim/runners.py", line 499, in optimize
    trajectory_reporter.report(state, step, model=model)
  File "/opt/miniconda/envs/env/lib/python3.12/site-packages/torch_sim/trajectory.py", line 264, in report
    self.trajectories[idx].write_state(substate, step, **self.state_kwargs)
  File "/opt/miniconda/envs/env/lib/python3.12/site-packages/torch_sim/trajectory.py", line 782, in write_state
    self.write_arrays(data, steps)
  File "/opt/miniconda/envs/env/lib/python3.12/site-packages/torch_sim/trajectory.py", line 510, in write_arrays
    self._validate_array(name, array, steps)
  File "/opt/miniconda/envs/env/lib/python3.12/site-packages/torch_sim/trajectory.py", line 587, in _validate_array
    raise ValueError(
ValueError: steps[0]=1 must be greater than the last recorded step 10 for array positions
```

The issue appears to be that in [`ts.optimize`](https://github.com/TorchSim/torch-sim/blob/main/torch_sim/runners.py#L460)/[`ts.integrate`](https://github.com/TorchSim/torch-sim/blob/main/torch_sim/runners.py#L202) we always start from `step=1`, even if we're in append mode. In that case, the validation will fail and lead to a crash. Thus, I don't see a way currently to actually append to a trajectory via `ts.optimize`/`ts.integrate`. 

From my perspective, a solution could be to auto-detect the current `step` from the trajectory if we're in `a` mode and start the integration from there, not 1. If you agree I can draft a small PR to address that.

Thanks!

orionarcher:
Yep, that makes sense to me! A PR would be awesome. Would you make sure to write a test that demonstrates the before/after difference? Happy to take a look when done.

danielzuegner:
[...]

Issue #43, opened 2025-03-14
Enable reproducible restarts from stopped simulations


orionarcher:
Currently there is no clean way to restart a simulation with the exact state from a previous simulation. This should be corrected.

CompRhys:
Resolution strategy will be to address #19 and deprecate hdf5 for h5md assuming nothing adverse arises in doing so. Then we should be able to write a utility that will allow resumption assuming the PRNG state has been saved in line with #460.

PR #361, opened 2025-11-25
Feature: Enable appending to trajectory when using `ts.optimize`/`ts.integrate`

danielzuegner:
## Summary

* Adds functionality to read the latest step per trajectory and trajectory reporter to set the initial step accordingly.
* Adds tests `test_optimize_append_to_trajectory` and `test_integrate_append_to_trajectory` to ensure trajectories are being appended to.
* Also fixes a bug where, whenever a structure has converged, all other trajectories are reset by re-opening them in `"w"` mode: https://github.com/TorchSim/torch-sim/blob/main/torch_sim/runners.py#L483.

Closes https://github.com/TorchSim/torch-sim/issues/356.
## Checklist

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -28,6 +28,9 @@ from torch_sim.typing import StateLike
 from torch_sim.units import UnitSystem
 
 
+logger = logging.getLogger(__name__)
+
+
 def _configure_reporter(
     trajectory_reporter: TrajectoryReporter | dict,
     *,
@@ -126,9 +129,11 @@ def _determine_initial_step_for_integrate(
                 "    reporter.truncate_to_step(min(reporter.last_step))\n\n"
                 "before calling integrate again."
             )
-        logging.info(
-            f"Detected existing trajectory with last step {last_logged_step}."
-            f" Resuming integration from step {initial_step}."
+        logger.info(
+            "Detected existing trajectory with last step %s. Resuming integration "
+            "from step %s.",
+            last_logged_step,
+            initial_step,
         )
     return initial_step
 


torch_sim/runners.py
```diff
@@ -485,9 +501,13 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
             and og_filenames is not None
             and (step == 1 or len(converged_states) > 0)
         ):
+            mode_before = trajectory_reporter.trajectory_kwargs["mode"]
+            # temporarily set to "append" mode to avoid overwriting existing files
+            trajectory_reporter.trajectory_kwargs["mode"] = "a"
```
orionarcher: Is the append mode fixing the bug or just necessary for appending?

More generally, this isn't grappling with the fact that some the existing trajectories might already exist or might have different numbers of steps

torch_sim/trajectory.py
```diff
@@ -302,6 +301,43 @@ def close(self) -> None:
...
+            return 0
+
+        max_step = 0
+        for trajectory in self.trajectories:
+            if trajectory._file.isopen:
+                last_step = trajectory.last_step
+            else:
+                with TorchSimTrajectory(trajectory._file.filename, mode="r") as traj:
+                    last_step = traj.last_step
+            max_step = max(max_step, last_step)
```
orionarcher: I think this needs to return `list[int]` and not `int`. TorchSim is fully batched and the trajectory reporter shouldn't make strong assumptions about homogenous content. This gets to your point about "gaps" in the trajectory.

danielzuegner:
> > Also fixes a bug where, whenever a structure has converged, all other trajectories are reset by re-opening them in "w" mode: https://github.com/TorchSim/torch-sim/blob/main/torch_sim/runners.py#L483.
> 
> Can you say more here? This would be a pretty serious bug because it would mean ll the optimization trajectories are wrong.
> 
> More generally, I think this PR needs to grapple with the fact that if we are appending, the trajectories we are appending to might have different past progress. In a batched setting, we can't just take the maximum progress of one as the progress of all. I realize that makes literally everything more annoying but such is the curse and blessing of TorchSim.

Unfortunately I believe this affects all optimization trajectories. Looking at [`load_new_trajectories`](https://github.com/TorchSim/torch-sim/blob/main/torch_sim/trajectory.py#L177), we can see that the trajectories are being instantiated from scratch with `self.trajectory_kwargs`, i.e., write mode by default. I've noticed that when I was debugging my PR that the longer trajectories are chopped at the start when another one finishes. I'd have steps `[10, 15, 20, ...]` in the trajectory instead of `[5, 10, ...]`. Happy to be proven wrong, but I think this is a bug that potentially affects all optimization trajectories currently.

tests/test_trajectory.py
```diff
@@ -834,3 +835,99 @@ def test_write_ase_trajectory_importerror(
...
+                # Check that the trajectory file has 5 frames
+                np.testing.assert_allclose(traj.get_steps("positions"), range(1, 6))
+
+        trajectory_reporter_2 = ts.TrajectoryReporter(
+            traj_files, state_frequency=1, trajectory_kwargs=dict(mode="a")
+        )
+        _ = ts.integrate(
+            system=int_state,
+            model=lj_model,
+            timestep=0.1,
```
thomasloux: This timestep is really big as the unit is ps, consider using 0.01 or 0.001

thomasloux:
I would rather push for a PR allowing for a restart of simulation #43. In practice the difference is that supposed if you run multiple time `ts.integrate(n_step=1000)`, the first call will run for longer but for the rest it will finish directly. You could add an argument to explicitly run for longer without reading yourself the trajectory.
Note: for a real restart or run for longer, this doesn't change the fact currently the TrajectoryReporter does only save SimState, and not any more variables of subclass. For instance, if you run a NVT Nosé Hoover MD, this means you loose the state of the thermostat. So even run for longer is not technically right. This would be true as well for a (near future) LBFGS optimization.

danielzuegner:
> I would rather push for a PR allowing for a restart of simulation #43. In practice the difference is that supposed if you run multiple time `ts.integrate(n_step=1000)`, the first call will run for longer but for the rest it will finish directly. You could add an argument to explicitly run for longer without reading yourself the trajectory. Note: for a real restart or run for longer, this doesn't change the fact currently the TrajectoryReporter does only save SimState, and not any more variables of subclass. For instance, if you run a NVT Nosé Hoover MD, this means you loose the state of the thermostat. So even run for longer is not technically right. This would be true as well for a (near future) LBFGS optimization.

@thomasloux happy to change it to this way of running. Just to make sure I understand. The proposal is to:
* Detect the last time step per trajectory (say, `[5, 10]` for a toy example of two systems).
* If `n_steps=15`, we run 10 more steps for the first one and 5 more steps for the second trajectory.

If that's the case, I think we have to do it like in the `optimize()` case, where `convergence_tensor` would be the tensor of bools whether each trajectory has reached its terminal step yet. Is that understanding correct?

torch_sim/trajectory.py
danielzuegner: Please take a look at the current implementation, it turned out to be not so difficult to implement. Now there's no "gap" anymore in the trajectories.

thomasloux:
For a MD simulations at least, there is no such thing, all systems evolve at the same time. I actually assumes it the same for optimize, if not for convergence criteria that would be met.
So in case of `ts.integrate`,  you would get `n_step` from trajectory, and you would run a for loop like `range(n_step_traj, n_step_total)`instead of the current PR suggestion `range(n_initial, n_initial + n_step)` and then having to manually remove n_initial when you want to access kT values for instance.

thomasloux:
I don't have any strict opinion on optimization as I don't use it this much

danielzuegner:
> For a MD simulations at least, there is no such thing, all systems evolve at the same time. I actually assumes it the same for optimize, if not for convergence criteria that would be met. So in case of `ts.integrate`, you would get `n_step` from trajectory, and you would run a for loop like `range(n_step_traj, n_step_total)`instead of the current PR suggestion `range(n_initial, n_initial + n_step)` and then having to manually remove n_initial when you want to access kT values for instance.

Thanks @thomasloux! I've updated the code to assume that for `integrate()` all trajectories start from the same time step, and then only integrate for `n_steps - initial_step` steps. Is that in line with what you had in mind?

danielzuegner:
Will take a look at the failures and hopefully fix them

danielzuegner:
Tests pass locally for me now, let's see what happens in the CI.

torch_sim/runners.py
```diff
@@ -197,14 +221,18 @@ def integrate[T: SimState](  # noqa: C901
...
 
         # run the simulation
-        for step in range(1, n_steps + 1):
+        for step in range(initial_step, n_steps + 1):
             state = step_func(
-                state=state, model=model, dt=dt, kT=kTs[step - 1], **integrator_kwargs
+                state=state,
+                model=model,
+                dt=dt,
+                kT=kTs[step - initial_step],
```
thomasloux: use `kT=kTs[step]`to be consistent. Use case: you restart a simulation where you progressively increase the temperature

torch_sim/runners.py
```diff
@@ -499,14 +538,21 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
...
-                trajectory_reporter.report(state, step, model=model)
-            step += 1
-            if step > max_steps:
-                # TODO: max steps should be tracked for each structure in the batch
-                warnings.warn(f"Optimize has reached max steps: {step}", stacklevel=2)
+                trajectory_reporter.report(
+                    state, (initial_step + steps_so_far).tolist(), model=model
+                )
+            steps_so_far += 1
+            exceeded_max_steps = (initial_step + steps_so_far) > max_steps
```
thomasloux: I did not run this code, but I highly suspect this is wrong. you add steps_so_far, that is a global counter of the while loop, while initial_step is a slice of the og_initial_step. This means, if state is large enough to be divided in multiple batches, you still account for the step of batch 1...n-1 for batch n. You should to `step = og_initial_step[autobatcher.current_idx]`and then incremement `step += 1`. This should increment the og_initial_step so that you are consistent on the step of each one.

thomasloux (changes requested):
One part to clarify for `ts.optimize`, otherwise it's a really nice work. I think in a later PR we should clarify the Trajectory objects. I'm not a fan of the mode 'r' vs 'a'. I would probably prefer an argument: either 'restart=True' or 'overwrite_if_exists=True'.

torch_sim/runners.py
danielzuegner: But since step starts at 1 by default, this leads to an indexing error for the last step, where we'd try to index `kTs[5]` while it is only length 5. Hence I think we need to subtract `initial_step` (by default 1) to avoid the off-by-one error?

danielzuegner: I will restore it to `step - 1` as it was before.

torch_sim/runners.py
danielzuegner: Ok, I've changed the behavior in the code. Unfortunately indexing `step = og_initial_step[autobatcher.current_idx]` creates a copy (as we're indexing with a tensor or list), so `step += 1` is not "persistent". In the current version I'm doing the increment explicitly via `step[autobatcher.current_idx] += 1` instead. Let me know what you think!

torch_sim/runners.py
thomasloux: yes sorry you're right it's rather `step-1`

torch_sim/runners.py
thomasloux: Looks good to me!

thomasloux (approved):
good for me now

torch_sim/trajectory.py
```diff
@@ -141,19 +140,39 @@ def __init__(
...
+    def filenames(self) -> list[str] | None:
+        """Get the list of trajectory filenames.
+
+        Returns:
+            list[str] | None: List of trajectory file paths,
+                or None if no trajectories are loaded.
+        """
+        if not self.trajectories:
+            return None
+        return [traj.filename for traj in self.trajectories]
```
orionarcher, line 171: This is a good change

torch_sim/trajectory.py
```diff
@@ -167,19 +186,23 @@ def load_new_trajectories(
...
+        self.trajectories = [
+            TorchSimTrajectory(
+                filename=filename,
+                metadata=self.metadata,
+                **self.trajectory_kwargs,
             )
+            for filename in filenames
+        ]
+        # Restore original mode
+        self.trajectory_kwargs["mode"] = _mode
```
orionarcher, line 206: Why shouldn't we overwrite them if mode is "w"? In my mind that's the intended behavior if we are in write mode, to overwrite all of the existing trajectories with new ones. This could be affected by the user by setting `trajectory_kwargs={"mode"="a"}.

If the feeling is that defaulting to "w" is too strong, the place to change it would be on line 137, where we set 
`self.trajectory_kwargs["mode"] = self.trajectory_kwargs.get("mode", "w")`

`_mode = self.trajectory_kwargs.get("mode", "w")`, also, not clear why we'd have "w" be the default original mode here.

Without this change, I still think `load_new_trajectories` makes sense than reopen.

torch_sim/trajectory.py
```diff
@@ -255,18 +280,19 @@ def report(
         all_props: list[dict[str, torch.Tensor]] = []
         # Process each system separately
         for idx, substate in enumerate(split_states):
+            _step = step[idx] if isinstance(step, list) else step
```
orionarcher: Could we call this `sys_step` or something like that? Naming with a leading underscore is the appropriate convention, that would pythonically imply that it isn't going to be used.

torch_sim/trajectory.py
```diff
@@ -302,6 +328,39 @@ def close(self) -> None:
...
+        if not self.trajectories:
+            return []
+        last_steps = []
+        for trajectory in self.trajectories:
+            if trajectory._file.isopen:
+                last_steps.append(trajectory.last_step)
+            else:
+                with TorchSimTrajectory(trajectory._file.filename, mode="r") as traj:
+                    last_steps.append(traj.last_step)
+        return last_steps
```
orionarcher, line 388: Why not use trajectory.last_step here?

torch_sim/runners.py
```diff
@@ -178,6 +181,27 @@ def integrate[T: SimState](  # noqa: C901
...
+            return initial_state  # type: ignore[return-value]
+        if len(set(last_logged_steps)) != 1:
+            warnings.warn(
+                "Trajectory files have different last steps. "
+                "Using the minimum last step for resuming integration."
+                "This means that some trajectories may be truncated.",
+                stacklevel=2,
+            )
+            for traj in trajectory_reporter.trajectories:
+                traj.truncate_to_step(last_logged_step)
```
orionarcher: All of this logic (and as much of the logic below) should be moved either to a utility function, akin to `_configure_reporter`, the `integrate` function itself should stay as tidy as possible because it is already doing a lot of work.

orionarcher (changes requested):
>  I've updated the code to assume that for integrate() all trajectories start from the same time step, and then only integrate for n_steps - initial_step steps. Is that in line with what you had in mind?

Sorry for being a bit late on this, but I am going to push back on this implementation.

I think that n_steps should mean the number of steps you take, not the number of steps you reach. My heavy prior would be that n_steps is a number of steps taken, as a user, I'd find it quite strange if that wasn't the case. What do you think?

I also don't think that if the steps don't match that we should truncate all the trajectories to the same length. In my mind, we should detect the current step across all the trajectories, then progressively append to that step + i, where i is the number of steps taken by the integrator. Does that make sense?

I want to say I am quite appreciative of the PR and I think being able to append with the runners is a good feature. I am being nit picky because I think it's challenging to get the right behavior and expectations for this in a batched context.

torch_sim/trajectory.py
```diff
@@ -675,9 +740,29 @@ def get_steps(
...
+            return 0
+
+        max_step = 0
+        for name in self.array_registry:
+            steps_node = self.get_steps(name)
+            if len(steps_node) > 0:
+                last_step = int(steps_node[-1])
+                max_step = max(max_step, last_step)
+
+        return max_step
```
orionarcher: I think `last_step` should actually be the last step with positions, not the last step across all arrays, because it is the positions that we will be reconstituting the state from, presumably. 

When we are resuming, I think we should then truncate all saved data beyond the last positional save, because those will be recalculated by the newly appended trajectory.

To discuss.

thomasloux:
Hey Orion, I think that both implementations make sense. I see two possibilities:
- The user has a mode="a" in mind, so it makes sense to have n_steps to be the n_steps to add and not the total number of steps in trajectory.
- The user wants to restart a trajectory (think of a failure or a slurm job that finished too early). Then I think it makes sense to run the same code and the that TorchSim can detect the traj file to start from it.

I think the 2nd option is more appealing then the first one. Also we could easily obtained the first behaviour by adding an argument to the functions.

The truncation is also a direct result of the fact I pushed for a restart behaviour. At the moment a MD simulation from ts.integrate should always be in sync for all states in batch, so it does not make sense to allow different indices in this PR. We could change that in another PR though

orionarcher:
Good framing. Both use cases are valid so perhaps it's worth considering the left out group. 

### Finishing the simulation if `n_steps` = additional steps

```python
total_steps = 1000
steps_completed_before_interrupt = my_trajectory_reporter.last_step  # using the last_step property introduced in this PR
n_steps = 1000 - steps_completed_before_interupt
integrate(...n_steps=n_steps)
```

### Appending 20 steps if `n_steps` = total steps

```python
new_steps = 1000
current_number_of_steps = my_trajectory_reporter.last_step
n_steps = current_number_of_steps - new_steps
integrate(...n_steps=n_steps)
```

Either is still accessible no matter what default is adopted and both are about the same amount of code. Ultimately, it comes down to which is more expected and which is more common. I am not sure which is more common, it very well might be restarting, but I feel that appending steps is more expected of an `n_steps` argument. 

> At the moment a MD simulation from ts.integrate should always be in sync for all states in batch, so it does not make sense to allow different indices in this PR. We could change that in another PR though

Good point that separate indices are a distinct feature and could go elsewhere. This is a potentially dangerous default though. Imagine I have 100 simulations and I can fit 10 on a GPU at a time. My simulation proceeds nicely through the first 90 and then only makes it 1 step into the the final 10. If I rerun the same script expecting to restart, I'll erase the progress on the first 90 simulations because the smallest `last_step` will be 1. Instead of truncating, perhaps we could enforce that all trajectories have the same `last_step`? That feels safer to me.

> The user wants to restart a trajectory (think of a failure or a slurm job that finished too early). Then I think it makes sense to run the same code and the that TorchSim can detect the traj file to start from it.

This is going to be difficult when some simulations will run to completion and others will not begin. I think some modifications will need to be made to the script to weed out already started PRs. 


OK, as I outline these challenges and think about it I am realizing that it's pretty hard to restart simulations in TorchSim right now and my proposal would keep it hard. Perhaps we can add some flag like `restart=False` in the kwargs that will nicely handle the complexities outlined above and allow a script to be rerun without worrying about any of this. That would require a bit of logic added to what we have now. I have to run right now but will think on this further.

torch_sim/trajectory.py
danielzuegner: My argument is that the mode (`"w"` vs `"a"`) is meant by the user only for the initial opening of the file, not the (potentially many) times we re-open the trajectory file during the optimization (e.g., when some of the trajectories finish). I don't think it is ever intentional to wipe all existing progress when that happens, even if the user specified `"w"` mode in the beginning.

Hence, in my PR I propose to use the user-specified mode only for the first time we open the trajectory files (in the constructor of the trajectory reporter), and afterwards avoid wiping then files by ensuring `"a"` mode for the times we re-open the files during the optimization.

What is your alternative proposal to ensure that (i) we use the user-intended file mode initially, but (ii) don't erase the files during the optimization?

torch_sim/trajectory.py
danielzuegner: Because the file could be closed and then we cannot read the last step from it. We can of course also open the file inside the trajectory but it felt a bit odd that the trajectory would "open itself" inside `last_step()`.

torch_sim/trajectory.py
orionarcher: Setting aside where it's intentional, I do think it should be possible to wipe the existing files when re-running, though I agree that by default appending is better. LAMMPS has an append mode when writing dump files that is by [default set to `False`](https://docs.lammps.org/dump.html), so this particular sharp edge isn't unusual.

I would propose that we set the default to "a" but wipe the files if "w" is set.

orionarcher:
In the case of an interupted simulation it's likely that there will be some files that are complete, some that are incomplete, and some that have not started. We need a way to resume simulations  that balances ease, sharp edges, and encouraging best practices.

Here is what I propose:

1. Require that all files in the TrajectoryReporter all have the same number of steps, either 0 (do not exist yet) or N. If a simulation is interrupted, it becomes the responsibility of the user to identify which systems have completed and discard those from the file list. A line in the script that discards completed files can always be included and will just do nothing if it's the first run.

2. Provide a utility for resetting many systems to the same step. This can be called externally when needed, but I don't think it should be applied blindly internally, lest we wipe lots of progress. For the partially complete files these can be wiped to the same step and n_steps can be adjusted to account for their length.

3. `n_steps` should be the number of new steps not total. I looked at OpenMM, LAMMPS, and ASE and even when appending `n_steps` is the number of new steps, not the total number.

Does this workflow feel too complicated? Let me know what y'all think.

torch_sim/trajectory.py
danielzuegner: I agree it should be possible to wipe existing files when re-running, and I think it's possible to do that with the existing code. In the [constructor](https://github.com/TorchSim/torch-sim/blob/6cba46f672617ac9aa1a98fb334c80ec5ad1f164/torch_sim/trajectory.py#L136) of `Trajectory`, we take the `mode` from the user-provided `kwargs`:

```pyhon
 self.trajectory_kwargs["mode"] = self.trajectory_kwargs.get("mode", "w")
```

This means that if we specify `"w"` mode and there's already existing trajectory files, they will be wiped. Since `reopen_trajectories` is only called **during** optimization/ dynamics, not at the start, I believe it should never wipe trajectories halfway an active trajectory (which is what currently happens before this bugfix). 

So, if you agree that 
1. it is possible to wipe a trajectory via constructing as `trajectory = Trajectory(..., trajectory_kwargs=dict(mode="w"))`, and that
2. we never want to wipe trajectories during an active run,

perhaps the current version of the code is fine? Otherwise please let me know what part I don't yet understand. Thanks!

torch_sim/trajectory.py
danielzuegner: thanks, done!

torch_sim/trajectory.py
danielzuegner: Thanks, added that to the code!

danielzuegner:
> In the case of an interupted simulation it's likely that there will be some files that are complete, some that are incomplete, and some that have not started. We need a way to resume simulations that balances ease, sharp edges, and encouraging best practices.
> 
> Here is what I propose:
> 
> 1. Require that all files in the TrajectoryReporter all have the same number of steps, either 0 (do not exist yet) or N. If a simulation is interrupted, it becomes the responsibility of the user to identify which systems have completed and discard those from the file list. A line in the script that discards completed files can always be included and will just do nothing if it's the first run.
> 2. Provide a utility for resetting many systems to the same step. This can be called externally when needed, but I don't think it should be applied blindly internally, lest we wipe lots of progress. For the partially complete files these can be wiped to the same step and n_steps can be adjusted to account for their length.
> 3. `n_steps` should be the number of new steps not total. I looked at OpenMM, LAMMPS, and ASE and even when appending `n_steps` is the number of new steps, not the total number.
> 
> Does this workflow feel too complicated? Let me know what y'all think.

Hi @orionarcher,

Thanks for the detailed suggestions. That makes sense to me, I'll incorporate these suggestions, though due to vacation might only have the time towards the end of the month. Let's also iron out the last two open discussions in this PR and then hopefully we're good to go :).

Daniel

torch_sim/trajectory.py
orionarcher: Ah! It was I who was misunderstanding. Yeah you are totally right, the trajectories will get wiped whenenver a state converges. Temporarily switching into append mode is a very good catch!

torch_sim/trajectory.py
orionarcher: Got i!

Then perhaps let's just change the name to `last_steps` to indicate that it is a list

```python
    @property
    def last_steps(self) -> list[int]:
```

danielzuegner:
@orionarcher could you take another look? I've changed the behavior to always append `n_steps` in the case of `integrate`. For `optimize` I've kept the behavior as-is because the argument is `max_steps` anyway. I've also added `TrajectoryReporter.truncate_to_step` to provide a higher-level API to make trajectories consistent in terms of number of steps.

The test failures here appear to be unrelated to this PR, feel free to let me know if you think otherwise.

Thanks!

danielzuegner:
@orionarcher ping :)

orionarcher:
Thanks for ping @danielzuegner! This got lost over the holidays, will review later today or tomorrow.

torch_sim/trajectory.py
```diff
@@ -763,10 +763,9 @@ def last_step(self) -> int:
         Returns:
             int: The last recorded step number, or 0 if no data exists
         """
-        if not self.array_registry:
+        if not self.array_registry or "positions" not in self.array_registry:
             return 0
-
-        return self.get_steps("positions")[-1]
+        return self.get_steps("positions")[-1].item()
```
orionarcher: What is the behavior if `positions` aren't reported? I worry it will throw an error. Could you leave the `or "positions" not in self.array_registry` check?

torch_sim/runners.py
```diff
@@ -119,20 +117,18 @@ def _determine_initial_step_for_integrate(
...
+                "You can truncate the trajectories to the same step using:\n\n"
+                "    trajectory_reporter.truncate_to_step(min(trajectory_reporter.last_step))\n\n"
+                "before calling integrate again."
             )
-            for traj in trajectory_reporter.trajectories:
-                traj.truncate_to_step(last_logged_step)
+        print(
+            f"Detected existing trajectory with last step {last_logged_step}."
+            f" Resuming integration from step {initial_step}."
+        )
```
orionarcher: Could you use log.info instead of print here?

orionarcher (approved):
Nicely done @danielzuegner! Thank you for your persistence in completing this PR. This is a very nice addition to the high-level runners.

Just a couple small things but I trust you can resolve them.

torch_sim/trajectory.py
danielzuegner: Current code has the `"positions" not in self.array_registry:` check to return 0 if positions are not recorded. If I understand your comment this is your preference, right? Otherwise please let me know.

danielzuegner: @orionarcher based on my previous comment I think this is resolved (please let me know otherwise), so I think you can go ahead and merge this PR whenever you see fit. Thanks for the discussion and helpful feedback!

orionarcher: Ok got it, makes sense! Will merge now, thank you!

orionarcher:
Excellent work @danielzuegner!

thomasloux:
Thank you @danielzuegner

Issue #89, opened 2025-04-01
Implement LBFGS optimizer


orionarcher:
LBFGS would be an excellent additional optimizer alongside FIRE and gradient descent

corinwagen:
agreed! FIRE is quick but doesn't seem to converge things as tightly as LBFGS

abhijeetgangan:
Do you have a comparison that it does generally better than FIRE?

Here are some references comparing different optimizers:
- https://www.sciencedirect.com/science/article/pii/S0927025620300756#s0010
- https://docs.matlantis.com/atomistic-simulation-tutorial/en/2_3_opt-algorithm.html
- https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.97.170201

FIRE seems to be more robust but can struggle with very small `fmax`. In any case adding LBFGS with linesearch would a good addition.

corinwagen:
Agreed re: FIRE vs BFGS, it does seem generally like small `fmax` can be tough but that FIRE's often more robust. Both seem good to have in the toolbox IMO. 

Here's a concrete case where BFGS performs much better than FIRE—optimization of valinomycin w/ MACE-MP-0 (or MACE-MPA-0, they keep stealth-changing which model is used and I can't keep track since there's no publication haha). Here's valinomycin:

<img width="627" alt="Image" src="https://github.com/user-attachments/assets/084b58f6-bcc4-4ac6-9b9e-a79418d08ce1" />

Script for BFGS, FIRE is the same if you replace `BFGS` with `FIRE`:

```
from ase.optimize import BFGS, FIRE, QuasiNewton
from mace.calculators import mace_mp
from ase.io import read

mol = read("valinomycin.xyz")
mol.calc = mace_mp()

BFGS(mol).run(fmax=0.01)
```

BFGS converges to a final energy of  -993.677917 in 405 steps. FIRE converges to a final energy of -993.098267 in 497 steps, so it doesn't get the right minimum. 

Here's the input file for valinomycin: 

```
168
valinomycin
N        2.98    2.984    1.224
C        2.288    3.512    2.4
C        0.808    3.268    2.176
C        2.768    2.832    3.664
C        4.272    3.148    3.956
C        1.92    3.248    4.876
O        0.312    2.208    1.86
O        0.088    4.372    2.44
C        -1.372    4.204    2.396
C        -1.828    4.144    0.92
C        -1.944    5.468    3.04
O        -3.012    3.704    0.768
[...]

PR #365, opened 2025-11-26
Implement lbfgs and bfgs

abhijeetgangan:
## Summary

* Add L-BFGS and BFGS optimization schemes.

## Checklist

### DIFF
--- a/examples/scripts/2_structural_optimization.py
+++ b/examples/scripts/2_structural_optimization.py
@@ -111,7 +111,7 @@ state = ts.fire_init(state=state, model=lj_model, dt_start=0.005)
 
 # Run optimization
 for step in range(N_steps):
-    if step % 20 == 0:
+    if step % 100 == 0:
         print(f"Step {step}: Potential energy: {state.energy[0].item()} eV")
     state = ts.fire_step(state=state, model=lj_model, dt_max=0.01)
 


abhijeetgangan:
See detailed comment [here](https://github.com/TorchSim/torch-sim/issues/89#issuecomment-3582360508).

torch_sim/optimizers/state.py
```diff
@@ -40,4 +40,84 @@ class FireState(OptimState):
     _system_attributes = OptimState._system_attributes | {"dt", "alpha", "n_pos"}  # noqa: SLF001
 
 
+@dataclass(kw_only=True)
+class BFGSState(OptimState):
```
orionarcher, line 55: According to our convention, these belong in the respective integrator files, not in `state.py`

abhijeetgangan: But the fire and gd states are in `state.py` was this changed?

orionarcher: I don't see those in `state.py`? If they are in there I think they should probably be moved back.

abhijeetgangan: This is on main - https://github.com/TorchSim/torch-sim/blob/4a90c21436e097f409f66dfd1a3fdb003027ffd1/torch_sim/optimizers/state.py#L28

orionarcher: Ahh, you are right sorry. I confused `optimizers/state.py` and `state.py`

abhijeetgangan: We should call it something else? Doesn't really matter but could avoid confusion.

orionarcher: I think it's ok, I probably should have caught it

torch_sim/optimizers/bfgs.py
```diff
@@ -0,0 +1,282 @@
...
+    Updates the Hessian estimate and moves atoms.
+
+    Args:
+        state: Current optimization state
+        model: Calculator model
+
+    Returns:
+        Updated state
+    """
+    eps = 1e-7
```
curtischong: One idea is we can surface this eps up as a param to the function (maybe change it to update_hessian_eps) so people can choose to **always** update the approximated hessian for all systems. If we choose to do this, I think we should maybe increase the + 1e-30 epsilon (when we divide the BFGS update terms by a and b) to 1e-7 (similar to ase - or we do `eps = 1e-8 if dtype == torch.float32 else 1e-16` - similar to our lbfgs implementation) so the blowup isn't as large when we divide by a number near 0

torch_sim/optimizers/lbfgs.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # Build new (s, y)
+    s_new = state.positions - state.prev_positions
+    y_new = -new_forces - (-state.prev_forces)  # g_new - g_prev = -(f_new - f_prev)
+
+    # Curvature check per system; if bad, clear history (conservative)
+    sy = tsm.batched_vdot(s_new, y_new, state.system_idx)
+    bad_curv = sy <= curvature_eps
+
+    if bad_curv.any():
```
curtischong: I think we can replace this section with

```
if bad_curv.any():
    # Clear history for systems with bad curves to preserve correctness
    keep_mask = ~bad_curv[state.system_idx].unsqueeze(1)  # [n_atoms, 1]
    state.s_history = state.s_history * keep_mask
    state.y_history = state.y_history * keep_mask

    s_new = s_new * keep_mask # set to 0 for bad curves (no-op)
    y_new = y_new * keep_mask

# Append and trim if needed
if state.s_history.shape[0] == 0:
    s_hist = s_new.unsqueeze(0)
    y_hist = y_new.unsqueeze(0)
else:
    s_hist = torch.cat([state.s_history, s_new.unsqueeze(0)], dim=0)
    y_hist = torch.cat([state.y_history, y_new.unsqueeze(0)], dim=0)
if s_hist.shape[0] > max_history:
    s_hist = s_hist[-max_history:]
    y_hist = y_hist[-max_history:]
```

so we only clear out the history for systems with bad curves.

torch_sim/optimizers/lbfgs.py
```diff
@@ -0,0 +1,286 @@
...
+        alpha=alpha_tensor,
+        n_iter=torch.zeros((n_systems,), device=model.device, dtype=torch.int32),
+    )
+
+
+def lbfgs_step(  # noqa: PLR0915
+    state: "LBFGSState",
+    model: "ModelInterface",
+    *,
+    max_history: int = 10,
```
curtischong: I feel like we can set this during the init. I don't see people changing this value throughout the optimization. Also, it allows us to preallocate the history tensors

torch_sim/optimizers/lbfgs.py
```diff
@@ -0,0 +1,286 @@
...
+            1.0 / (ys + eps),
+            torch.zeros_like(ys),
+        )
+        yz = tsm.batched_vdot(y_i, z, state.system_idx)
+        beta = rho * yz
+
+        alpha = alphas[state.s_history.shape[0] - 1 - i]
+        # z <- z + s_i * (alpha - beta)
+        coeff = (alpha - beta)[state.system_idx].unsqueeze(-1)
+        z = z + coeff * s_i
```
curtischong: nit: change to z = z + s_i * coeff so it matches the comment

torch_sim/optimizers/lbfgs.py
```diff
@@ -0,0 +1,286 @@
...
+    step = t_atoms * d
+
+    # Per-atom norms
+    norms = torch.linalg.norm(step, dim=1)
+
+    # Per-system max norm
+    sys_max = torch.zeros(state.n_systems, device=device, dtype=dtype)
+    sys_max.scatter_reduce_(0, state.system_idx, norms, reduce="amax", include_self=False)
+
+    # Scaling factors per system: <= 1.0
```
curtischong, line 461: consider saying: "scale down step so atoms move at most max_step" or "# Scale step if it exceeds max_step" like you mention in bfgs

torch_sim/optimizers/bfgs.py
```diff
@@ -0,0 +1,282 @@
...
+
+        state.hessian[idx] = H - term1 - term2
+
+    # Calculate step direction using eigendecomposition
+    # gradient: (n_systems, D, 1)
+    # Step p = H^-1 * F
+    direction = forces_new.unsqueeze(2)
+
+    # omega: (n_systems, D), V: (n_systems, D, D)
+    omega, V = torch.linalg.eigh(state.hessian)
```
curtischong: ik that ase doesn't do this, but I'd appreciate a comment here saying that we do an eigendecomp (rather than simply updating the inverse Hessian directly) so we can get the eigenvalues and do torch.abs(omega) to ensure that the approximated hessian is always positive definite (and we're always going downwards - which is more important for atomistic systems). I was a bit confused when this didn't match the vanilla BFGS algorithm

abhijeetgangan:
Thanks for the comments. I will update those over the weekend.

orionarcher:
@abhijeetgangan is this still on the docket for you?

I know it's a much requested feature so it'd be awesome to get it in. Thanks for your time on it.

abhijeetgangan:
Will prioritize this on the weekend.

abhijeetgangan:
Hi @sihoonchoi and @Andrew-S-Rosen the integration is pretty much done. I also tested on 1K structures on WBM with mace comparing FIRE and L-BFGS (with cell filters):

For fmax of 0.02 eV/A
```
Optimizer    Time (s)     Converged       Max Force (eV/Å)
FIRE         288.55       1000/1000          0.0200         
L-BFGS       101.91       1000/1000          0.0196  
```

For fmax of 1e-4 eV/A

```
Optimizer    Time (s)     Converged       Max Force (eV/Å)
FIRE         744.09       995/1000          0.0145         
L-BFGS       192.27       1000/1000          0.0001 
```

Clearly, here it seem that FIRE can struggle sometimes to relax especially for small fmax. It did not finish all even after 500 steps. BFGS is working but need linesearch to the competitive. That's going to be a separate PR. I am going to merge this as the tests are very comprehensive but would be nice if someone can try it out.

```
import time

import torch
from matbench_discovery.data import DataFiles, ase_atoms_from_zip

import torch_sim as ts


N_STRUCTURES = 1000
MAX_STEPS = 500
FORCE_TOL = 1e-4

dtype = torch.float64
raw_mace = mace_mp(model="small", return_raw_model=True, default_dtype="float64")
mace_model = MaceModel(model=raw_mace, device=device, dtype=dtype, compute_stress=True)

ase_atoms_list = ase_atoms_from_zip(DataFiles.wbm_initial_atoms.path, limit=N_STRUCTURES)
initial_state = ts.io.atoms_to_state(ase_atoms_list, device=device, dtype=dtype)

convergence_fn = ts.generate_force_convergence_fn(
    force_tol=FORCE_TOL, include_cell_forces=True
)

optimizers = [
    ("FIRE", ts.Optimizer.fire, {}),
    ("L-BFGS", ts.Optimizer.lbfgs, {"max_history": 20, "step_size": 0.5}),
]

results = {}

for name, optimizer, init_kwargs in optimizers:
    print(f"Running {name}...")

    state = initial_state.clone()

    autobatcher = ts.autobatching.InFlightAutoBatcher(
        model=mace_model,
        memory_scales_with="n_atoms",
        max_memory_scaler=600.0,
        max_iterations=MAX_STEPS // 2,
    )

    if device.type == "cuda":
        torch.cuda.synchronize()
    start_time = time.perf_counter()

    final_state = ts.optimize(
        system=state,
        model=mace_model,
        optimizer=optimizer,
        max_steps=MAX_STEPS,
        convergence_fn=convergence_fn,
        init_kwargs=dict(cell_filter=ts.CellFilter.frechet, **init_kwargs),
        autobatcher=autobatcher,
    )

    if device.type == "cuda":
        torch.cuda.synchronize()
    elapsed = time.perf_counter() - start_time

    max_forces = torch.norm(final_state.forces, dim=-1).max().item()
    converged = convergence_fn(final_state, None)
    n_converged = int(converged.sum().item())

    results[name] = {
        "time": elapsed,
        "max_force": max_forces,
        "n_converged": n_converged,
    }

    n_total = len(ase_atoms_list)
    print(
        f"Time: {elapsed:.2f}s | Converged: {n_converged}/{n_total} | "
        f"Max force: {max_forces:.4f} eV/Å"
    )

print("SUMMARY")
print(f"{'Optimizer':<12} {'Time (s)':<12} {'Converged':<15} {'Max Force (eV/Å)':<15}")

for name, data in results.items():
    print(
        f"{name:<12} {data['time']:<12.2f} "
        f"{data['n_converged']}/{len(ase_atoms_list):<13} {data['max_force']:<15.4f}"
    )
```

sihoonchoi:
@abhijeetgangan I've tried L-BFGS and BFGS along with FIRE with UMA-OMC on 100 structures and it works well. This is great - thanks for your work!

```
Optimizer    Time (s)     Converged    
FIRE         838.81       100/100          
L-BFGS       530.74       100/100          
BFGS         545.62       100/100          
```

Below is a comparison with serial relaxations in ASE.
```
Optimizer    Time (s)     Converged    
FIRE         1670.04     
L-BFGS       1001.56     
BFGS         1133.86     
```

abhijeetgangan:
@sihoonchoi Thanks for looking into it. This is very valuable! 

There are more improvements that can be made on top of this like introducing line-search and trust regions. Some of which is batching friendly and some of it is not. I am pretty sure the optimization algorithms can be made more robust with it (ASE doesn't use many of these known tricks). I would be curious on what the community thinks but given that there are no benchmarks in matsci it's hard to make comparisons. Maybe guys at Rowan can set this up? Would like to know what others think.

Andrew-S-Rosen:
I personally find it a bit wild that we don't have good benchmarks on this, but I guess it's historically because DFT is expensive to run. Rowan doesn't have a ton of infra for materials but they have obviously thought about this. 

Actually, my group has a funded NSF collaboration with the Argonne folks supporting TorchSim (Ben Blaiszik et al) that might be relevant here. That might be a good avenue to kickstart this. I'd be interested in seeing if we can start something through that and open it up to the community to participate.

abhijeetgangan:
One thing I always found confusing was when comparing MLIPs vs DFT for relaxation. MLIPs would use FIRE but VASP uses CG  and QE uses BFGS both with linesearch and trust regions which makes them very different. Another aspect is the cost. I am sure the devs of DFT codes must have chosen this so that it reduces the DFT calls during relaxation. I think a fair comparison would be ideal but that means going away from exact ASE behavior.

> Actually, my group has a funded NSF collaboration with the Argonne folks supporting TorchSim (Ben Blaizsik et al) that might be relevant here. That might be a good avenue to kickstart this. I'd be interested in seeing if we can start something through that and open it up to the community to participate.

I think this is a great idea! Maybe we can start an email thread to discuss this? I can add more optimizers to test.

Andrew-S-Rosen:
> One thing I always found confusing was when comparing MLIPs vs DFT for relaxation. MLIPs would use FIRE but VASP uses CG and QE uses BFGS both with linesearch and trust regions which makes them very different. Another aspect is the cost. I am sure the devs of DFT codes must have chosen this so that it reduces the DFT calls during relaxation. I think a fair comparison would be ideal but that means going away from exact ASE behavior.

I think the reason people have been using MLIPs with FIRE is simply a cultural thing, not based on much other than the original FIRE paper. FIRE isn't in VASP (without external plugins), but yeah, CG is simple enough and works.

> I think this is a great idea! Maybe we can start an email thread to discuss this? I can add more optimizers to test.

Sure, happy to kick off an email tomorrow. Can you send me an email from your account? I'm not sure what your email is.

abhijeetgangan:
> Sure, happy to kick off an email tomorrow. Can you send me an email from your account? I'm not sure what your email is.

Sounds good! abhijeetgangan@g.ucla.edu

PR #367, opened 2025-11-27
Feat/allow different temperatures ts integrate

thomasloux:
## Summary

Change ts.integrate to accept temperature of size (n_system,), so that its system can have its own temperature.

## Checklist

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -650,3 +650,34 @@ def test_state_to_device_no_side_effects(si_sim_state: SimState) -> None:
             "New state doesn't have correct device!"
         )
         assert si_sim_state is not new_state_gpu, "New state is not a different object!"
+
+
+def test_state_set_cell(ti_sim_state: SimState) -> None:
+    """Test the set_cell method of SimState."""
+    new_cell = (
+        torch.diag_embed(
+            torch.tensor(
+                [3.0, 4.0, 5.0], device=ti_sim_state.device, dtype=ti_sim_state.dtype
+            )
+        )
+        @ ti_sim_state.cell
+    )
+    ase_atoms = ti_sim_state.to_atoms()[0]
+    ti_sim_state.set_cell(new_cell, scale_atoms=True)
+    ase_atoms.set_cell(new_cell[0].T.cpu().numpy(), scale_atoms=True)
+    assert torch.allclose(
+        ti_sim_state.positions.cpu(), torch.from_numpy(ase_atoms.positions)
+    )
+
+    M = torch.tensor(
+        [[[1.0, 0.2, 0.0], [0.1, 1.5, 0.0], [0.0, 0.0, 2.0]]],
+        device=DEVICE,
+        dtype=ti_sim_state.dtype,
+    )
+    new_cell = M @ ti_sim_state.cell
+    ase_atoms = ti_sim_state.to_atoms()[0]
+    ti_sim_state.set_cell(new_cell, scale_atoms=True)
+    ase_atoms.set_cell(new_cell[0].T.cpu().numpy(), scale_atoms=True)
+    assert torch.allclose(
+        ti_sim_state.positions.cpu(), torch.from_numpy(ase_atoms.positions)
+    )
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -251,6 +251,31 @@ class SimState:
         """
         self.cell = value.mT
 
+    def set_cell(
+        self,
+        cell: torch.Tensor,
+        scale_atoms: bool = False,  # noqa: FBT001, FBT002
+    ) -> None:
+        """Set the unit cell of the system, optionally scaling atomic positions.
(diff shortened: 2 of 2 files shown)


thomasloux:
I didn't add in tests, but it's working:

```
import numpy as np
import torch
from ase.build import bulk
from pymatgen.core import Structure

import torch_sim as ts
from torch_sim.trajectory import TorchSimTrajectory, TrajectoryReporter
from torch_sim.units import MetalUnits

n=4
n_steps = 1500
trajectory_file = [f"lj_trajectory{i}.h5md" for i in range(n)]
# report potential energy every 10 steps and kinetic energy every 20 steps
prop_calculators = {
    10: {"potential_energy": lambda state: state.energy},
    20: {
        "kinetic_energy": lambda state: ts.calc_kinetic_energy(
            momenta=state.momenta, masses=state.masses
        ),
        "temperature": lambda state: state.calc_temperature(),
    },
}

reporter = TrajectoryReporter(
    trajectory_file,
    # report state every 10 steps
    state_frequency=10,
    prop_calculators=prop_calculators,
)


lj_model = LennardJonesModel(
    sigma=2.0,  # Å, typical for Si-Si interaction
    epsilon=0.1,  # eV, typical for Si-Si interaction
    device=torch.device("cpu"),
    dtype=torch.float64,
)

si_atoms = bulk("Si", "fcc", a=5.43, cubic=True)
temperatures = (torch.arange(n) + 1) * 500
# temperatures = temperatures.unsqueeze(0).repeat(n_steps, 1)

final_state = ts.integrate(
    system=[si_atoms.copy() for _ in range(n)],
    model=lj_model,
    integrator=ts.Integrator.nvt_langevin,
    n_steps=n_steps,
    temperature=temperatures,
    timestep=0.002,
    trajectory_reporter=reporter,
)

temperatures = []
for filename in trajectory_file:
    with TorchSimTrajectory(filename) as traj:
        temperatures.append(traj.get_array("temperature").flatten())

temperatures = np.stack(temperatures, axis=1)  # shape (n_steps, n)
import matplotlib.pyplot as plt
# smoothen temperatures
temperatures = np.stack([np.convolve(temperatures[:, i], np.ones(20)/20, mode='valid') for i in range(temperatures.shape[1])], axis=1)
plt.plot(temperatures)
```

image

torch_sim/runners.py
```diff
@@ -100,6 +100,67 @@ def _configure_batches_iterator(
...
+
+    if temps.shape[0] == 1:
+        # A single value in a 1-element list/tensor
+        return temps.repeat(n_steps)
+
+    # This assumes that in case n_systems == n_steps, the user wants to apply
+    # different temperatures per system, not per step.
+    if temps.shape[0] == initial_state.n_systems:
+        # Interpret as single-step multi-system temperatures → broadcast over steps
+        return temps.unsqueeze(0).expand(n_steps, -1)  # (n_steps, n_systems)
```
orionarcher: I honestly might just throw a well-documented error here and demand the user provide a 2D tensor if `n_systems == n_steps`. It's a rare edge case and the consequences of incorrectly guessing the default is potentially many hours of wasted debugging.

orionarcher (commented):
Awesome! This is a nice feature. I'd suggest adding a bit more testing fo the different temperature shapes though.

thomasloux:
Note for me: I need to add a test to check that it works with the autobatcher

torch_sim/runners.py
thomasloux: I added a warning, let me know if you want to be more strict and raise an error

thomasloux:
@orionarcher I modified the autobatcher.bin_index which was a list[dict[int, float]] which I think is just the results of the original use of the binpacking from the first version of autobatcher #39. I changed the bin_index to actually be list[list[indices]]. Otherwise
`for state, indices in batcher`: indices would be a super weird like {1: float, 2: float} and so you can't use it directly like: myTensor[indices]

torch_sim/runners.py
```diff
@@ -123,7 +189,11 @@ def integrate[T: SimState](  # noqa: C901
             (init_func, step_func) functions.
         n_steps (int): Number of integration steps
         temperature (float | ArrayLike): Temperature or array of temperatures for each
-            step
+            step or system:
+            Float: used for all steps and systems
+            1D array of length n_steps: used for each step
+            1D array of length n_systems: used for each system
+            2D array of shape (n_steps, n_systems): used for each step and system.
```
orionarcher, line 200: Could we add a test for the 2D array too for completeness?

orionarcher (approved):
Would be nice to have a test with a 2D array but otherwise LGTM

torch_sim/runners.py
thomasloux: Done

torch_sim/runners.py
orionarcher: Sounds good to me

orionarcher:
Ready to merge @thomasloux?

PR #372, opened 2025-12-01
Handle tensor PBC input in FairChemV1Model

WillEngler:
## Summary

We were seeing [test failures for fairchem-legacy](https://github.com/TorchSim/torch-sim/actions/runs/19800780243/job/56727753272#step:9:6001) that looked like this ...

```
>       if self.pbc != state.pbc:
           ^^^^^^^^^^^^^^^^^^^^^
E       RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

It seems like maybe we missed fairchem-legacy when we moved to support mixed PBC in #320? But because that PR was done from a fork, the actual fairchem-legacy tests didn't run. Similarly, in my last PR #369, that was done from a fork too so the contents of the test did not run.

I've gone and taken a stab at updating FairChemV1Model to account for the fact that inheritors of ModelInterface can receive a tensor-valued PBC. But given how legacy fairchem is implemented, this still enforces uniform PBC. (if it gets a tensor it needs to be [True, True, True] or [False, False, False]).

## Checklist

### DIFF
--- a/tests/models/test_fairchem_legacy.py
+++ b/tests/models/test_fairchem_legacy.py
@@ -2,7 +2,9 @@ import os
 import traceback
 
 import pytest
+import torch
 
+import torch_sim as ts
 from tests.conftest import DEVICE
 from tests.models.conftest import (
     consistency_test_simstate_fixtures,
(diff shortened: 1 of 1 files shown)


WillEngler:
Nice, and thanks for adding the tests

PR #373, opened 2025-12-04
Support for electrostatics

orionarcher:
## Summary

- [x] Add charge and spin to SimState
- [x] Add charge and spin support for MaceModel
- [x] Add charge and spin support for FairchemModel

## Checklist

### DIFF
--- a/torch_sim/io.py
+++ b/torch_sim/io.py
@@ -232,21 +232,12 @@ def atoms_to_state(
     if not all(np.all(np.equal(at.pbc, atoms_list[0].pbc)) for at in atoms_list[1:]):
         raise ValueError("All systems must have the same periodic boundary conditions")
 
-    # Extract charge and spin from atoms.info if available
-    charge_list = []
-    spin_list = []
-    for at in atoms_list:
-        # ASE convention: store charge and spin in atoms.info dict
-        charge_list.append(at.info.get("charge", 0.0))
-        spin_list.append(at.info.get("spin", 0.0))
-
-    # Create tensors if any non-zero values exist
-    charge = None
-    spin = None
-    if any(c != 0.0 for c in charge_list):
-        charge = torch.tensor(charge_list, dtype=dtype, device=device)
-    if any(s != 0.0 for s in spin_list):
-        spin = torch.tensor(spin_list, dtype=dtype, device=device)
+    charge = torch.tensor(
+        [at.info.get("charge", 0.0) for at in atoms_list], dtype=dtype, device=device
+    )
+    spin = torch.tensor(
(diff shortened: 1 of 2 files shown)


torch_sim/io.py
```diff
@@ -232,13 +232,31 @@ def atoms_to_state(
...
+        charge_list.append(at.info.get("charge", 0.0))
+        spin_list.append(at.info.get("spin", 0.0))
+
+    # Create tensors if any non-zero values exist
+    charge = None
+    spin = None
+    if any(c != 0.0 for c in charge_list):
+        charge = torch.tensor(charge_list, dtype=dtype, device=device)
+    if any(s != 0.0 for s in spin_list):
+        spin = torch.tensor(spin_list, dtype=dtype, device=device)
```
thomasloux: Potentially I would rewrite that:
charge = torch.tensor([at.info.get("charge", 0.0) for at in atoms_list], dtype=dtype, device=device)
spin = torch.tensor([at.info.get("spin", 0.0) for at in atoms_list], dtype=dtype, device=device)

given the standard value used in SimState. If you still want to deal with standard value in SimState (instead of in io). Maybe just add a condition
if all("charge" in at.info for at in atoms_list). Although this does not totally solve the standard value to use.

torch_sim/models/mace.py
```diff
@@ -330,19 +330,27 @@ def forward(  # noqa: C901
...
+            cell=sim_state.row_vector_cell,
+            positions=sim_state.positions,
+            edge_index=edge_index,
+            unit_shifts=unit_shifts,
+            shifts=shifts,
+        )
+
+        # Add charge and spin (MACE expects "total_charge" and "total_spin")
+        # SimState always has these initialized (defaulting to zeros if not provided)
+        data_dict["total_charge"] = sim_state.charge
```
thomasloux: why not add directly these fields when you create data_dict?

thomasloux (commented):
Look good to me.
Are we sure that the standard value for all of these models is spin=0 charge=0. It would make sense but I'm not sure it's case when I looked at it, but I'm not sure.

curtischong (approved):
I think this looks fine

orionarcher:
> Are we sure that the standard value for all of these models is spin=0 charge=0. It would make sense but I'm not sure it's case when I looked at it, but I'm not sure.

Good question. I am not sure, but it seems like a better default to me than torch.nan. If someone is using one of the models, I think it's reasonable to ask them to attend to setting the charge and spin.

torch_sim/io.py
orionarcher: Good call, done

torch_sim/models/mace.py
orionarcher: good catch

PR #380, opened 2025-12-10
Fix Monte Carlo Swap ragged systems

curtischong:
## Summary

The swap function perviously failed to properly calculate the index of atoms to swap with if there are multiple systems with different numbers of atoms.

I wrote a test to see that it failed, then wrote the fix.

## Checklist


torch_sim/monte_carlo.py
```diff
@@ -87,7 +87,7 @@ def generate_swaps(state: SimState, rng: torch.Generator | None = None) -> torch
     first_index = torch.multinomial(weights, 1, replacement=False, generator=rng)
 
     # Process each system - we need this loop because of ragged systems
-    system_starts = system_lengths.cumsum(dim=0) - system_lengths[0]
```
curtischong, line 90: before, we think that system 2 starts at len(sys1) + len(sys2) - len(sys1) = len(sys2). this is wrong. we need to have proper broadcasting so system2 starts at len(sys1) + len(sys2) - len(sys2). this subtraction does the shift properly

PR #400, opened 2026-01-09
Use upstream NequipTorchSimModel

CompRhys:
Use the TorchSim Calc from https://github.com/mir-group/nequip rather than duplicating code. Not clear how to get dtype from AOTInductor, upstream to match the contract they just set the dtype to float64. Adds warning for end users. Implement basic cache for the compiled models to speed up repeat tests locally.

### DIFF
--- a/torch_sim/models/nequip_framework.py
+++ b/torch_sim/models/nequip_framework.py
@@ -11,30 +11,18 @@ import traceback
 import warnings
 from typing import Any
 
-import torch
-
 
 try:
     from nequip.integrations.torchsim import NequIPTorchSimCalc
 
     # Re-export with backward-compatible name
     class NequIPFrameworkModel(NequIPTorchSimCalc):
-        """NequIP framework model wrapper for torch-sim.
-        
-        NOTE: Nequip framework models always take float64 inputs and then cast
-        internally to the dtype of the compiled AOTInductor model.
+        """NequIP model framework wrapper for torch-sim.
+
+        NOTE: NequIPFrameworkModel.dtype is always set to torch.float64.
+        The AOTInductor may actually contain a different dtype but the
+        model will cast to the correct dtype internally.
         """
-
-        def dtype(self) -> torch.dtype:
-            """The data type of the model."""
-            warnings.warn(
-                "NequIPFrameworkModel.dtype is always set to torch.float64. "
-                "The AOTInductor may actually contain a different dtype but the "
-                "model will cast to the correct dtype internally.",
-                stacklevel=2,
-            )
-            return super().dtype
-
 
 except ImportError as exc:
     warnings.warn(f"NequIP import failed: {traceback.format_exc()}", stacklevel=2)
@@ -42,9 +30,9 @@ except ImportError as exc:
     from torch_sim.models.interface import ModelInterface
 
     class NequIPFrameworkModel(ModelInterface):  # type: ignore[no-redef]
-        """NequIP model wrapper for torch-sim.
+        """NequIP model framework wrapper for torch-sim.
 
-        This class is a placeholder when NequIP is not installed.
+        NOTE:This class is a placeholder when NequIP is not installed.
         It raises an ImportError if accessed.
         """
 


CompRhys:
@cw-tan Can you give this a look over and okay?

torch_sim/models/nequip_framework.py
```diff
@@ -1,352 +1,51 @@
...
-    """
-
-    def __init__(
-        self,
-        model: str | Path | torch.nn.Module | None = None,
-        *,
-        r_max: float,
-        type_names: list[str],
-        device: torch.device | None = None,
-        neighbor_list_fn: Callable = torchsim_nl,
```
CompRhys, line 168: Here this the main regression in so far as we loose the ability to vary the nl if using the nequip implementation upstream.

tests/models/test_nequip_framework.py
```diff
@@ -1,75 +1,107 @@
...
     pytest.skip(
         f"nequip not installed: {traceback.format_exc()}", allow_module_level=True
     )
 
 
-class NequIPUrls(StrEnum):
-    """Checkpoint download URLs for NequIP models."""
-
-    Si = "https://github.com/abhijeetgangan/pt_model_checkpoints/raw/refs/heads/main/nequip/Si.nequip.pth"
+# Cache directory for compiled models (under tests/ for easy cleanup)
```
CompRhys, line 25: Introduced just to speed up repeat testing locally. Nequip tests will take a long time in CI due to needing to compile step. Should consider if we can cache compiled models in CI also.

torch_sim/models/nequip_framework.py
```diff
@@ -1,352 +1,51 @@
...
-    from nequip.scripts._compile_utils import ASE_OUTPUTS, PAIR_NEQUIP_INPUTS
+    from nequip.integrations.torchsim import NequIPTorchSimCalc
+
+    # Re-export with backward-compatible name
+    class NequIPFrameworkModel(NequIPTorchSimCalc):
+        """NequIP model wrapper for torch-sim."""
+
+        def dtype(self) -> torch.dtype:
+            """The data type of the model."""
+            warnings.warn(
```
CompRhys: The dtype is hard coded and so this warning is there to help users debug if this does lead to an issue.

CompRhys:
```Python
# Load model
model = NequIPFrameworkModel.from_compiled_model(
    output_path, device=DEVICE, chemical_species_to_atom_type_map=True
)

# Check dtypes
print(f"NequIPFrameworkModel._dtype: {model._dtype}")

aot_model = model.model.model  # DictInputOutputWrapper -> AOTICompiledModel
aot_metadata = aot_model.get_metadata()
print(f"AOTInductor model_dtype: {aot_metadata.get('model_dtype')}")
```

Actually we can get this but it's unclear whether this is true dtype given that it's hard coded. If we change the test DTYPE to float32 to match the `model_dtype` inside the AOTICompiledModel then the tests fail.

tests/models/test_nequip_framework.py
```diff
@@ -1,84 +1,115 @@
...
-        r_max=r_max,
-        type_names=type_names,
-        device=DEVICE,
+    return NequIPCalculator.from_compiled_model(
+        str(compiled_ase_nequip_model_path), device=DEVICE
     )
-    assert model._device == DEVICE  # noqa: SLF001
 
 
+# NOTE: we take [:-1] to skip benzene. This is because the stress calculation in NequIP
```
CompRhys, line 99: Benzene test failed on github CI runner. Similar numerical issue existed for SevenNet when PBC turned off. Stress is not meaningful when PBC is off.

torch_sim/models/nequip_framework.py
cw-tan: All NequIP framework models take inputs that are in float64 an return outputs that are in float64. Casting to `model_dtype` happens within the model.

torch_sim/models/nequip_framework.py
```diff
@@ -1,352 +1,51 @@
...
+                "NequIPFrameworkModel.dtype is always set to torch.float64. "
+                "The AOTInductor may actually contain a different dtype.",
```
CompRhys:
```suggestion
                "The AOTInductor may actually contain a different dtype but the "
                "model will cast to the correct dtype internally.",
```

torch_sim/models/nequip_framework.py
```diff
@@ -1,352 +1,51 @@
...
 
 try:
-    from nequip.model.inference_models import load_compiled_model
-    from nequip.nn import graph_model
-    from nequip.scripts._compile_utils import ASE_OUTPUTS, PAIR_NEQUIP_INPUTS
+    from nequip.integrations.torchsim import NequIPTorchSimCalc
+
+    # Re-export with backward-compatible name
+    class NequIPFrameworkModel(NequIPTorchSimCalc):
+        """NequIP model wrapper for torch-sim."""
```
CompRhys:
```suggestion
        """NequIP framework model wrapper for torch-sim.
        
        NOTE: Nequip framework models always take float64 inputs and then cast
        internally to the dtype of the compiled AOTInductor model.
        """
```

PR #402, opened 2026-01-14
fixed offsets in merge_constraints

falletta:
## Summary

* Fixed bug in calculating the offsets in `merge_constraints`
* Updated the corresponding unit test to verify the correct behavior

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -783,9 +783,9 @@ def test_merge_constraints(mixed_double_sim_state: ts.SimState) -> None:
         FixCom([0]),
     ]
 
-    # System 2: Fix atom 0 and fix COM for system 0
+    # System 2: Fix atoms 2, 3 and fix COM for system 0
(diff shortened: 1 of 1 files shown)


tests/test_constraints.py
```diff
@@ -769,10 +769,10 @@ def test_atom_indexed_constraint_update_and_select() -> None:
     assert sub_constraint is None
 
 
-def test_merge_constraints(ar_double_sim_state: ts.SimState) -> None:
+def test_merge_constraints(mixed_double_sim_state: ts.SimState) -> None:
```
CompRhys, line 772: here the test failed to discriminate because n1=n0 for the double Ar state.

falletta: Yes, the buggy `merge_constraints` works for `ar_double_sim_state`, but fails for `mixed_double_sim_state`. The fix makes `merge_constraints` work in both cases.

PR #403, opened 2026-01-14
Reduce test wall time (Round 2 Electric Boogaloo)

CompRhys:
Reduce the test matrix size in order to reduce overall test wall time. Leave lowest on py312 and highest on py313 to get best and worse case coverage of being fully upgraded and fully delayed in upgrading.


orionarcher (approved):
LGTM, thanks for getting around to it

PR #404, opened 2026-01-14
Better default force convergence function

orionarcher:
Make sure state has cell_forces attr so the default force convergence function works without and cell filter

## Summary

* This error bit me so I am fixing it.

## Checklist


orionarcher:
> yolo not sure of context, can this be tested?

Looks like the current tests are unhappy... will investigate

orionarcher:
I decided to change the default instead of internal logic. Now the default will work whether or not you have a cell filter.

Also added a failing test that I fixed.

PR #405, opened 2026-01-16
Pin scipy for fairchem tests

CompRhys:
See https://github.com/facebookresearch/fairchem/issues/1717


orionarcher:
Ahh, back to our normal flakey tests

Thanks @CompRhys!

PR #412, opened 2026-01-20
Fix memory scaling calculation for non-periodic boundary conditions

orionarcher:
## Summary

When the MACE model (or other `n_atoms_x_density` based models) were run with non-periodic systems, they would fail because the cell is zeroed out. I am replacing the volume calculation with a bounding box of the positions. Is this a reasonable solution?

## Checklist

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -907,3 +907,57 @@ def test_fire_fixed_cell_unit_cell_consistency(  # noqa: C901
             f"Energy for system {step} doesn't match position only optimization: "
             f"system={energy_unit_cell}, individual={individual_energies_fire[step]}"
         )
+
+
+# Test for charge and spin preservation
+# GitHub Issue https://github.com/TorchSim/torch-sim/issues/389
+@pytest.mark.parametrize(
+    ("optimizer_fn", "cell_filter"),
+    [
+        (ts.Optimizer.fire, None),
+        (ts.Optimizer.gradient_descent, None),
+        (ts.Optimizer.fire, ts.CellFilter.unit),
+        (ts.Optimizer.gradient_descent, ts.CellFilter.frechet),
+    ],
+)
+def test_optimizer_preserves_charge_spin(
+    optimizer_fn: ts.Optimizer,
+    cell_filter: ts.CellFilter | None,
+    ar_supercell_sim_state: SimState,
+    lj_model: ModelInterface,
+) -> None:
+    """Test that optimizers preserve charge and spin through initialization and steps."""
+    # Add perturbation to positions for meaningful optimization
+    ar_supercell_sim_state.positions = (
+        ar_supercell_sim_state.positions
+        + torch.randn_like(ar_supercell_sim_state.positions) * 0.1
+    )
+
+    # Set non-zero charge and spin values
+    original_charge = torch.tensor(
+        [5.0], device=ar_supercell_sim_state.device, dtype=ar_supercell_sim_state.dtype
+    )
+    original_spin = torch.tensor(
+        [6.0], device=ar_supercell_sim_state.device, dtype=ar_supercell_sim_state.dtype
+    )
+    ar_supercell_sim_state.charge = original_charge.clone()
+    ar_supercell_sim_state.spin = original_spin.clone()
+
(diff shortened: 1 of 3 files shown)


torch_sim/autobatching.py
```diff
@@ -363,7 +363,11 @@ def calculate_memory_scaler(
     if memory_scales_with == "n_atoms":
         return state.n_atoms
     if memory_scales_with == "n_atoms_x_density":
-        volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
+        if all(state.pbc):
+            volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
+        else:
+            bbox = state.positions.max(dim=0).values - state.positions.min(dim=0).values
```
CompRhys, line 369: would using this as the general case fail? for example if we had a 2d system or surface with a lot of vacuum the `cell` is not useful for determining the density?

orionarcher: Good point. Neither is perfect in those cases but I agree a bounding box is better than cell. I am happy to make it the general case. Does a clamp value of 2 A make sense to you? Needed for flat molecules like benzene (though benzene isn't actually flat).

CompRhys: Clamp is harder to reason about because there are some systems say metallic lithium where the unit cell is ~1A.

CompRhys: I guess the limiting case is something flat anthracene in a vacuum. If you have two molecules far enough apart then this heuristic would fail.

 n_atoms*density is just to say it scales as the number of nearest neighbors? why not call a nl algorithm to estimate and go based on that?

orionarcher: I think that's right but I am not quite intuiting how that would resolve the 2D vs 3D tradeoff. What would that look like in practice, call the neighbor list with a 5-6 A cutoff and then calculate `number_density` from that?

A couple drawbacks:
- a switch wouldn't be backwards compatible, any users saved `n_atoms_x_density` metrics would need to be recomputed (though that shouldn't stop us)
- it's more expensive and needs to be executed on every system

torch_sim/autobatching.py
```diff
@@ -363,7 +363,11 @@ def calculate_memory_scaler(
     if memory_scales_with == "n_atoms":
         return state.n_atoms
     if memory_scales_with == "n_atoms_x_density":
-        volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
+        if all(state.pbc):
+            volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
```
curtischong, line 367: ik it's not added in this PR, but this 1000 feels like a magic number to me. we should at minimum have a comment explaining it - or make it configurable at most.

torch_sim/autobatching.py
curtischong: I think we should probably use a bounding box algorithm to start (it's very fast&simple logic).

If we do want to do something more complicated (using neighborlists), we'd either have to calculate it twice or refactor a bit of code to get it to work well (since I think we determine the batches before we calculate the neighbors). If we do want to support neighborlist-based memory scaling, we'd probably make a new `memory_scales_with` kind

torch_sim/autobatching.py
orionarcher: ill add a comment, once I remember what the 1000 is for... I think it's an A^3 -> nm^3 conversion

torch_sim/autobatching.py
orionarcher: I agree, I'll just do BB and add 2 A in every non-periodic direction to account for 2D systems and slabs

orionarcher:
I've implemented the bounding box solution + 2 A in every non-periodic direction. I think thats a good intermediate solution that covers the 2D and slab cases reasonably well.

curtischong (commented):
I think we should just remove:
```
if all(state.pbc):
    volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
else:
```
and just always use the for loop you added to calculate the bbox.

This is because when people make a vacuum (e.g. for a catalyst simulation) they may just say that all axes are periodic (when only the x and z axis are periodic). In that case, we will incorrectly calculate the volume

PR #413, opened 2026-01-20
Add charge and spin to common_args dicts (fix #389)

### LINKED ISSUE #389: Bug: Spin and charge get lost in optimizers
Related to #253 #277 , initially adressed in  #373 

The passing of the spin and charge parameters from ASE to the Torchsim SimState works.
When passing this SimState through a single model iteration (tested with fairchem), also everything works fine.

The problem occurs in optimization runs - where the spin/charge information as far as I can see gets lost in the init_fn of the optimizer (I observed this using fire, but probably gradient_descent faces the same problem). 
Suggested fix: Copy spin and charge from the SimState argument to the common_args dict in optimizers/fire.py init_fire().

Definitely need to check if the problem also persists in the step_fn, as attributes are also passed here - I am unsure about this.

orionarcher:
## Summary

Implemented fix for issue #389 identified by @niklashoelter. Solution suggested by @thomasloux.

## Checklist

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -909,9 +909,8 @@ def test_fire_fixed_cell_unit_cell_consistency(  # noqa: C901
         )
 
 
-# Test for charge and spin preservation (GitHub Issue #389)
-
-
+# Test for charge and spin preservation
+# GitHub Issue https://github.com/TorchSim/torch-sim/issues/389
 @pytest.mark.parametrize(
     ("optimizer_fn", "cell_filter"),
     [


tests/test_optimizers.py
```diff
@@ -907,3 +907,58 @@ def energy_converged(current_energy: torch.Tensor, prev_energy: torch.Tensor) ->
             f"Energy for system {step} doesn't match position only optimization: "
             f"system={energy_unit_cell}, individual={individual_energies_fire[step]}"
         )
+
+
+# Test for charge and spin preservation (GitHub Issue #389)
```
curtischong: I'd prefer a link to the issue instead

thomasloux:
@curtischong I think no model on torch_sim currently output anything else than energy, forces and stress ? I don't even know if model taking spin and charges ouput some updates values for spin and charge.

thomasloux (commented):
Looks good to me. Maybe this highlights that we should handle differently the initialization of subclass of SimState. For instance, we have the same issue with StaticState that is used in runners.static

orionarcher:
Going to merge for now and deal with #414 in a future PR, which is a great point.

PR #415, opened 2026-01-21
Feature: Enable user to save initial state of trajectory

danielzuegner:
## Summary

Enables the user to write the initial state for a relaxation/ simulation into the trajectory (for step 0). This means that the trajectory file is "self-contained", i.e., contains **all** steps including the initial state. This is optional and the default is to not save the initial state, so the behavior will be the same as before to the user (unless explicitly passing the flag to save the initial state).

## Checklist

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -132,12 +132,13 @@ def _determine_initial_step_for_integrate(
                 "    reporter.truncate_to_step(min(reporter.last_step))\n\n"
                 "before calling integrate again."
             )
-        logger.info(
-            "Detected existing trajectory with last step %s. Resuming integration "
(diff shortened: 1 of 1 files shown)


CompRhys:
> Great addition, thank you!
> 
> Instead of or addition to adding a kwarg, we could just make `save_initial_state` the default? It seems generally nice to have the initial state and it comes at very little cost. What do you think?

+1 no kwarg just save the zero point and view this as squashing a bug.

danielzuegner:
Thanks for the feedback @orionarcher @CompRhys. I've removed the kwarg such that the initial state is now always stored in `ts.integrate` and `ts.optimize`.

PR #417, opened 2026-01-22
memory_scales_with uses only use bbox if _is_rectangular_cell

curtischong:
## Summary

- If we use bbox for non-rectangular cells, then we may incorrectly say that the volume is larger than it should be.
- So for non-rectangular cells we just use the determinant. This will offer the best volume estimation (especially because people only use a vacuum for rectangular cells)

## Checklist

### DIFF
--- a/torch_sim/autobatching.py
+++ b/torch_sim/autobatching.py
@@ -389,9 +389,8 @@ def _is_rectangular_cell(state: SimState, atol: float = 1e-6) -> bool:
     dot_01 = torch.dot(cell[0], cell[1])
     dot_02 = torch.dot(cell[0], cell[2])
     dot_12 = torch.dot(cell[1], cell[2])
-    return torch.allclose(
-        torch.stack([dot_01, dot_02, dot_12]), torch.zeros(3), atol=atol
-    )
+    dots = torch.stack([dot_01, dot_02, dot_12])
+    return torch.allclose(dots, torch.zeros_like(dots), atol=atol)
 
 
 def estimate_max_memory_scaler(


torch_sim/autobatching.py
```diff
@@ -363,22 +363,36 @@ def calculate_memory_scaler(
...
             bbox = state.positions.max(dim=0).values - state.positions.min(dim=0).values
             # add 2 A in non-periodic directions to account for 2D systems and slabs
             for i, periodic in enumerate(state.pbc):
                 if not periodic:
                     bbox[i] += 2.0
             volume = bbox.prod() / 1000  # convert A^3 to nm^3
+        else:
+            # since this is a non-rectangular cell, it's unlikely for there to be
+            # excess vacuum, so use the determinant of the cell as the volume
+            volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
```
orionarcher, line 378: I don't think this is right?

If all(state.pbc), then the determinant is in general the most accurate measure of density, there is no need to use the bbox and it's more heuristic. If not all(state.pbc) then any attempt to calculate the determinant may fail because the cell might be weirdly defined.

While I agree that the bounding box will fail for non-rectangular and non-periodic cells, I don't think this is the right fix.

curtischong: for catalytic systems (e.g. we have a slab and an adsorbate) some testcases set pbc for all axes to true and have a vacuum for a big area above the slab. Using the determinant will say that the volume is larger than it actually is.

curtischong: "If not all(state.pbc) then any attempt to calculate the determinant may fail because the cell might be weirdly defined." Do you have an example you're thinking of?

I think it's fine because if a parallelogram is PBC for one axis, the determinant is still the volume of that parallelogram

orionarcher: > for catalytic systems (e.g. we have a slab and an adsorbate) some testcases set pbc for all axes to true and have a vacuum for a big area above the slab. Using the determinant will say that the volume is larger than it actually is.

Very true.

> "If not all(state.pbc) then any attempt to calculate the determinant may fail because the cell might be weirdly defined." Do you have an example you're thinking of?

I'm imagining 2D material where the cell happens to be non-rectangular and one dimension is very small

I see your points. At the very least, I'd like for rectangular and periodic cells to still use the determinant, because that requires no heuristics and is more accurate.

curtischong: Ahh never dealt with 2D materials so that scenario didn't occur to me. I am fine not merging this in because we do have mixed PBC support so people can just omit PBC on the y-axis (in the excess volume case I mentioned). I think that using the determinant for parallelepipeds is fine.

Issue #381, opened 2025-12-11
Automatically Fetching Public Pre-trained Mace Models in Mace Interface via Model Name


curtischong:
Similar to how we can pass in a Fairchem model name:

` model = FairChemModel(model="uma-s-1", task_name=task_name, cpu=DEVICE.type == "cpu")`

and it will automatically find the pretrained model and automatically load it, we should do the same with MACE. I've always found having to find the proper imports to write:

```
# Load the MACE "small" foundation model
mace = mace_mp(model="small", return_raw_model=True)
mace_model = MaceModel(
    model=mace,
    device=device,
    dtype=torch.float64,
    compute_forces=True,
)
```

quite annoying. Making this change would also make our tutorials a bit shorter and easier to read!

orionarcher:
Would be awesome to have!

CompRhys:
Duplicate to https://github.com/TorchSim/torch-sim/issues/18

PR #419, opened 2026-01-23
Specify MACE models by URL or by name

curtischong:
Updates the Mace model to allow people to pass in URLs to load models (https://github.com/TorchSim/torch-sim/issues/381) - so we don't need to call mace_mp

- 3_dynamics.py still uses the old load_model method since we're sharing the model in 2 places. Also it's good documentation for how to do it the load_model way as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,10 @@ jobs:
             model: { name: fairchem-legacy, test_path: "tests/models/test_fairchem_legacy.py" }
           - version: { python: '3.13', resolution: highest }
             model: { name: fairchem-legacy, test_path: "tests/models/test_fairchem_legacy.py" }
+          - version: { python: '3.13', resolution: lowest-direct }
+            model: { name: nequip, test_path: "tests/models/test_nequip_framework.py" }
+          - version: { python: '3.13', resolution: highest }
+            model: { name: nequip, test_path: "tests/models/test_nequip_framework.py" }
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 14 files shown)


torch_sim/models/mace.py
```diff
@@ -103,7 +108,7 @@ class MaceModel(ModelInterface):
 
     def __init__(
         self,
-        model: str | Path | torch.nn.Module | None = None,
+        model: str | Path | torch.nn.Module | MACECalculator | None = None,
```
curtischong, line 126: I added the MACECalculator type since `mace_mp` returns it and it's an accepted type to this function

torch_sim/models/mace.py
```diff
@@ -33,11 +33,31 @@
 
 
 try:
+    from mace.calculators.foundations_models import mace_mp
+
+    try:
```
curtischong, line 38: I added this second try catch since I checked to see when was mace_mp_urls added and it was only added 8 months ago, so older mace versions may not have it https://github.com/ACEsuit/mace/commit/ab913f8df166b022874529a16731ae3bf5694d95

tests/test_elastic.py
```diff
@@ -257,10 +255,8 @@ def test_get_elementary_deformations_strain_consistency(
 @pytest.fixture
 def mace_model() -> MaceModel:
     """Create a MACE model fixture for testing."""
-    mace_model = mace_mp(model="medium", default_dtype="float64", return_raw_model=True)
-
     return MaceModel(
-        model=mace_model,
+        model="https://github.com/ACEsuit/mace-mp/releases/download/mace_mp_0/2023-12-03-mace-128-L1_epoch-199.model",
```
curtischong, line 262: we don't want people with older models of mace to not run these tests

curtischong:
closing since we are going to move away from hosting model interfaces here and it doesn't make sense to add more features to something we're going to deprecate https://github.com/TorchSim/torch-sim/issues/428

PR #420, opened 2026-01-23
Replace manual initialization with from_state across integrators and optimizers

### LINKED ISSUE #389: Bug: Spin and charge get lost in optimizers

### LINKED ISSUE #414: No SimState subclass takes automatically the newly added spin and charge
Current initialization of subclass of SimState (MDState, all intergrators init, optimizer init, StaticState) don't take spin and charge. So this information is lost and not provided to the model. Currently a fix is proposed for optimizer #413 but this probably does not scale well. Not sure other information will be added in Atoms any time soon.
So either, the same fix is applied to all init function. Or we refactor init function, for instance adding a class method SubClass.from_simstate(state: SimState, **kwargs) where kwargs correspond to SubClass specific variables.
@orionarcher do you have a preference?

orionarcher:
## Summary

- Fixes #414 by updating `from_state` to allow general state transformations and using it across the package.
- Fixes #389 which was not fully resolved previously. The `FairchemModel` was not correctly passing charge and spin through the interface. 

## Checklist

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -183,8 +183,7 @@ class SimState:
             raise ValueError("All tensors must be on the same device")
 
     @classmethod
-    @property
-    def _all_attributes(cls) -> set[str]:
+    def _get_all_attributes(cls) -> set[str]:
         """Get all attributes of the SimState."""
         return (
             cls._atom_attributes
@@ -234,7 +233,7 @@ class SimState:
     @property
     def attributes(self) -> dict[str, torch.Tensor]:
         """Get all public attributes of the state."""
-        return {attr: getattr(self, attr) for attr in self._all_attributes}
+        return {attr: getattr(self, attr) for attr in self._get_all_attributes()}
 
     @property
     def column_vector_cell(self) -> torch.Tensor:
@@ -396,7 +395,7 @@ class SimState:
         """
         attrs = {}
         for attr_name, attr_value in state.attributes.items():
-            if attr_name in cls._all_attributes:
+            if attr_name in cls._get_all_attributes():
                 if isinstance(attr_value, torch.Tensor):
                     attrs[attr_name] = attr_value.clone()
                 else:


torch_sim/io.py
```diff
@@ -55,6 +56,10 @@ def state_to_atoms(state: "ts.SimState") -> list["Atoms"]:
     system_indices = state.system_idx.detach().cpu().numpy()
     pbc = state.pbc.detach().cpu().numpy()
 
+    # Extract charge and spin if available (per-system attributes)
+    charge = state.charge.detach().cpu().numpy() if state.charge is not None else None
+    spin = state.spin.detach().cpu().numpy() if state.spin is not None else None
```
curtischong: inside state.py, charge is never None. Since we always init it as non-none. We should change the typechecking block inside state.py to say it can never be None

curtischong: that also means that these if statements will never trigger.

curtischong:
this PR is breaking since we change the functionality of `from_state`

torch_sim/io.py
```diff
@@ -68,6 +73,13 @@ def state_to_atoms(state: "ts.SimState") -> list["Atoms"]:
         atoms = Atoms(
             symbols=symbols, positions=system_positions, cell=system_cell, pbc=pbc
         )
+
+        # Preserve charge and spin in atoms.info (as integers for FairChem compatibility)
+        if charge is not None:
+            atoms.info["charge"] = int(charge[sys_idx].item())
```
curtischong, line 79: I think we should change `int` to `float` since we use float in other places in the code when talking about charges/spins

orionarcher: i think we should probably change to `int` it in the other places then. charge and spin are physically integer values.

curtischong: I think we should too. it's a bit annoying how charges is a list of float in ASE https://ase-lib.org/ase/atoms.html

torch_sim/state.py
```diff
@@ -389,13 +390,22 @@ def from_state(cls, state: "SimState", **additional_attrs: Any) -> Self:
             ...     momenta=torch.zeros_like(sim_state.positions),
             ... )
         """
-        # Copy all attributes from the source state
+        # Get attributes that the TARGET class accepts
+        target_attrs = (
```
thomasloux: I feel like we may want to have a "_all_attributes" that would give this dictionary, otherwise one would likely forget about "_constraints"

curtischong: if we do add it to this PR, remember to change `def attributes(self) -> dict[str, torch.Tensor]:` to use it

orionarcher: Good suggestion, i was thinking of that too, I can add it

torch_sim/io.py
CompRhys: > spin are physically integer values.

https://en.wikipedia.org/wiki/Fractionalization .... but here yes

PR #427, opened 2026-01-30
Accelerated ts.static + added scaling scripts

falletta:
## Summary

----------------------
**Changes**
* Optimized the memory scaler and split operations, providing a substantial speedup for ts.static (up to 48% for batches larger than 5000).
* Added scaling scripts for ts.static, ts.relax, and ts.integrate (NVE, NVT) to analyze scaling performance.
* Added tests for memory scaler values for non periodic systems.
* Updated unit tests and tutorials for the batched version of calculate memory scaler.

----------------------
**Results**
The figure below shows the speedup achieved for static evaluations, 10-step atomic relaxation, 10-step NVE MD, and 10-step NVT MD. The test is performed for a 8-atom cubic supercell of MgO using the `mace-mpa` model. Prior results are shown in blue, while new results are shown in red. The speedup is calculated as
`speedup (%) = (baseline_time / current_time − 1) × 100`. We observe that:

* `ts.static` achieves a 52.6% speedup for 100,000 structures
* `ts.relax` achieves a 4.8% speedup for 1,500 structures
* `ts.integrate` (NVE) achieves a 0.9% speedup for 10,000 structures
* `ts.integrate` (NVT) achieves a 1.4% speedup for 10,000 structures

Screenshot 2026-02-03 at 5 04 11 PM

----------------------
**Profiling**
The figure below shows a detailed performance profile. Additional optimization can be achieved by disabling the trajectory reporter when not needed, which will be addressed in a separate PR.

Screenshot 2026-02-03 at 5 47 53 PM

----------------------
**Comments**

From the scaling plots, we can see that the timings of `ts.static` and `ts.integrate` are all consistent with each other. Indeed:
* `ts.static` → 85s for 100'000 evaluations
* `ts.integrate` NVE → 87s for 10'000 structures (10 MD steps each) → 87s for 100'000 evaluations
* `ts.integrate` NVT → 89s for 10'000 structures (10 MD steps each) → 89s for 100'000 evaluations

However, when looking at the relaxation:
* `ts.relax` → 63s for 1'000 structures (10 relax steps each) → 63s for 10'000 evaluations → ~630s for 100'000 evaluations

So `ts.relax` is about **7x slower** than `ts.static` or `ts.integrate`. The unbatched FrechetCellFilter clearly contributes to that, and will be the focus on a separate PR.

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -913,12 +913,10 @@ def _pop_states[T: SimState](
     if isinstance(pop_indices, list):
         pop_indices = torch.tensor(pop_indices, device=state.device, dtype=torch.int64)
 
-    # Build boolean masks then convert to integer indices
-    system_range = torch.arange(state.n_systems, device=state.device)
-    pop_system_mask = torch.isin(system_range, pop_indices)
+    # Derive keep/pop atom and system indices
+    all_systems = torch.arange(state.n_systems, device=state.device)
+    keep_system_indices = all_systems[~torch.isin(all_systems, pop_indices)]
     pop_atom_mask = torch.isin(state.system_idx, pop_indices)
-
-    keep_system_indices = torch.where(~pop_system_mask)[0]
     keep_atom_indices = torch.where(~pop_atom_mask)[0]
     pop_atom_indices = torch.where(pop_atom_mask)[0]
 


torch_sim/autobatching.py
```diff
@@ -359,20 +391,17 @@ def calculate_memory_scaler(
...
-            volume = torch.abs(torch.linalg.det(state.cell[0])) / 1000
-        else:
-            bbox = state.positions.max(dim=0).values - state.positions.min(dim=0).values
-            # add 2 A in non-periodic directions to account for 2D systems and slabs
-            for i, periodic in enumerate(state.pbc):
-                if not periodic:
-                    bbox[i] += 2.0
-            volume = bbox.prod() / 1000  # convert A^3 to nm^3
-        number_density = state.n_atoms / volume.item()
+        # Use cell volume (O(1)); SimState always has a cell. Avoids O(N) position scan.
```
orionarcher: non-periodic systems don't have a sensible cell, see #412

torch_sim/autobatching.py
```diff
@@ -556,11 +585,17 @@ def load_states(self, states: T | Sequence[T]) -> float:
...
-        self.state_slices = states.split() if isinstance(states, SimState) else states
-        self.memory_scalers = [
-            calculate_memory_scaler(state_slice, self.memory_scales_with)
-            for state_slice in self.state_slices
-        ]
+        if isinstance(states, SimState):
+            self.memory_scalers = calculate_batched_memory_scalers(
+                states, self.memory_scales_with
+            )
+            self.state_slices = states.split()
```
orionarcher: batching makes sense here

torch_sim/autobatching.py
```diff
@@ -589,9 +624,14 @@ def load_states(self, states: T | Sequence[T]) -> float:
         )  # list[dict[original_index: int, memory_scale:float]]
         # Convert to list of lists of indices
         self.index_bins = [list(batch.keys()) for batch in self.index_bins]
-        self.batched_states = []
-        for index_bin in self.index_bins:
-            self.batched_states.append([self.state_slices[idx] for idx in index_bin])
+        # Build batches: one slice per bin from batched state (no single-system states)
+        if isinstance(states, SimState):
+            self.batched_states = [[states[index_bin]] for index_bin in self.index_bins]
```
orionarcher: state.split() is identical to this and faster

examples/scaling/scaling_nve.py
```diff
@@ -0,0 +1,76 @@
+"""Scaling for TorchSim NVE."""
```
orionarcher: These will all run in CI, so maybe we could put them all in a single script instead of four to speed up testing?

torch_sim/autobatching.py
falletta: Reusing `self.state_slices` instead of calling `states.split()` again makes the code 5% faster, so I'd keep it

torch_sim/autobatching.py
falletta: I now minimized the differences compared to the initial code

falletta: In addition, I added explicit tests for the memory scaler values and verified that the changes in this PR do not affect the test’s success

examples/scaling/__init__.py
CompRhys, line 1: this file isn't needed

tests/test_autobatching.py
```diff
@@ -95,12 +95,14 @@ def test_calculate_scaling_metric(si_sim_state: ts.SimState) -> None:
...
     n_atoms_metric = calculate_memory_scaler(si_sim_state, "n_atoms")
     assert n_atoms_metric == si_sim_state.n_atoms
+    assert n_atoms_metric == 8
 
     # Test n_atoms_x_density metric
     density_metric = calculate_memory_scaler(si_sim_state, "n_atoms_x_density")
     volume = torch.abs(torch.linalg.det(si_sim_state.cell[0])) / 1000
     expected = si_sim_state.n_atoms * (si_sim_state.n_atoms / volume.item())
     assert pytest.approx(density_metric, rel=1e-5) == expected
+    assert pytest.approx(density_metric, rel=1e-5) == (8**2) * 1000 / (5.43**3)
```
CompRhys: why is this added?

torch_sim/autobatching.py
```diff
@@ -556,11 +591,17 @@ def load_states(self, states: T | Sequence[T]) -> float:
...
-            calculate_memory_scaler(state_slice, self.memory_scales_with)
-            for state_slice in self.state_slices
-        ]
+        if isinstance(states, SimState):
+            self.memory_scalers = calculate_batched_memory_scalers(
+                states, self.memory_scales_with
+            )
+            self.state_slices = states.split()
+        else:
+            self.state_slices = states
```
CompRhys: why not concat and then called the batched logic?

torch_sim/state.py
```diff
@@ -430,16 +430,71 @@ def to_phonopy(self) -> list["PhonopyAtoms"]:
         """
         return ts.io.state_to_phonopy(self)
 
-    def split(self) -> list[Self]:
-        """Split the SimState into a list of single-system SimStates.
+    def split(self) -> Sequence[Self]:  # noqa: C901
```
CompRhys: this whole code block looks hard to understand/maintain. why did the external `_split_state` functional programming method need to be removed? 

This looks like it just creates an efficient slicing iter, can we break those parts out more cleanly in the same functional pattern as before? i.e. `_get_system_slice(sim_state: SimState, i: int)` as a function

torch_sim/autobatching.py
falletta: In the `if` branch, the input is already a single batched SimState, so we call `calculate_batched_memory_scalers` and then `split()` once to get `state_slices`. No concatenation is needed.

In the `else` branch, the input is a list of states, so we keep `state_slices = states` and compute scalers per state. We avoid concatenating and using the batched path, since that would require a `concat` followed by a `split()`, resulting in extra passes and higher peak memory for the same outcome.

torch_sim/state.py
falletta: Ok I got rid of the intermediate class and made the code more functional

falletta:
I’ve implemented the revisions and added an additional optimization to the state creation. Please let me know if you need any further edits.

orionarcher:
I will give another review in the next few days, thanks @fallett!

falletta:
Thank you @orionarcher! I made further edits to follow the original logic more closely. Among the changes, I generalized the existing `calculate_memory_scaler` function to a batched version that now returns a list of floats. This required adjusting the unit tests and tutorials, but I think it’s a cleaner approach that avoids having two subroutines for the memory scaler and improves type checking. In addition, I tried to mimic as close as possible the logic for the split function, the `_split_state` subroutine however still requires substantial changes. Let me know your thoughts and I'm happy to keep revising the PR if necessary.

tests/test_autobatching.py
```diff
@@ -94,13 +94,15 @@ def test_calculate_scaling_metric(si_sim_state: ts.SimState) -> None:
     """Test calculation of scaling metrics for a state."""
     # Test n_atoms metric
     n_atoms_metric = calculate_memory_scaler(si_sim_state, "n_atoms")
-    assert n_atoms_metric == si_sim_state.n_atoms
+    assert n_atoms_metric == [si_sim_state.n_atoms]
+    assert n_atoms_metric == [8]
```
orionarcher: let's get rid of all the hardcoded values in the tests, it just makes them more brittle

torch_sim/autobatching.py
```diff
@@ -325,58 +325,39 @@ def determine_max_batch_size(
...
+                volume = torch.abs(torch.linalg.det(s.cell[0])) / 1000
+            else:
+                bbox = s.positions.max(dim=0).values - s.positions.min(dim=0).values
+                for i, periodic in enumerate(s.pbc):
+                    if not periodic:
+                        bbox[i] += 2.0
+                volume = bbox.prod() / 1000
+            scalers.append(s.n_atoms * (s.n_atoms / volume.item()))
+        return scalers
+    raise ValueError(f"Invalid metric: {memory_scales_with}")
```
orionarcher: the new error value just strips important context, why change it?

torch_sim/autobatching.py
```diff
@@ -325,58 +325,39 @@ def determine_max_batch_size(
...
-            volume = bbox.prod() / 1000  # convert A^3 to nm^3
-        number_density = state.n_atoms / volume.item()
-        return state.n_atoms * number_density
-    raise ValueError(
-        f"Invalid metric: {memory_scales_with}, must be one of {get_args(MemoryScaling)}"
-    )
+        if state.n_systems > 1 and state.pbc.all().item():  # vectorized path
+            n = state.n_atoms_per_system.to(state.volume.dtype)
+            vol = torch.abs(state.volume) / 1000  # A^3 -> nm^3
+            return torch.where(vol > 0, n * n / vol, n).tolist()
```
orionarcher: why is the variable volume elsewhere but vol here?

would also change n -> n_atoms

torch_sim/autobatching.py
```diff
@@ -325,58 +325,39 @@ def determine_max_batch_size(
 def calculate_memory_scaler(
```
orionarcher: I suggest changing the name to calculate_memory_scalers since this now returns a list

torch_sim/autobatching.py
```diff
@@ -589,9 +570,8 @@ def load_states(self, states: T | Sequence[T]) -> float:
         )  # list[dict[original_index: int, memory_scale:float]]
         # Convert to list of lists of indices
         self.index_bins = [list(batch.keys()) for batch in self.index_bins]
-        self.batched_states = []
-        for index_bin in self.index_bins:
-            self.batched_states.append([self.state_slices[idx] for idx in index_bin])
+        # Build batches: one sliced state per bin
+        self.batched_states = [[batched[index_bin]] for index_bin in self.index_bins]
```
orionarcher: why use batched instead of state_slices which is already split?

torch_sim/state.py
```diff
@@ -137,34 +137,35 @@ def __post_init__(self) -> None:  # noqa: C901
...
-            self.charge = torch.zeros(
-                self.n_systems, device=self.device, dtype=self.dtype
-            )
-        elif self.charge.shape[0] != self.n_systems:
-            raise ValueError(f"Charge must have shape (n_systems={self.n_systems},)")
+            self.charge = torch.zeros(n_systems_val, device=self.device, dtype=self.dtype)
+        elif self.charge.shape[0] != n_systems_val:
+            raise ValueError(f"Charge must have shape (n_systems={n_systems_val},)")
         if self.spin is None:
-            self.spin = torch.zeros(self.n_systems, device=self.device, dtype=self.dtype)
```
orionarcher: why is any of this logic changed? seems unrelated to rest of PR

torch_sim/state.py
```diff
@@ -644,6 +645,58 @@ def deform_grad(self) -> torch.Tensor:
         return self._deform_grad(self.reference_row_vector_cell, self.row_vector_cell)
 
 
+def _split_state[T: SimState](state: T) -> Sequence[T]:
+    """Return a lazy Sequence view of state split into single-system states.
+
+    Each single-system state is created on first access, so the call is O(1).
```
orionarcher: I guess I don't really follow how this is faster. It just shifts the cost to later down the line by making the evaluation lazy instead of eager, which isn't necessarily better given the code is harder to follow. Could you explain how this is saving time?

torch_sim/state.py
```diff
@@ -934,12 +917,12 @@ def _slice_state[T: SimState](state: T, system_indices: list[int] | torch.Tensor
     """Slice a substate from the SimState containing only the specified system indices.
 
     Creates a new SimState containing only the specified systems, preserving
-    all relevant properties.
+    their requested order (not natural 0,1,2 order).
```
orionarcher: what does this mean?

torch_sim/state.py
```diff
@@ -1111,8 +1124,8 @@ def concatenate_states[T: SimState](  # noqa: C901, PLR0915
 
 def initialize_state(
     system: StateLike,
-    device: torch.device | None = None,
-    dtype: torch.dtype | None = None,
+    device: torch.device,
+    dtype: torch.dtype,
```
orionarcher: why?

torch_sim/state.py
orionarcher, line 937: the rewrite of this function adds a lot of code and it's unclear it makes anything faster

examples/scaling/scaling_nve.py
falletta: Actually, these tests should not run in CI given the changes in `.github/workflows/test.yml`. In their current form, these are not really intended as tests (there are no assert statements), but rather as performance checks. I think keeping them separate makes it easier to isolate the scaling behavior we want to examine for potential optimizations. 

That said, we could turn these into actual tests and include them in the `tests` repository, for example by checking that the timings stay below the current thresholds. Would you prefer that approach?

tests/test_autobatching.py
falletta: sure!

torch_sim/autobatching.py
falletta: sure!

torch_sim/autobatching.py
falletta: sure!

torch_sim/autobatching.py
falletta: `batched[index_bin]` does a single tensor slice operation under the hood. If you used `state_slices`, you'd have to pull individual `SimState` objects from a Python list and then call `concatenate_states()` to merge them back - that's extra Python loop overhead plus a concatenation step you can skip entirely.

torch_sim/state.py
falletta: The `n_systems` property calls `torch.unique()` every time it's accessed. The old code called this property 4-5 times during `__post_init__`. The new code compute the value once into a local variable (`n_systems_val`) and reuses it, avoiding redundant `unique()` calls. This speeds up state initialization, especially for large batches. It’s a minor optimization, but I think it's still relevant.

torch_sim/state.py
falletta: The eager implementation creates all `N` `SimState` objects upfront by running `torch.split()` across every attribute, which is `O(N)` work regardless of how many states are actually used. The lazy implementation only computes a cumsum (effectively `O(1)`) and defers creating states until `__getitem__` is called.

This matters because `estimate_max_memory_scaler` only accesses 2 states (argmin/argmax) out of potentially hundreds—so lazy evaluation builds 2 states, while eager still builds all `N`.

Benchmarking with `n=10000` systems shows the lazy version finishes in **9.89 s** vs **11.37 s** for eager (~15% faster). If all states are accessed, the total work is basically the same, but the common autobatching paths don’t require all states to be materialized at once.

torch_sim/state.py
falletta: see comment above

torch_sim/state.py
falletta: That refers to the fact that calling `_slice_state(state, [3, 1, 4])` returns a new state where the systems appear in the exact order requested (3→0, 1→1, 4→2), instead of being sorted in ascending order. I updated the docstring to better reflect this point.

torch_sim/state.py
falletta: This is a minor fix, but I think making `device` and `dtype` required improves type checking and forces callers to be explicit. With `None` defaults, `.to(None, None)` is a no-op, so the state silently stays where it was, potentially causing device mismatches with the model later.

falletta:
Thanks Orion for reviewing. I applied a few cosmetic changes to address your comments. Regarding your points:

1. Great!
2. The changes to the state manipulations are essential, as described in my comments. The scaling scripts provide a benchmark that highlights the performance speedup.
3. All of the changes were intentional and came out of carefully profiling many functions in the code—from the major bottlenecks in `memory_scaler` and `split` to smaller optimizations in the state initialization.

torch_sim/autobatching.py
falletta: sure!

torch_sim/state.py
orionarcher: > This matters because estimate_max_memory_scaler only accesses 2 states (argmin/argmax) out of potentially hundreds—so lazy evaluation builds 2 states, while eager still builds all N.

I don't follow? How can you get the argmin over values that haven't been evaluated? Don't you need to evaluate to know the minimum?

orionarcher: Could you point me to where the eager and lazy are benchmarked against eachother?

examples/scaling/scaling_nve.py
orionarcher: I do worry about these decaying over time if they aren't tested. The reason to run all scripts in CI is because it makes sure they don't break over time. We shouldn't include them as tests but they should be runnable scripts like the other scripts we have.

falletta: Right, good point. I merged all the tests into `examples/scripts/8.scaling.py` and set the maximum number of structures to relatively small values so the test runs quickly.

torch_sim/state.py
falletta: Please find the benchmark script below. The script defines an `OldState(SimState)` class using the old `_split_state` method and produces the results shown. Alternatively, you can run `8.scaling.py` with the current version of the code and with the split method reverted, and you’ll get the same results.

```
=== Comparison ===
  n=   1: eager=2.387301s, lazy=0.026688s, speedup=89.45x
  n=   1: eager=0.275248s, lazy=0.022556s, speedup=12.20x
  n=   1: eager=0.677568s, lazy=0.022692s, speedup=29.86x
  n=   1: eager=0.023758s, lazy=0.021832s, speedup=1.09x
  n=  10: eager=0.283295s, lazy=0.025066s, speedup=11.30x
  n= 100: eager=0.700301s, lazy=0.108458s, speedup=6.46x
  n= 250: eager=1.020541s, lazy=0.252591s, speedup=4.04x
  n= 500: eager=0.572549s, lazy=0.506986s, speedup=1.13x
  n=1000: eager=1.115220s, lazy=0.989014s, speedup=1.13x
  n=1500: eager=1.665096s, lazy=1.469927s, speedup=1.13x
  n=5000: eager=5.755523s, lazy=5.062751s, speedup=1.14x
  n=10000: eager=10.948717s, lazy=9.643398s, speedup=1.14x
```

**Benchmark Script**
```
"""Test comparing OldState (eager split) vs State (lazy split) performance."""

import time
import typing
from typing import Self
from unittest.mock import patch
import torch
from ase.build import bulk

import torch_sim as ts
from torch_sim.models.mace import MaceModel, MaceUrls
from torch_sim.state import SimState, get_attrs_for_scope


DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
DTYPE = torch.float64
N_STRUCTURES = [1, 1, 1, 1, 10, 100, 250, 500, 1000, 1500, 5000, 10000]
MAX_MEMORY_SCALER = 400_000
MEMORY_SCALES_WITH = "n_atoms_x_density"



class OldState(SimState):
    """Old state representation that uses eager splitting.

    This class inherits from SimState but overrides split() to use the old eager
    approach where all sub-states are created upfront when split() is called.
    """

    def split(self) -> list[Self]:
        """Split the OldState into a list of single-system OldStates (EAGER).

        This is the OLD approach that creates ALL states upfront, which is O(n)
        where n is the number of systems.

        Returns:
            list[OldState]: A list of OldState objects, one per system
        """
        return self._split_state(self)
            
    @staticmethod
    def _split_state[T: SimState](state: T) -> list[T]:
        """Split a SimState into a list of states, each containing a single system.

        Divides a multi-system state into individual single-system states, preserving
        appropriate properties for each system.

        Args:
            state (SimState): The SimState to split

        Returns:
            list[SimState]: A list of SimState objects, each containing a single
                system
        """
        system_sizes = state.n_atoms_per_system.tolist()

        split_per_atom = {}
        for attr_name, attr_value in get_attrs_for_scope(state, "per-atom"):
            if attr_name != "system_idx":
                split_per_atom[attr_name] = torch.split(attr_value, system_sizes, dim=0)

        split_per_system = {}
        for attr_name, attr_value in get_attrs_for_scope(state, "per-system"):
            if isinstance(attr_value, torch.Tensor):
                split_per_system[attr_name] = torch.split(attr_value, 1, dim=0)
            else:  # Non-tensor attributes are replicated for each split
                split_per_system[attr_name] = [attr_value] * state.n_systems

        global_attrs = dict(get_attrs_for_scope(state, "global"))

        # Create a state for each system
        states: list[T] = []
        n_systems = len(system_sizes)
        zero_tensor = torch.tensor([0], device=state.device, dtype=torch.int64)
        cumsum_atoms = torch.cat((zero_tensor, torch.cumsum(state.n_atoms_per_system, dim=0)))
        for sys_idx in range(n_systems):
            # Build per-system attributes (padded attributes stay padded for consistency)
            per_system_dict = {
                attr_name: split_per_system[attr_name][sys_idx]
                for attr_name in split_per_system
            }

            system_attrs = {
                # Create a system tensor with all zeros for this system
                "system_idx": torch.zeros(
                    system_sizes[sys_idx], device=state.device, dtype=torch.int64
                ),
                # Add the split per-atom attributes
                **{
                    attr_name: split_per_atom[attr_name][sys_idx]
                    for attr_name in split_per_atom
                },
                # Add the split per-system attributes (with unpadding applied)
                **per_system_dict,
                # Add the global attributes
                **global_attrs,
            }

            atom_idx = torch.arange(cumsum_atoms[sys_idx], cumsum_atoms[sys_idx + 1])
            new_constraints = [
                new_constraint
                for constraint in state.constraints
                if (new_constraint := constraint.select_sub_constraint(atom_idx, sys_idx))
            ]

            system_attrs["_constraints"] = new_constraints
            states.append(type(state)(**system_attrs))  # type: ignore[invalid-argument-type]

        return states


    @classmethod
    def from_sim_state(cls, state: SimState) -> "OldState":
        """Create OldState from a SimState."""
        return cls(
            positions=state.positions.clone(),
            masses=state.masses.clone(),
            cell=state.cell.clone(),
            pbc=state.pbc.clone() if isinstance(state.pbc, torch.Tensor) else state.pbc,
            charge=state.charge.clone() if state.charge is not None else None,
            spin=state.spin.clone() if state.spin is not None else None,
            system_idx=state.system_idx.clone() if state.system_idx is not None else None,
            _constraints=state.constraints.copy(),
        )


def run_torchsim_static(
    n_structures_list: list[int],
    base_structure,
    model,
    device: torch.device,
) -> list[float]:
    """Run static calculations for each n using batched path, return timings."""
    autobatcher = ts.BinningAutoBatcher(
        model=model,
        max_memory_scaler=MAX_MEMORY_SCALER,
        memory_scales_with=MEMORY_SCALES_WITH,
    )
    times: list[float] = []
    for n in n_structures_list:
        structures = [base_structure] * n
        if device.type == "cuda":
            torch.cuda.synchronize()
        t0 = time.perf_counter()
        ts.static(structures, model, autobatcher=autobatcher)
        if device.type == "cuda":
            torch.cuda.synchronize()
        elapsed = time.perf_counter() - t0
        times.append(elapsed)
        print(f"  n={n} static_time={elapsed:.6f}s")
    return times
  


if __name__ == "__main__":

    # Setup
    mgo_atoms = bulk("MgO", crystalstructure="rocksalt", a=4.21, cubic=True)

    print("Loading MACE model...")
    loaded_model = mace_mp(
        model=MaceUrls.mace_mpa_medium,
        return_raw_model=True,
        default_dtype="float64",
        device=str(DEVICE),
    )
    mace_model = MaceModel(
        model=typing.cast("torch.nn.Module", loaded_model),
        device=DEVICE,
        compute_forces=True,
        compute_stress=True,
        dtype=DTYPE,
        enable_cueq=False,
    )

    print("\n=== Static Benchmark Comparison: Eager vs Lazy Split ===\n")

    # Run with eager split (OldState behavior)
    print("Running with eager split (OldState):")
    with patch.object(SimState, "split", lambda self: OldState._split_state(self)):
        eager_times = run_torchsim_static(
            N_STRUCTURES, mgo_atoms, mace_model, DEVICE
        )

    # Run with lazy split (default State behavior)
    print("\nRunning with lazy split (State):")
    lazy_times = run_torchsim_static(
        N_STRUCTURES, mgo_atoms, mace_model, DEVICE
    )

    # Print comparison
    print("\n=== Comparison ===")
    for i, n in enumerate(N_STRUCTURES):
        speedup = eager_times[i] / max(lazy_times[i], 1e-9)
        print(
            f"  n={n:4d}: eager={eager_times[i]:.6f}s, "
            f"lazy={lazy_times[i]:.6f}s, speedup={speedup:.2f}x"
        )

```

falletta: The `argmin` runs over `metric_values`, which is just a list of floats computed from the batched state tensors before we split anything. This gives us integer indices, and only when we call `state_list[idx]`we do actually create the 2 needed states.

torch_sim/state.py
orionarcher: Got it, makes sense

torch_sim/state.py
orionarcher: Thanks for the detailed benchmark and explanation, that speedup makes sense then!

That said, the type("SplitSeq", (Sequence,), ...) pattern for creating a runtime class is pretty unusual and hard to follow. Could we get the same lazy benefit with a simpler design? 

CC suggestion: A _LazySplitView class with explicit __len__ and __getitem__ would be much more readable and debuggable while preserving the lazy semantics.

torch_sim/state.py
orionarcher: Got it. I prefer this behaviour and I am happy to leave this in this PR but I am noting that 1) this is a breaking change and 2) this is somewhat unrelated to the other changes in this PR.

I am a bit surprised that the old tests didn't catch this. Could you write a test or two that illustrates the difference and that fails with the old behavior and passes with the new?

torch_sim/state.py
orionarcher: Let's not introduce a breaking API change here. If you feel strongly on this let's do it elsewhere.

orionarcher:
I better understand the optimization now, thanks for the clarifications.

> All of the changes were intentional and came out of carefully profiling many functions in the code

I don't see how this is true for the `_slice_state` or `initialize_state` changes, but everything else seems well-founded in optimization. Let's keep the `_slice_state` with an added test and drop the `initialize_state` change.

torch_sim/state.py
falletta: Ok, I’m now using a `LazySplitView` class, and I’ve confirmed that performance is preserved.

torch_sim/state.py
falletta: Sure, I reverted it back to a less strict type check.

orionarcher:
Also maybe rename 8_scaling.py folder to 8_benchmarking.py?

EDIT: Sorry, didn't mean to close! Misclick

torch_sim/state.py
falletta: Great idea. I added two dedicated tests to `test_state.py`, namely `test_slice_state_permuted_order` and `test_slice_state_reversed_subset`. They pass with the new `_slice_state` implementation and fail with the old one. Below is a standalone test script that confirms the behavior:

```
"""Test that the new _slice_state preserves system ordering while the old
mask-based implementation does not.
"""

import pytest
import torch

import torch_sim as ts
from torch_sim.state import (
    SimState,
    _filter_attrs_by_mask,
    _slice_state,
)


def _old_slice_state(state: SimState, system_indices):
    """Old _slice_state that uses boolean masks — loses ordering."""
    if isinstance(system_indices, list):
        system_indices = torch.tensor(
            system_indices, device=state.device, dtype=torch.int64
        )
    if len(system_indices) == 0:
        raise ValueError("system_indices cannot be empty")

    system_range = torch.arange(state.n_systems, device=state.device)
    system_mask = torch.isin(system_range, system_indices)
    atom_mask = torch.isin(state.system_idx, system_indices)

    filtered_attrs = _filter_attrs_by_mask(state, atom_mask, system_mask)
    return type(state)(**filtered_attrs)


@pytest.fixture
def _three_system_state(
    si_sim_state: SimState,
    ar_supercell_sim_state: SimState,
    fe_supercell_sim_state: SimState,
) -> tuple[SimState, SimState, SimState, SimState]:
    """Concatenate Si, Ar, Fe into a 3-system state."""
    concatenated = ts.concatenate_states(
        [si_sim_state, ar_supercell_sim_state, fe_supercell_sim_state]
    )
    return concatenated, si_sim_state, ar_supercell_sim_state, fe_supercell_sim_state


# --- New _slice_state: these pass ---


def test_permuted_order(_three_system_state) -> None:
    """Slice with [2, 0, 1] should yield [Fe, Si, Ar]."""
    concatenated, si, ar, fe = _three_system_state
    result = _slice_state(concatenated, [2, 0, 1])

    assert torch.allclose(result[0].positions, fe.positions)
    assert torch.allclose(result[1].positions, si.positions)
    assert torch.allclose(result[2].positions, ar.positions)
    assert torch.allclose(result.cell[0], fe.cell[0])
    assert torch.allclose(result.cell[1], si.cell[0])
    assert torch.allclose(result.cell[2], ar.cell[0])


def test_reversed_subset(_three_system_state) -> None:
    """Slice with [2, 0] should match concatenate([Fe, Si])."""
    concatenated, si, _ar, fe = _three_system_state
    result = _slice_state(concatenated, [2, 0])
    expected = ts.concatenate_states([fe, si])

    assert torch.allclose(result.positions, expected.positions)
    assert torch.allclose(result.cell, expected.cell)
    assert torch.allclose(result.atomic_numbers, expected.atomic_numbers)
    assert torch.allclose(result.masses, expected.masses)
    assert result.n_systems == expected.n_systems


# --- Old _slice_state: identical tests, expected to fail ---


@pytest.mark.xfail(
    raises=(AssertionError, RuntimeError),
    reason="Old mask-based _slice_state does not preserve ordering",
    strict=True,
)
def test_old_permuted_order(_three_system_state) -> None:
    """Slice with [2, 0, 1] should yield [Fe, Si, Ar]."""
    concatenated, si, ar, fe = _three_system_state
    result = _old_slice_state(concatenated, [2, 0, 1])

    assert torch.allclose(result[0].positions, fe.positions)
    assert torch.allclose(result[1].positions, si.positions)
    assert torch.allclose(result[2].positions, ar.positions)
    assert torch.allclose(result.cell[0], fe.cell[0])
    assert torch.allclose(result.cell[1], si.cell[0])
    assert torch.allclose(result.cell[2], ar.cell[0])


@pytest.mark.xfail(
    raises=(AssertionError, RuntimeError),
    reason="Old mask-based _slice_state does not preserve ordering",
    strict=True,
)
def test_old_reversed_subset(_three_system_state) -> None:
    """Slice with [2, 0] should match concatenate([Fe, Si])."""
    concatenated, si, _ar, fe = _three_system_state
    result = _old_slice_state(concatenated, [2, 0])
    expected = ts.concatenate_states([fe, si])

    assert torch.allclose(result.positions, expected.positions)
    assert torch.allclose(result.cell, expected.cell)
    assert torch.allclose(result.atomic_numbers, expected.atomic_numbers)
    assert torch.allclose(result.masses, expected.masses)
    assert result.n_systems == expected.n_systems



```

falletta:
Thanks for reviewing, @orionarcher! I’ve applied a few additional changes to `8.benchmarking.py` to align with the existing examples for running in CI, please double check. Otherwise, it should be all set on my end. Let me know if there’s anything else to address.

torch_sim/state.py
falletta: To further complement this point, below is a script demonstrating that the new `__post_init__` speeds up initialization by up to 4X compared to the previous implementation. In absolute terms, the performance gain is marginal, but worth keeping

```
"""Benchmark n_systems caching: local var (new) vs property (old) in __post_init__."""

import time

import torch

from torch_sim.state import SimState

from torch_sim.constraints import validate_constraints

def _old_post_init(self) -> None:
    """Initialize the SimState and validate the arguments."""
    # Check that positions, masses and atomic numbers have compatible shapes
    shapes = [
        getattr(self, attr).shape[0]
        for attr in ("positions", "masses", "atomic_numbers")
    ]

    if len(set(shapes)) > 1:
        raise ValueError(
            f"Incompatible shapes: positions {shapes[0]}, "
        )

    if isinstance(self.pbc, bool):
        self.pbc = [self.pbc] * 3
    if not isinstance(self.pbc, torch.Tensor):
        self.pbc = torch.tensor(self.pbc, dtype=torch.bool, device=self.device)

    initial_system_idx = self.system_idx
    if initial_system_idx is None:
        self.system_idx = torch.zeros(
            self.n_atoms, device=self.device, dtype=torch.int64
        )
    else:  # assert that system indices are unique consecutive integers
        _, counts = torch.unique_consecutive(initial_system_idx, return_counts=True)
        if not torch.all(counts == torch.bincount(initial_system_idx)):
            raise ValueError("System indices must be unique consecutive integers")

    if self.constraints:
        validate_constraints(self.constraints, state=self)

    if self.charge is None:
        self.charge = torch.zeros(
            self.n_systems, device=self.device, dtype=self.dtype
        )
    elif self.charge.shape[0] != self.n_systems:
        raise ValueError(f"Charge must have shape (n_systems={self.n_systems},)")
    if self.spin is None:
        self.spin = torch.zeros(self.n_systems, device=self.device, dtype=self.dtype)
    elif self.spin.shape[0] != self.n_systems:
        raise ValueError(f"Spin must have shape (n_systems={self.n_systems},)")

    if self.cell.ndim != 3 and initial_system_idx is None:
        self.cell = self.cell.unsqueeze(0)

    if self.cell.shape[-2:] != (3, 3):
        raise ValueError("Cell must have shape (n_systems, 3, 3)")

    if self.cell.shape[0] != self.n_systems:
        raise ValueError(
            f"Cell must have shape (n_systems, 3, 3), got {self.cell.shape}"
        )

    # if devices aren't all the same, raise an error, in a clean way
    devices = {
        attr: getattr(self, attr).device
        for attr in (
            "positions",
            "masses",
            "cell",
            "atomic_numbers",
            "pbc",
            "system_idx",
        )
    }
    if len(set(devices.values())) > 1:
        raise ValueError("All tensors must be on the same device")

# The "new" post_init is simply the current SimState.__post_init__
_new_post_init = SimState.__post_init__


def _time_construction(
    raw_kwargs: dict,
    post_init_fn,
    n_warmup: int = 3,
    n_trials: int = 20,
) -> tuple[float, float]:
    original = SimState.__post_init__
    SimState.__post_init__ = post_init_fn

    try:
        for _ in range(n_warmup):
            SimState(**raw_kwargs)

        times = []
        for _ in range(n_trials):
            t0 = time.perf_counter()
            SimState(**raw_kwargs)
            t1 = time.perf_counter()
            times.append(t1 - t0)
    finally:
        SimState.__post_init__ = original

    t = torch.tensor(times)
    return t.mean().item(), t.std().item()


if __name__ == "__main__":
    atoms_per_system = 8
    system_counts = [1, 10, 50, 100, 500, 1_000, 5_000, 10_000, 100_000]

    header = (
        f"{'n_atoms':>10} | "
        f"{'old (ms)':>16} | "
        f"{'new (ms)':>16} | "
        f"{'speedup':>8}"
    )
    sep = "-" * len(header)
    print(header)
    print(sep)

    for n_sys in system_counts:
        n_atoms = n_sys * atoms_per_system
        dtype = torch.float64

        positions = torch.randn(n_atoms, 3, device=device, dtype=dtype)
        masses = torch.ones(n_atoms, device=device, dtype=dtype) * 28.0
        atomic_numbers = torch.full(
            (n_atoms,), 14, device=device, dtype=torch.int64
        )
        cell = torch.eye(3, device=device, dtype=dtype).unsqueeze(0).expand(n_sys, 3, 3)
        cell = cell.contiguous() * 5.43
        pbc = torch.tensor([True, True, True], device=device, dtype=torch.bool)
        system_idx = torch.repeat_interleave(
            torch.arange(n_sys, device=device, dtype=torch.int64),
            atoms_per_system,
        )

        raw = dict(
            positions=positions,
            masses=masses,
            cell=cell,
            pbc=pbc,
            atomic_numbers=atomic_numbers,
            system_idx=system_idx,
        )

        old_mean, old_std = _time_construction(raw, _old_post_init)
        new_mean, new_std = _time_construction(raw, _new_post_init)

        speedup = old_mean / new_mean if new_mean > 0 else float("inf")

        print(
            f"{n_atoms:>10} | "
            f"{old_mean * 1e3:>8.3f} ± {old_std * 1e3:>4.3f} | "
            f"{new_mean * 1e3:>8.3f} ± {new_std * 1e3:>4.3f} | "
            f"{speedup:>7.2f}x"
        )

    print(sep)
```

Results:
```
   n_atoms |         old (ms) |         new (ms) |  speedup
-----------------------------------------------------------
         8 |    0.226 ± 0.007 |    0.116 ± 0.005 |    1.95x
        80 |    0.231 ± 0.003 |    0.120 ± 0.002 |    1.92x
       400 |    0.229 ± 0.005 |    0.120 ± 0.003 |    1.91x
       800 |    0.227 ± 0.003 |    0.119 ± 0.003 |    1.91x
      4000 |    0.430 ± 0.006 |    0.119 ± 0.002 |    3.60x
      8000 |    0.422 ± 0.007 |    0.121 ± 0.003 |    3.49x
     40000 |    0.484 ± 0.003 |    0.144 ± 0.002 |    3.37x
     80000 |    0.481 ± 0.008 |    0.139 ± 0.004 |    3.47x
    800000 |    0.620 ± 0.002 |    0.158 ± 0.001 |    3.93x
-----------------------------------------------------------
```

falletta:
Sure, thanks for the suggestion, this solution seems indeed cleaner. Please have another look, it looks like we are converging

torch_sim/autobatching.py
```diff
@@ -845,8 +847,9 @@ def load_states(self, states: Sequence[T] | Iterator[T] | T) -> None:
             so any ongoing processing will be restarted when this method is called.
         """
         if isinstance(states, SimState):
-            states = states.split()
-        if isinstance(states, list | tuple):
+            sim_state = states
+            states = (sim_state[i] for i in range(sim_state.n_systems))
+        if not isinstance(states, Iterator):
             states = iter(states)
```
orionarcher, line 852: why not use split here? this is a splitting operation. this change makes no sense to me

torch_sim/autobatching.py
```diff
@@ -962,7 +966,9 @@ def _get_first_batch(self) -> T:
             )
             self.max_memory_scaler = self.max_memory_scaler * self.max_memory_padding
             newer_states = self._get_next_states()
-            states = [*states, *newer_states]
+            if newer_states:
+                return ts.concatenate_states([concatenated_state, *newer_states])
+            return concatenated_state
         return ts.concatenate_states([first_state, *states])
```
orionarcher: three return statements feels very noisy and hard to follow, there must be a better way

torch_sim/state.py
```diff
@@ -137,34 +137,35 @@ def __post_init__(self) -> None:  # noqa: C901
             self.system_idx = torch.zeros(
                 self.n_atoms, device=self.device, dtype=torch.int64
             )
+            n_systems_val = 1
         else:  # assert that system indices are unique consecutive integers
             _, counts = torch.unique_consecutive(initial_system_idx, return_counts=True)
+            n_systems_val = len(counts)
```
orionarcher: just call it `n_systems`

orionarcher (commented):
Few comments to keep code readability and quality high but otherwise I think this is looking pretty much there

torch_sim/autobatching.py
falletta: Sure!

torch_sim/autobatching.py
falletta: the three return statements are not collapsed into one

falletta:
Addressed the last comments, performance is maintained and all tests pass. Thanks for the careful review @orionarcher, feel free to merge

torch_sim/autobatching.py
```diff
@@ -962,8 +966,10 @@ def _get_first_batch(self) -> T:
             )
             self.max_memory_scaler = self.max_memory_scaler * self.max_memory_padding
             newer_states = self._get_next_states()
-            states = [*states, *newer_states]
-        return ts.concatenate_states([first_state, *states])
+            if newer_states:
+                all_states = [concatenated_state, *newer_states]
+
+        return ts.concatenate_states(all_states)
```
orionarcher, line 970: I don't think you need this if statement because `newer_state` should always be a list (double check that). If so, this is just needless complexity.

orionarcher: Why do we have a concatenation operation twice? Can we move it up so it only has to be called once?

I hate to pick on this but is core code and maintaining a high standard for simplicity is important. At the end of the day, that's what makes LLM frameworks so effective in this codebase. Please spend some time with this and get it to a point where it doesn't have uneeded calls and checks.

falletta: Ok, I got rid of the if statement

falletta: Ok, done! No worries at all, I agree with you and feel free to point out anything else that catches your eye

orionarcher: There are still two concatenation operations

orionarcher: lines 958 and 971

orionarcher: idk if there is a commit that didn't get pushed but please do confirm that the changes were actually made before saying they have been

falletta: I did reduce the number of calls to `concatenate_states`. If you're asking whether we can get it down to a single call entirely, I'm not sure it's feasible without refactoring `estimate_max_memory_scaler`. Currently the first concatenation is needed because that function indexes into a batched state to extract the smallest and largest systems for trial forward passes, which refines the memory budget and allows more states to be added. The second concatenation then includes those additional states in the final batch.

falletta: but if you have some ideas in mind, please suggest the code and I'll test it

falletta:
Thanks @orionarcher for the effort, feel free to merge.

Issue #423, opened 2026-01-27
Wrapped/unwrapped positions produce different results


chlwjd1234:
Using Sevennet and MACE, I observed that simulation results change depending on whether atomic positions are wrapped into the cell.

This can be easily reproduced by the following code and two equivalent but wrapped/unwrapped structure files.

<details>
<summary>POSCAR_wrapped</summary>

```
POSCAR file written by OVITO Basic 3.14.1
1
12.1886482239 0.0 0.0
0.0 12.1886482239 0.0
0.0 0.0 12.1886482239
Si C N 
40 40 40 
Cartesian
1.62960804 7.81111288 11.52471447
0.10904183 11.84466934 10.90873718
9.15765572 0.42337838 10.97032547
9.93355465 3.08735299 10.84395313
0.22936374 2.51467252 0.00688695
6.10603428 2.66809821 6.50035667
9.18716908 5.29740238 0.84695166
3.63698077 10.76674652 3.37370563
10.13137341 1.99530399 4.41941452
11.22895622 0.70702899 1.9626857
6.85616112 8.43644238 2.80080009
1.92299962 0.67620301 4.25293064
7.53706646 4.58360958 11.76811314
11.18088341 4.62180805 4.59584332
0.04918295 3.16794014 8.98231316
0.84643483 9.23329067 6.38460732
5.9513917 11.80080032 3.85442424
10.62050819 9.05090523 4.90256119
2.18369842 4.12739229 3.88087988
10.20464706 10.77087212 8.8745594
2.48243117 7.23433065 1.58311677
7.89527082 3.53034306 4.04023361
4.13870525 9.32768822 11.5232563
1.26553833 4.66457653 6.83879137
3.13174224 6.05740261 11.15625477
11.0031147 6.96509409 11.2675209
3.01780796 4.68837023 7.95611
8.68793774 3.49162698 8.47819042
9.20871639 4.19576883 5.84572124
2.04565835 2.17499804 6.02265453
10.26705742 8.612607 8.92945766
6.72325945 1.3569802 10.16003227
8.28885269 9.26705647 0.35722363
11.58921146 3.34976077 2.98282909
5.72420597 6.34954786 10.00457859
4.19156075 6.19931269 3.00485206
6.07885361 11.66264629 8.30989075
3.57719469 6.17771053 6.01888514
6.44051933 6.75858593 5.59578705
1.38739586 4.79369783 11.62009716
9.65887547 6.37036753 4.19613981
[...]

PR #436, opened 2026-01-31
Handle wrapping correctly for default neighborlist

abhijeetgangan:
Wraps positions inside classical iteraction models and MACE to address #423

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,21 +1,34 @@
 from typing import Any
 
-import numpy as np
 import pytest
 import torch
-import torch.distributions.weibull
 from ase import Atoms
 from ase.build import bulk, molecule
-from ase.spacegroup import crystal
 from phonopy.structure.atoms import PhonopyAtoms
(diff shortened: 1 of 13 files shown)


CompRhys (approved):

torch_sim/models/mace.py
```diff
@@ -300,9 +300,21 @@ def forward(  # noqa: C901
         ):
             self.setup_from_system_idx(sim_state.atomic_numbers, sim_state.system_idx)
 
+        # Wrap positions into the unit cell
```
CompRhys, line 303: what about the alternative of moving thing inside the nl function?

abhijeetgangan: Yes, I tried that but it seems that then both the model and nl would require to use the same coodinates. I.e wrapping for nl vs non-wrapping for model doesn't work (eg. computing edge vectors). This is a issue with nl. I will raise an issue about this as I believe the nl should give the same graph irrespective of wrapping or not.

PR #438, opened 2026-02-03
Feature: Add fix symmetry constraint

danielzuegner:
## Summary
This PR adds a `FixSymmetry` constraint inspired by the way `ASE` implements it. Disclaimer: I used an LLM to support this PR, and vetted the code quite thoroughly.

## Checklist

### DIFF
--- a/torch_sim/constraints.py
+++ b/torch_sim/constraints.py
@@ -136,6 +136,7 @@ class Constraint(ABC):
         """
 
     @classmethod
+    @abstractmethod
     def merge(cls, constraints: list[Self]) -> Self:
         """Merge multiple already-reindexed constraints into one.
 
@@ -145,10 +146,13 @@ class Constraint(ABC):
         Args:
             constraints: Constraints to merge (all same type, already reindexed)
         """
-        raise NotImplementedError(
-            f"Constraint type {cls.__name__} does not implement merge. "
-            "Override this method to support state concatenation."
-        )
+
+
+def _cumsum_with_zero(tensor: torch.Tensor) -> torch.Tensor:
+    """Cumulative sum with a leading zero, e.g. [3, 2, 4] -> [0, 3, 5, 9]."""
+    return torch.cat(
+        [torch.zeros(1, device=tensor.device, dtype=tensor.dtype), tensor.cumsum(dim=0)]
+    )
 
 
 def _mask_constraint_indices(idx: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
@@ -372,22 +376,12 @@ def merge_constraints(
 
     # Calculate cumulative offsets for atoms and systems
     device, dtype = num_atoms_per_state.device, num_atoms_per_state.dtype
-    atom_offsets = torch.cat(
-        [
-            torch.zeros(1, device=device, dtype=dtype),
-            torch.cumsum(num_atoms_per_state[:-1], dim=0),
-        ]
-    )
+    atom_offsets = _cumsum_with_zero(num_atoms_per_state[:-1])
     if num_systems_per_state is None:
         num_systems_per_state = torch.ones(
             len(constraint_lists), device=device, dtype=dtype
         )
-    system_offsets = torch.cat(
-        [
-            torch.zeros(1, device=device, dtype=dtype),
-            torch.cumsum(num_systems_per_state[:-1], dim=0),
-        ]
-    )
+    system_offsets = _cumsum_with_zero(num_systems_per_state[:-1])
 
     # Reindex each constraint to global coordinates, then group by type
     grouped: dict[type[Constraint], list[Constraint]] = defaultdict(list)
@@ -777,13 +771,7 @@ class FixSymmetry(SystemConstraint):
         from torch_sim.symmetrize import prep_symmetry, refine_and_prep_symmetry
 
         rotations, symm_maps = [], []
-        n_per = state.n_atoms_per_system
-        cumsum = torch.cat(
-            [
-                torch.zeros(1, device=state.device, dtype=torch.long),
-                torch.cumsum(n_per, dim=0),
-            ]
-        )
+        cumsum = _cumsum_with_zero(state.n_atoms_per_system)
 
         for sys_idx in range(state.n_systems):
             start, end = cumsum[sys_idx].item(), cumsum[sys_idx + 1].item()
@@ -862,9 +850,9 @@ class FixSymmetry(SystemConstraint):
         for ci, si in enumerate(self.system_idx):
             cur_cell = state.row_vector_cell[si]
             new_row = new_cell[si].mT  # column → row convention
-            deform_delta = torch.linalg.inv(cur_cell) @ new_row - identity
+            deform_delta = torch.linalg.solve(cur_cell, new_row) - identity
             max_delta = torch.abs(deform_delta).max().item()
-            if max_delta > 0.25:
+            if not (max_delta <= 0.25):  # catches NaN via negated comparison
                 raise RuntimeError(
                     f"FixSymmetry: deformation gradient {max_delta:.4f} > 0.25 "
                     f"too large. Use smaller optimization steps."
@@ -877,12 +865,7 @@ class FixSymmetry(SystemConstraint):
         """Symmetrize a rank-1 tensor in-place for each constrained system."""
         from torch_sim.symmetrize import symmetrize_rank1
 
-        cumsum = torch.cat(
-            [
-                torch.zeros(1, device=state.device, dtype=torch.long),
-                torch.cumsum(state.n_atoms_per_system, dim=0),
-            ]
-        )
+        cumsum = _cumsum_with_zero(state.n_atoms_per_system)
         dtype = vectors.dtype
         for ci, si in enumerate(self.system_idx):
             start, end = cumsum[si].item(), cumsum[si + 1].item()
@@ -902,8 +885,8 @@ class FixSymmetry(SystemConstraint):
     def reindex(self, atom_offset: int, system_offset: int) -> Self:  # noqa: ARG002
         """Return copy with system indices shifted by system_offset."""
         return type(self)(
-            self.rotations,
-            self.symm_maps,
+            list(self.rotations),
+            list(self.symm_maps),
             self.system_idx + system_offset,
             adjust_positions=self.do_adjust_positions,
             adjust_cell=self.do_adjust_cell,
@@ -914,6 +897,15 @@ class FixSymmetry(SystemConstraint):
         """Merge by concatenating rotations, symm_maps, and system indices."""
         if not constraints:
             raise ValueError("Cannot merge empty constraint list")
+        if any(
+            c.do_adjust_positions != constraints[0].do_adjust_positions
+            or c.do_adjust_cell != constraints[0].do_adjust_cell
+            for c in constraints[1:]
+        ):
+            raise ValueError(
+                "Cannot merge FixSymmetry constraints with different "
+                "adjust_positions/adjust_cell settings"
+            )
         rotations = [r for c in constraints for r in c.rotations]
         symm_maps = [s for c in constraints for s in c.symm_maps]
         system_idx = torch.cat([c.system_idx for c in constraints])
@@ -935,10 +927,10 @@ class FixSymmetry(SystemConstraint):
         mask = torch.isin(self.system_idx, keep)
         if not mask.any():
             return None
-        indices = [idx for idx in range(len(mask)) if mask[idx]]
+        local_idx = mask.nonzero(as_tuple=False).flatten().tolist()
         return type(self)(
-            [self.rotations[idx] for idx in indices],
-            [self.symm_maps[idx] for idx in indices],
+            [self.rotations[idx] for idx in local_idx],
+            [self.symm_maps[idx] for idx in local_idx],
             _mask_constraint_indices(self.system_idx[mask], system_mask),
             adjust_positions=self.do_adjust_positions,
             adjust_cell=self.do_adjust_cell,
--- a/torch_sim/optimizers/cell_filters.py
+++ b/torch_sim/optimizers/cell_filters.py
@@ -65,6 +65,8 @@ def _get_constrained_stress(
     model_output: dict[str, torch.Tensor], state: SimState
 ) -> torch.Tensor:
     """Clone stress from model output and apply constraint symmetrization."""
+    if not state.constraints:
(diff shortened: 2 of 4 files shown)


torch_sim/constraints.py
```diff
@@ -296,35 +399,32 @@ def merge_constraints(
...
-                )
-            constraint_indices[type(constraint)].append(idxs + offset)
+            constraints, indices = constraints_by_type[type(constraint)]
+            constraints.append(constraint)
+            indices.append(state_idx)
+
+    # Merge each group using the constraint's merge method
+    result = []
+    for constraint_type, (constraints, state_indices) in constraints_by_type.items():
+        merged = constraint_type.merge(constraints, state_indices, atom_offsets)
```
orionarcher: `merge` is a nice addition to the contraint API

pyproject.toml
```diff
@@ -44,8 +44,10 @@ test = [
     "pymatgen>=2025.6.14",
     "pytest-cov>=6",
     "pytest>=8",
+    "spglib>=2.6",
 ]
 io = ["ase>=3.26", "phonopy>=2.37.0", "pymatgen>=2025.6.14"]
+symmetry = ["spglib>=2.6"]
```
thomasloux: What about having spglib in deps directly?

pyproject.toml
```diff
@@ -44,8 +44,10 @@ test = [
     "pymatgen>=2025.6.14",
     "pytest-cov>=6",
     "pytest>=8",
+    "spglib>=2.6",
```
thomasloux, line 48: pymatgen already depends on spglib

torch_sim/optimizers/cell_filters.py
```diff
@@ -222,9 +233,10 @@ def unit_cell_step[T: AnyCellState](state: T, cell_lr: float | torch.Tensor) ->
 
     # Update cell from new positions
     cell_update = cell_positions_new / cell_factor_expanded
-    state.row_vector_cell = torch.bmm(
-        state.reference_cell.mT, cell_update.transpose(-2, -1)
-    )
+    new_cell = torch.bmm(state.reference_cell.mT, cell_update.transpose(-2, -1))
+
+    # Apply cell constraints (in-place, column vector convention)
+    state.set_constrained_cell(new_cell.mT.clone())
```
thomasloux: not sure clone is needed here

torch_sim/optimizers/cell_filters.py
```diff
@@ -249,15 +261,23 @@ def frechet_cell_step[T: AnyCellState](state: T, cell_lr: float | torch.Tensor)
     new_row_vector_cell = torch.bmm(
         state.reference_cell.mT, deform_grad_new.transpose(-2, -1)
     )
-    state.row_vector_cell = new_row_vector_cell
+
+    # Apply cell constraints (in-place, column vector convention)
+    state.set_constrained_cell(new_row_vector_cell.mT.clone())
```
thomasloux: same for clone

torch_sim/optimizers/fire.py
```diff
@@ -405,17 +405,15 @@ def _ase_fire_step[T: "FireState | CellFireState"](  # noqa: C901, PLR0915
...
         cur_deform_grad = cell_filters.deform_grad(
             state.reference_cell.mT, state.row_vector_cell
         )
-        state.set_constrained_positions(
-            torch.linalg.solve(
-                cur_deform_grad[state.system_idx], state.positions.unsqueeze(-1)
-            ).squeeze(-1)
-            + dr_atom
-        )
+        frac_positions = torch.linalg.solve(
```
thomasloux, line 412: Good catch, I was a bit brutal replacing all `state.positions` by the constrained version

torch_sim/constraints.py
```diff
@@ -198,6 +255,29 @@ def select_sub_constraint(
...
+            state_indices: Index of the source state for each constraint
+            atom_offsets: Cumulative atom counts for offset calculation
+
+        Returns:
+            A single merged constraint with adjusted atom indices
+        """
+        all_indices = []
+        for constraint, state_idx in zip(constraints, state_indices, strict=False):
+            offset = atom_offsets[state_idx]
+            all_indices.append(constraint.atom_idx + offset)
```
thomasloux: The constraint in a state is supposed to already have the correct atom_idx (done when batching the list of SimState). So in this case you only need to concatenate the indices.

torch_sim/constraints.py
```diff
@@ -280,6 +360,29 @@ def select_sub_constraint(
...
+            state_indices: Index of the source state for each constraint
+            atom_offsets: Cumulative atom counts (unused for SystemConstraint)
+
+        Returns:
+            A single merged constraint with adjusted system indices
+        """
+        all_indices = []
+        for constraint, state_idx in zip(constraints, state_indices, strict=False):
+            # For SystemConstraint, the offset is the state index itself
+            all_indices.append(constraint.system_idx + state_idx)
```
thomasloux: Same, no need for the offset

torch_sim/constraints.py
thomasloux: Ok just saw that you replaced the merging of constraint with this code.

torch_sim/constraints.py
```diff
@@ -615,3 +743,494 @@ def validate_constraints(constraints: list[Constraint], state: SimState) -> None
...
+        FixSymmetry doesn't explicitly remove DOF in the same way as FixAtoms.
+        This matches ASE's FixSymmetry behavior which also raises NotImplementedError.
+
+        Args:
+            state: Simulation state
+
+        Raises:
+            NotImplementedError: FixSymmetry does not support DOF counting.
+        """
+        raise NotImplementedError("FixSymmetry does not implement get_removed_dof.")
```
thomasloux: return 0, otherwise state.get_number_of_degrees_of_freedom which call constraint.get_removed_dof will fail. This is used for any temperature computation

thomasloux: Although this is only a problem when running MD, not optimization, which is typically how you want to use it. So I'm ok with both option

torch_sim/constraints.py
```diff
@@ -615,3 +743,494 @@ def validate_constraints(constraints: list[Constraint], state: SimState) -> None
...
+            # Get current cell for this system
+            cell = state.row_vector_cell[sys_idx_global]
+
+            # Symmetrize directly
+            symmetrized = symmetrize_rank1(
+                cell,
+                sys_vectors,
+                self.rotations[sys_idx_local].to(dtype=dtype),
+                self.symm_maps[sys_idx_local],
+            )
```
thomasloux: Maybe I don't understand what's happening, but to me:
- sys_idx_local: goes from 0 to n_atoms
- self.rotations and self.sym_maps are of size n_systems.
Aren't you applying multiple times the same constraint?

torch_sim/optimizers/cell_filters.py
```diff
@@ -110,7 +110,13 @@ def unit_cell_filter_init[T: AnyCellState](
     model_output = model(state)
 
     # Calculate initial cell forces
-    stress = model_output["stress"]
+    stress = model_output["stress"].clone()
+
+    # Apply stress constraints if any exist
+    if state.constraints:
```
thomasloux: For this case, maybe implementing a `ts.constraints.constrain_stress` or maybe rather `state.constrain_stress(stress)` could be relevant

torch_sim/constraints.py
```diff
@@ -615,3 +743,494 @@ def validate_constraints(constraints: list[Constraint], state: SimState) -> None
...
+        for sys_idx_local, sys_idx_global in enumerate(self.system_idx):
+            # Get current cell and symmetrize stress directly
+            cur_cell = state.row_vector_cell[sys_idx_global]
+            sys_stress = stress[sys_idx_global]
+            symmetrized = symmetrize_rank2(
+                cur_cell, sys_stress, self.rotations[sys_idx_local].to(dtype=dtype)
+            )
+            stress[sys_idx_global] = symmetrized
+
+    def _symmetrize_rank1(self, state: SimState, vectors: torch.Tensor) -> None:
```
thomasloux: same sys_idx_local

torch_sim/symmetrize.py
```diff
@@ -0,0 +1,464 @@
...
+            scaled_positions=scaled_positions,
+            atomic_numbers=single_state.atomic_numbers,
+            symprec=symprec,
+        )
+        datasets.append(dataset)
+
+    return datasets
+
+
+def _get_scaled_positions(
```
thomasloux: already in transforms.get_fractional_coordinates

torch_sim/symmetrize.py
```diff
@@ -0,0 +1,464 @@
+"""Symmetry refinement utilities for crystal structures.
+
+This module provides functions for refining and symmetrizing atomic structures
+using spglib. It is adapted from ASE's spacegroup.symmetrize module but
+reimplemented to work with torch tensors directly.
+
+The main entry point is `refine_symmetry` which symmetrizes both the cell
+and atomic positions according to the detected space group symmetry.
+"""
```
thomasloux: Could you just add somewhere the meaning of n_ops, I though at first that you meant n_systems. And also that all of these functions are not batched, I prefer to make sure people understand that it's not batched here.

thomasloux (changes requested):
Great work, I'm mainly dubious about the sys_idx_local, sys_idx_global when applying the constraint, but maybe I just misunderstood something.
I haven't really deeped dive the algorithm though, but the code seems overall correct.

pyproject.toml
CompRhys: `spglib` -> `moyo` will happen I believe https://github.com/spglib/moyo

CompRhys: can implementation be changed to use moyo?

janosh:
thanks a lot for this contribution @danielzuegner. super useful! just took it for a spin and encountered a bug: `FixSymmetry.merge()` produces duplicate system indices when used with `InFlightAutoBatcher`

```
ValueError: Duplicate system indices found in `SystemConstraint`.
```

`FixSymmetry.merge()` uses `state_indices` as offsets for system indices. works for single-system constraints but breaks for multi-system ones.

Example: merging constraint_A (5 systems) with constraint_B (3 systems) using `state_indices=[0, 1]` produces `system_indices=[0,1,2,3,4, 1,2,3]` — duplicates at 1, 2, 3.

[`c6a8d0f`](https://github.com/janosh/torch-sim/commit/c6a8d0f) changes to a cumulative running offset instead of `state_indices`, so merged indices are always `[0,1,...,N-1]`.

```diff
-        for constraint, offset in zip(constraints, state_indices, strict=False):
-            for i in range(len(constraint.rotations)):
-                system_indices.append(offset + i)
+        cumulative_offset = 0
+        for constraint in constraints:
+            for idx in range(len(constraint.rotations)):
+                system_indices.append(cumulative_offset + idx)
+            cumulative_offset += len(constraint.rotations)
```

i verified on 1, 100, and 10k structures with autobatching — all work now and symmetries go up, i did not encounter a case where symmetry broke

janosh:
found another issue (cc @abhijeetgangan) `lbfgs_step` (merged just yesterday https://github.com/TorchSim/torch-sim/pull/365) was still  assigning `state.row_vector_cell` directly, bypassing `set_constrained_cell()` so `adjust_cell` was never called. Changed to `set_constrained_cell(new_col_vector_cell, scale_atoms=True)` matching `fire_step`.

janosh:
i think 73532ac addressed most (or all) of the outstanding PR comments

| Reviewer | Comment | Action |
|---|---|---|
| **CompRhys** | [Switch from spglib to moyopy](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2771222626) | Rewrote `symmetrize.py` to use moyopy's `MoyoDataset` for all symmetry detection. `refine_symmetry` reimplemented using metric tensor polar decomposition + position orbit averaging (no longer depends on spglib internals). |
| **thomasloux** | [`get_removed_dof` should return 0, not raise](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2769637772) | Changed to return `torch.zeros(n_systems)` so temperature calculations work in MD. Updated test accordingly. |
| **thomasloux** | [`.clone()` needed in cell_filters?](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2769313945) | Yes — `adjust_cell` modifies tensors in-place, and `.mT` is a view. Without `.clone()`, `state.reference_cell` could be corrupted. Kept as-is. |
| **thomasloux** | [Use `transforms.get_fractional_coordinates`](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2769699765) | Added cross-reference in docstring. Kept local `_get_scaled_positions` since it's unbatched-only (simpler). |
| **thomasloux** | [Add docs about n_ops and non-batched nature](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2769737016) | Added to module docstring of `symmetrize.py`. |
| **thomasloux** | [`sys_idx_local` naming confusion](https://github.com/TorchSim/torch-sim/pull/438#discussion_r2769674541) | Renamed to `constraint_idx` / `sys_idx` — makes clear that `constraint_idx` indexes into `self.rotations`/`self.symm_maps` while `sys_idx` indexes into the state. |

danielzuegner:
Thanks everyone! Let me know if there's anything left for me to do.

janosh:
while doing some cleanup/refactoring, i just noticed that `FixCom.merge([c1, c2])` with multi-system constraints produces `[0, 1, 1, 2]` which hits the same duplicate validation bug already fixed for `FixSymmetry` in [c6a8d0f](https://github.com/TorchSim/torch-sim/pull/438/commits/c6a8d0f50e4107f5671618cd7c32ae802b2bd5d4). i changed `FixCom` to use cumulative offsets as well (i.e. same approach as `FixSymmetry.merge`).

thomasloux:
@janosh Could you add some tests for that? Would be great, this is clearly something we don't want to mess up !

janosh:
seems like the current `merge()` conflates two distinct responsibilities: reindexing (shifting indices to global coordinates) and concatenating (combining multiple constraints into one). The `state_indices` parameter means different things for different constraint types, and `FixSymmetry` ignores it entirely to do its own thing. i split this into [cea861d](https://github.com/TorchSim/torch-sim/pull/438/commits/cea861d29ecbc89431887074827682b154bcb4b8) two distinct methods `reindex()` + `merge()` and line count in `torch_sim/constraints.py` went down by 45 lines as it does away with the brittle cumulative offsets counting i mentioned. usually a good sign for a method split if line count goes down

PR #441, opened 2026-02-04
Accelerated TrajectoryReporter when trajectory output is disabled

falletta:
By avoid unnecessary state splitting operations when trajectories aren’t required, we avoid additional computational overhead (see profiling plot in this [PR](https://github.com/TorchSim/torch-sim/pull/427)).

### DIFF
--- a/torch_sim/trajectory.py
+++ b/torch_sim/trajectory.py
@@ -354,8 +354,17 @@ class TrajectoryReporter:
         Returns:
             list[dict[str, torch.Tensor]]: Property dictionaries, one per system.
         """
-        sizes = state.n_atoms_per_system.tolist()
         n_sys = state.n_systems
+        n_atoms = getattr(
+            state,
+            "n_atoms",
+            state.system_idx.shape[0],
+        )
+        sizes = getattr(
+            state,
+            "n_atoms_per_system",
+            torch.bincount(state.system_idx, minlength=n_sys),
+        ).tolist()
         all_props: list[dict[str, torch.Tensor]] = [{} for _ in range(n_sys)]
 
         for frequency, calculators in self.prop_calculators.items():
@@ -367,10 +376,12 @@ class TrajectoryReporter:
                     result = result.unsqueeze(0)
 
                 # infer per-atom vs per-system from shape
-                if result.shape[0] == state.n_atoms:
+                if result.shape[0] == n_atoms:
                     splits = torch.split(result, sizes)
+                elif result.shape[0] == n_sys:
+                    splits = [result[i : i + 1] for i in range(n_sys)]
                 else:
-                    splits = [result[i] for i in range(n_sys)]
+                    splits = [result.clone() for _ in range(n_sys)]
 
                 for idx in range(n_sys):
                     sys_step = step[idx] if isinstance(step, list) else step


torch_sim/runners.py
```diff
@@ -812,13 +813,27 @@ class StaticState(SimState):
...
+            for sys_idx in range(sub_state.n_systems):
+                atom_mask = sub_state.system_idx == sys_idx
+                base_props: dict[str, torch.Tensor] = {
+                    "potential_energy": static_state.energy[sys_idx],
+                }
+                if model.compute_forces:
+                    base_props["forces"] = static_state.forces[atom_mask]
+                if model.compute_stress:
+                    base_props["stress"] = static_state.stress[sys_idx]
+                all_props.append(base_props)
```
orionarcher: Are you sure this is faster?

The profiling data doesn't really say anything about how long the report call takes, and this isn't benchmarked against the current code.

torch_sim/runners.py
```diff
@@ -709,7 +709,7 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
     return state  # type: ignore[return-value]
 
 
-def static(
+def static(  # noqa: C901
```
orionarcher: I don't favor disabling the complexity limit here. A lot of effort has been put into keeping these functions minimal and readable.

torch_sim/runners.py
falletta: Yes, I confirm the speedup. Below are the results from running the `run_torchsim_static` function in `8.scaling.py` (see the [static PR](https://github.com/TorchSim/torch-sim/pull/427)) before and after the fix. We observe a speedup of up to 24.3%, and it continues to increase with system size.  In addition, I verified that the cost associated to the trajectory reporter disappears from the profiling analysis.

Previous results:
```
=== Static benchmark ===
  n=1 static_time=1.928943s
  n=1 static_time=0.272846s
  n=1 static_time=0.683335s
  n=1 static_time=0.026675s
  n=10 static_time=0.281990s
  n=100 static_time=0.705871s
  n=500 static_time=1.510273s
  n=1000 static_time=1.528872s
  n=2500 static_time=3.809000s
  n=5000 static_time=7.890238s
```

New results:
```
  n=1 static_time=2.165601s
  n=1 static_time=0.271961s
  n=1 static_time=0.665016s
  n=1 static_time=0.022899s
  n=10 static_time=0.295468s
  n=100 static_time=0.692651s
  n=500 static_time=1.411905s
  n=1000 static_time=1.291772s -> 18.3% speedup
  n=2500 static_time=3.175455s -> 19.9% speedup
  n=5000 static_time=6.348887s -> 24.3% speedup
```

torch_sim/runners.py
falletta: Ok, I introduced helper functions to reduce the complexity of `static`. However, note that `optimize` has a `# noqa: C901, PLR0915`, which disables the complexity limit, similarly to my original implementation. We might consider introducing helper functions in `optimize` as well, for consistency and style with `static`.

orionarcher: Thanks! looks much better

torch_sim/runners.py
orionarcher: I'm confused, is the 24% speedup here coming from just from removing the report call or all of the changes in the other PR too? I'm curious of the effect just the report call has, I just want to make sure that specific optimizations are well-founded, I'm not opposed to making this change.

orionarcher:
Upon further review, it seems the slowdown from calling `report` is coming from calling `state.split()` inside `report()`, which doesn't need to be called when there are no files to write. Rather than bypassing the reporter in `static` and duplicating property extraction in `_collect_base_properties`, what about adding a fast path inside report() itself?

When `self.filenames is None`, we can early-return by calling prop_calculators on the batched state and splitting the result tensors by n_atoms_per_system, avoiding N SimState constructions entirely. That way static doesn't need to change and we don't end up with two property-extraction code paths. Something like this (from CC):

```python
  def report(self, state, step, model=None):
      if self.filenames is None:
          return self._extract_props_batched(state, step, model)

      # ... existing split-based logic unchanged ...

  def _extract_props_batched(self, state, step, model):
      sizes = state.n_atoms_per_system.tolist()
      n_sys = state.n_systems
      all_props: list[dict[str, torch.Tensor]] = [{} for _ in range(n_sys)]

      for frequency, calculators in self.prop_calculators.items():
          if frequency == 0:
              continue
          for prop_name, prop_fn in calculators.items():
              result = prop_fn(state, model)
              if result.dim() == 0:
                  result = result.unsqueeze(0)

              # infer per-atom vs per-system from shape
              if result.shape[0] == state.n_atoms:
                  splits = torch.split(result, sizes)
              else:
                  splits = [result[i] for i in range(n_sys)]

              for idx in range(n_sys):
                  sys_step = step[idx] if isinstance(step, list) else step
                  if sys_step % frequency == 0:
                      all_props[idx][prop_name] = splits[idx]

      return all_props
```

torch_sim/runners.py
falletta: Yes, this optimization is actually coming solely from the trajectory reporter fix (surprising but true!). I recalculated the timings using your code suggestion, and here are the results from two consecutive runs of the `8.benchmarking.py` script (static only):

Previous results:
```
  n=1 static_time=2.158836s
  n=1 static_time=0.276653s
  n=1 static_time=0.663958s
  n=1 static_time=0.024062s
  n=10 static_time=0.291398s
  n=100 static_time=0.729588s
  n=500 static_time=1.541228s
  n=1000 static_time=1.596516s
  n=2500 static_time=3.966619s
  n=5000 static_time=8.203963s
```

New results:
```
  n=1 static_time=1.974184s
  n=1 static_time=0.275181s
  n=1 static_time=0.672710s
  n=1 static_time=0.023083s
  n=10 static_time=0.277349s
  n=100 static_time=0.691769s
  n=500 static_time=1.412826s
  n=1000 static_time=1.297880s
  n=2500 static_time=3.220407s
  n=5000 static_time=6.433095s
```

falletta:
Thanks @orionarcher for the review. I like your code suggestion, I implemented it and confirmed the performance gain is maintained. I also verified that the trajectory reporter operations don’t appear in the profiling output, as expected. Note that the profiling below does not account for the speedups in `_split_state`, which previously took a significant amount of compute and motivated [PR #427](https://github.com/TorchSim/torch-sim/pull/427).
Screenshot 2026-02-10 at 12 08 53 PM

falletta:
Great, I applied a few fixes, and now the tests are passing

PR #444, opened 2026-02-06
Vectorize atom index computation in L-BFGS optimizer

falletta:
Replace Python for-loops over systems with vectorized tensor operations in the L-BFGS optimizer's padding/unpadding helpers and mask construction.

### DIFF
--- a/torch_sim/optimizers/lbfgs.py
+++ b/torch_sim/optimizers/lbfgs.py
@@ -534,6 +534,11 @@ def lbfgs_step(  # noqa: PLR0915, C901
     if is_cell_state:
         cell_filters.compute_cell_forces(model_output, state)
 
+    # Update state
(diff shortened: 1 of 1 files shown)


abhijeetgangan:
Could you run https://github.com/TorchSim/torch-sim/pull/365#issuecomment-3848084056 and see the performance did not regress?

falletta:
Sure, I’ve done two consecutive runs of the script in #365 with and without these changes, and obtained:
```
Before:  L-BFGS       178.79       1000/1000          0.0001
After:   L-BFGS       165.76       1000/1000          0.0001
```

A second pair of consecutive runs gives:
```
Before:  L-BFGS       185.18       1000/1000          0.0001
After:   L-BFGS       171.04       1000/1000          0.0001
```

So it looks like, besides timing fluctuations, there might also be some minor optimization.

PR #446, opened 2026-02-07
Make GraphPES tests less flaky as failures are not related to ts code.

CompRhys:
Reduce GraphPES test issues. Also rename graphpes_framework to better reflect package nature. Adds file to be backwards compatible for the time being.


CompRhys:
@thomasloux all test failures are due to the issue you raised for fairchem

PR #447, opened 2026-02-09
Add cumulative strain guard to FixSymmetry

janosh:
## Summary

- Store initial refined cells as `reference_cells` in `from_state()` and track cumulative strain in `adjust_cell()`
- When total deformation exceeds `max_cumulative_strain` (default 0.5), clamp the cell update to stay within the strain envelope
- Clamp (instead of reject) per-step deformations > 0.25 so the optimizer can still make progress
- Propagate `reference_cells` through `reindex()`, `merge()`, `select_constraint()`, and `select_sub_constraint()`

Builds on #438 (FixSymmetry constraint). The per-step deformation check (`max_delta <= 0.25`) catches large single-step cell changes but not slow cumulative drift across many steps that can cause phase transitions (e.g. hexagonal → tetragonal cell collapse).

Complementary to #443 which fixes L-BFGS history contamination from unsymmetrized forces — that PR addresses force-direction drift, this one addresses cell-deformation drift.

## Test plan

- [x] Existing `test_large_deformation_clamped` updated: verifies per-step clamping and cumulative strain bound
- [x] New `test_cumulative_strain_clamp_direct`: 20 small anisotropic stretches that individually pass per-step check but cumulatively exceed the limit; verifies clamping and symmetry preservation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,30 +65,49 @@ jobs:
           - { python: '3.12', resolution: lowest-direct }
           - { python: '3.13', resolution: highest }
         model:
-          - { name: fairchem, test_path: "tests/models/test_fairchem.py" }
-          - { name: fairchem-legacy, test_path: "tests/models/test_fairchem_legacy.py" }
-          - { name: graphpes, test_path: "tests/models/test_graphpes.py" }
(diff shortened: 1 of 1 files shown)


orionarcher:
This looks like a nice change @janosh. IMO in general it'd be best to get a review on non-trivial changes before merging in.

PR #449, opened 2026-02-10
Fix for type mismatch bug and numerical stability improvements

shantanusharma:
This PR addresses a type mismatch bug in eigenvalue merging and improves numerical stability in Lennard-Jones force calculations.
    
    Key changes:
    - In `torch_sim/math.py`: Fixed a `TypeError` in `_determine_eigenvalue_case` where boolean masks were incorrectly combined with integer tensors. Added proper device handling for `torch.arange`.
    - In `tests/test_math.py`: Added comprehensive tests for `_determine_eigenvalue_case` and `_matrix_log_33` with numerically close eigenvalues to prevent regressions.
    - In `torch_sim/models/lennard_jones.py`: Implemented safe division in `unbatched_forward` to avoid division by zero when particles overlap (distances = 0).
    - In `torch_sim/transforms.py`: Added a check to `inverse_box` to raise a `ValueError` if the box contains zero values, preventing division by zero.
    
    Checklist:
    - [x] Doc strings have been added in the Google docstring format.
    - [x] Run ruff on your code.
    - [x] Tests have been added for any new functionality or bug fixes.

### DIFF
--- a/torch_sim/trajectory.py
+++ b/torch_sim/trajectory.py
@@ -290,12 +290,16 @@ class TrajectoryReporter:
         Raises:
             ValueError: If number of systems doesn't match number of trajectory files
         """
+        # Fast path: no files to write, avoid N SimState constructions from split()
+        if self.filenames is None:
+            return self._extract_props_batched(state, step, model)
+
         # Get unique system indices
         system_indices = range(state.n_systems)
         # system_indices = torch.unique(state.system_idx).cpu().tolist()
 
         # Ensure we have the right number of trajectories
-        if self.filenames is not None and len(system_indices) != len(self.trajectories):
+        if len(system_indices) != len(self.trajectories):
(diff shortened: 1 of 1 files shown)


CompRhys:
Could you split these into two PRs as they're not related changes. I think that setting the force to zero at zero distance makes sense so don't think we'll adopt that change. The convention I would expect to use is more like:

```python
def stable_lj(r, epsilon=1.0, sigma=1.0, alpha=1e-5):
    """
    Numerically stable LJ potential using a soft core
    """
    # Prevent division by zero and extreme values
    r_sq = r**2 + alpha * sigma**2
    
    # Lennard-Jones formula using modified r_sq
    sr6 = (sigma**2 / r_sq)**3
    sr12 = sr6**2
    
    return 4 * epsilon * (sr12 - sr6)
```

PR #450, opened 2026-02-10
Clamp Frechet log-space deformation gradient to prevent NaN from matrix_exp overflow

janosh:
## Problem

The Frechet cell filter accumulates cell deformation in log-space (`cell_positions`), recovering the actual cell via `matrix_exp(cell_positions / cell_factor)`. Over many LBFGS steps, `cell_positions` can grow unbounded for diverging structures, causing `matrix_exp` to overflow to Inf/NaN. This NaN then propagates to positions and cell, and on the next step `matrix_log_33` in `compute_cell_forces` crashes with a cusolver error, killing the entire batched optimization.

## Fix

Clamp the log-space deformation gradient to `[-2, 2]` (corresponding to ~7.4x max strain per axis) in two places:

1. **`lbfgs.py`**: After computing `deform_grad_log_new = cell_positions / cell_factor`, clamp before calling `matrix_exp`. Emits a warning with the number of affected systems. Writes clamped values back to `cell_positions` so they don't re-accumulate.

2. **`cell_filters.py`**: Same clamp in `compute_cell_forces` after `matrix_log_33`, before `expm_frechet`, to prevent NaN from propagating into the Frechet derivative computation.

## Test

`test_frechet_lbfgs_clamps_extreme_deformation`: Injects extreme `cell_positions` (log-deform = 10.0, well above the 2.0 clamp threshold) into an LBFGS state, runs one step, and verifies:
- Log-space deformation is bounded after the step (≤ 2.5)
- Positions and cell remain finite
- The clamp warning fires

The test **fails without the fix** (log-deform stays at 10.0) and **passes with it** (clamped to ≤ 2.0).

### DIFF
--- a/torch_sim/optimizers/lbfgs.py
+++ b/torch_sim/optimizers/lbfgs.py
@@ -11,15 +11,14 @@ The prev_forces and prev_positions are stored in the scaled/fractional space to 
 ASE's behavior exactly.
 """
 
-import warnings
 from typing import TYPE_CHECKING, Any
 
 import torch
 
 import torch_sim as ts
 from torch_sim.optimizers.cell_filters import (
-    MAX_LOG_DEFORM,
     CellLBFGSState,
+    _clamp_deform_grad_log,
     compute_cell_forces,
     deform_grad,
     frechet_cell_filter_init,
@@ -497,33 +496,9 @@ def lbfgs_step(  # noqa: PLR0915, C901
             # Frechet: deform_grad = exp(cell_positions / cell_factor)
             cell_factor_reshaped = state.cell_factor.view(n_systems, 1, 1)
             deform_grad_log_new = cell_positions_new / cell_factor_reshaped  # [S, 3, 3]
-
-            # Clamp log-space deformation to prevent matrix_exp overflow.
-            # Without this, cumulative LBFGS steps can push cell_positions to
-            # extreme values, causing matrix_exp to overflow to Inf/NaN.
-            # The NaN then propagates to the cell and positions, crashing
-            # matrix_log_33 in the next compute_cell_forces call and killing
-            # the entire batched optimization.
-            # exp(2) ≈ 7.4x strain per axis — structures deforming beyond
-            # this from their reference cell are diverging.
-            exceeds = deform_grad_log_new.abs() > MAX_LOG_DEFORM
-            if exceeds.any():
-                n_clamped = int(exceeds.any(dim=(-2, -1)).sum().item())
-                warnings.warn(
-                    f"Clamping log-space deformation gradient for {n_clamped} "
-                    f"system(s) to [-{MAX_LOG_DEFORM}, {MAX_LOG_DEFORM}] "
-                    f"(max |log(F)| = "
-                    f"{deform_grad_log_new.abs().max().item():.2f}). "
-                    f"This prevents matrix_exp overflow from diverging cell "
-                    f"optimization.",
-                    stacklevel=2,
-                )
-                deform_grad_log_new = deform_grad_log_new.clamp(
-                    -MAX_LOG_DEFORM, MAX_LOG_DEFORM
-                )
-                # Write back clamped values to cell_positions
-                cell_positions_new = deform_grad_log_new * cell_factor_reshaped
-
+            deform_grad_log_new, cell_positions_new = _clamp_deform_grad_log(
+                deform_grad_log_new, cell_positions_new, cell_factor_reshaped
+            )
             deform_grad_new = torch.matrix_exp(deform_grad_log_new)  # [S, 3, 3]
         else:
             # UnitCell: deform_grad = cell_positions / cell_factor
--- a/torch_sim/optimizers/bfgs.py
+++ b/torch_sim/optimizers/bfgs.py
@@ -20,7 +20,11 @@ import torch
 
 import torch_sim as ts
 from torch_sim.optimizers import cell_filters
-from torch_sim.optimizers.cell_filters import CellBFGSState, frechet_cell_filter_init
+from torch_sim.optimizers.cell_filters import (
+    CellBFGSState,
+    _clamp_deform_grad_log,
+    frechet_cell_filter_init,
+)
 from torch_sim.state import SimState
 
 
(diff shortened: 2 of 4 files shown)


abhijeetgangan:
This numerical isn't specific to lbfgs and happens with all other torchsim optimizers and ASE ones too. The cleaner approach is to have a better `cell_factor` at initialization or dynamically adjust than clamping. The best approach is to use a trust region which is not implemented yet. Running the test gives same results for all optimizers. Reducing the `cell_factor` to around 1 fixes it. 99% of the cases are mostly bad initialization or bad force fields.

```
import numpy as np
import torch
from ase.build import bulk
from ase.build import bulk as ase_bulk
from ase.calculators.lj import LennardJones as ASE_LJ
from ase.filters import FrechetCellFilter
from ase.optimize import LBFGS as ASE_LBFGS

import torch_sim as ts


DEVICE = torch.device("cpu")
DTYPE = torch.float64

ar_atoms = bulk("Ar", "fcc", a=5.26, cubic=True).repeat([2, 2, 2])
ar_state = ts.io.atoms_to_state(ar_atoms, DEVICE, DTYPE)

lj_model = LennardJonesModel(
    use_neighbor_list=True,
    sigma=3.405,
    epsilon=0.0104,
    device=DEVICE,
    dtype=DTYPE,
    compute_forces=True,
    compute_stress=True,
    cutoff=2.5 * 3.405,
)

identity = torch.eye(3, device=DEVICE, dtype=DTYPE).unsqueeze(0)

# Initial cell
initial_cell = ar_state.cell.squeeze(0)


def run_extreme_deformation_test(optimizer_name: str, init_fn, step_fn) -> dict:
    """Run the extreme deformation test for a given optimizer."""
    print(f"{optimizer_name}: Frechet extreme-deformation test")

    # Initialize optimizer with frechet cell filter
    state = init_fn(
        state=ar_state.clone(),
        model=lj_model,
        cell_filter=ts.CellFilter.frechet,
        # cell_factor=1, # Reduce the cell factor for more extreme deformation
    )

    initial_cell_positions = state.cell_positions.clone()
    # Inject extreme cell_positions: log-deform = cell_positions/cell_factor = 10
    state.cell_positions = (
        state.cell_positions + 10.0 * state.cell_factor.view(-1, 1, 1) * identity
    )

    state = step_fn(state=state, model=lj_model)

    has_nan_pos = torch.isnan(state.positions).any().item()
    has_inf_pos = torch.isinf(state.positions).any().item()
    has_nan_cell = torch.isnan(state.cell).any().item()
    has_inf_cell = torch.isinf(state.cell).any().item()

    # Print assertion values
    print(f"Positions NaN : {has_nan_pos}  (expect False)")
    print(f"Positions Inf : {has_inf_pos}  (expect False)")
    print(f"Cell NaN : {has_nan_cell}  (expect False)")
    print(f"Cell Inf : {has_inf_cell}  (expect False)")

    return {
        "cell": state.cell.squeeze(0),
        "initial_cell": initial_cell_positions,
        "initial_deformed_cell": state.cell_positions,
    }


# Run tests for both optimizers
lbfgs_result = run_extreme_deformation_test("LBFGS", ts.lbfgs_init, ts.lbfgs_step)
fire_result = run_extreme_deformation_test("FIRE", ts.fire_init, ts.fire_step)

# Summary
print("Initial cell:")
print(initial_cell.numpy())
print("LBFGS cell after:")
print(lbfgs_result["cell"].numpy())
print("FIRE cell after:")
print(fire_result["cell"].numpy())

print("Initial cell_positions:")
print(lbfgs_result["initial_cell"].numpy())
print("FIRE initial cell_positions:")
print(fire_result["initial_deformed_cell"].numpy())
print("LBFGS initial deformed cell:")
print(lbfgs_result["initial_deformed_cell"].numpy())

# ASE comparison: same extreme deformation test
print("ASE: Frechet extreme-deformation test")

# Create same Ar supercell with ASE
ase_atoms = ase_bulk("Ar", "fcc", a=5.26, cubic=True).repeat([2, 2, 2])
ase_atoms.calc = ASE_LJ(sigma=3.405, epsilon=0.0104, rc=2.5 * 3.405)

print(f"ASE initial cell\n:{ase_atoms.get_cell().array}")

# Wrap with FrechetCellFilter
filtered_atoms = FrechetCellFilter(ase_atoms)

reference_cell = ase_atoms.get_cell().array.copy()
extreme_cell = reference_cell * np.exp(10.0)
ase_atoms.set_cell(extreme_cell, scale_atoms=True)
print(f"ASE cell after extreme injection:\n{ase_atoms.get_cell().array}")

# Re-wrap with FrechetCellFilter after cell change
filtered_atoms = FrechetCellFilter(ase_atoms)

# Run one LBFGS step
ase_opt = ASE_LBFGS(filtered_atoms, logfile=None)
ase_opt.step()

print(f"ASE cell after LBFGS step:\n{ase_atoms.get_cell().array}")
```

torch_sim/optimizers/lbfgs.py
```diff
@@ -494,12 +494,41 @@ def lbfgs_step(  # noqa: PLR0915, C901
             # Frechet: deform_grad = exp(cell_positions / cell_factor)
             cell_factor_reshaped = state.cell_factor.view(n_systems, 1, 1)
             deform_grad_log_new = cell_positions_new / cell_factor_reshaped  # [S, 3, 3]
+
+            # Clamp log-space deformation to prevent matrix_exp overflow.
+            # Without this, cumulative LBFGS steps can push cell_positions to
```
abhijeetgangan: This is not because cumulative steps but rather the immediate step that causes it. Also, not specific to lbfgs and can happen with other too (Repro for more info). Maybe should be done for both positions and cell updates.

abhijeetgangan:
@CompRhys Let's keep this open for now. Either we handle the cell factor in the autobatching logic or implement https://en.wikipedia.org/wiki/Trust_region.

janosh:
i think @abhijeetgangan is right, trust regions sound like the more robust fix. not sure this was worth merging. not an urgent fix in any case

CompRhys:
I think the challenge is that implementing any other strategy breaks equivalence to ASE's implementation which is something that there are tests against. I don't think that there's a strong reason to stay aligned to ASE provided the physics makes sense and the performance doesn't shift in an unexpected way.

I was always skeptical as to why ASE saved the initial state for use as the reference in arbitrary length optimization but given that choice I think this clamping is fine to merge in so far as it's better than what was there yesterday even if it's not a final fix.

PR #456, opened 2026-02-17
Use first-party implementation of Orb Model Interface.

CompRhys:
This is a minimal wrapper designed to maintain the privileged position of Orb within the package as we move to adopt an external model posture. 

This should wait until a new orb release to be merged.


CompRhys:
Awaiting future release of orb-models (https://github.com/orbital-materials/orb-models/releases, https://pypi.org/project/orb-models/, need > v0.5.5)

CompRhys:
Tests fail locally due to a torchinductor compile error.

PR #457, opened 2026-02-18
Fix constraint atom_idx not remapped on reordered slice

falletta:
When slicing a batched SimState with a non-sequential system order (e.g. `state[[1, 0]]`), `select_constraint` correctly filters constraint atom indices by the kept-atom mask, but the resulting indices were not remapped to reflect the new atom ordering. This caused `FixAtoms` (and any `AtomConstraint`) to target the wrong atoms in the sliced state.

This PR adds a remap step in `_filter_attrs_by_index` that translates constraint `atom_idx` from dense-mask positions to their final reordered positions, so constraints remain physically consistent after any reordering slice.

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -799,10 +799,10 @@ def _filter_attrs_by_index(
     # Remap constraint atom_idx to account for reordering by atom_indices
     atom_remap = torch.empty(state.n_atoms, dtype=torch.long, device=state.device)
     atom_remap[atom_indices] = torch.arange(len(atom_indices), device=state.device)
-    dense_to_reordered = atom_remap[torch.where(atom_mask)[0]]
+    new_atom_idx = atom_remap[torch.where(atom_mask)[0]]
     for c in filtered_attrs["_constraints"]:
         if hasattr(c, "atom_idx") and isinstance(c.atom_idx, torch.Tensor):
-            c.atom_idx = dense_to_reordered[c.atom_idx]
+            c.atom_idx = new_atom_idx[c.atom_idx]
 
     # Build inverse map for system_idx remapping (old index -> new position)
     if len(system_indices) == 0:


tests/test_constraints.py
```diff
@@ -909,6 +909,33 @@ def test_atom_indexed_constraint_update_and_select() -> None:
...
 
 
+def test_constraint_atom_idx_remapped_on_reordered_slice(
+    mixed_double_sim_state: ts.SimState,
+) -> None:
+    """Slicing a batched state with reversed system order must remap constraint
+    atom_idx so the constraint targets the same physical atoms in the new ordering.
+    """
+    state = mixed_double_sim_state
+    n_sys0 = (state.system_idx == 0).sum().item()
```
curtischong: maybe state.n_atoms_per_system()[0].item() is more clear

tests/test_constraints.py
```diff
@@ -909,6 +909,33 @@ def test_atom_indexed_constraint_update_and_select() -> None:
...
+    state.constraints = [FixAtoms(atom_idx=torch.tensor([fix_global_idx]))]
+
+    # Reverse system order: system 1's atoms come first in the new ordering,
+    # so the fixed atom (second of old system 1) lands at new index 1.
+    sliced = state[[1, 0]]
+
+    constraint = sliced.constraints[0]
+    assert isinstance(constraint, FixAtoms)
+
+    assert constraint.atom_idx.item() == 1
```
curtischong: this line of code is slightly hard to follow because I need to look around and figure out what 1 represents. can we use like a variable to describe it? Maybe we also wouldn't need to use the (state.system_idx == 0).sum().item() above. maybe we can just use constants to say: "original_system1_idx = 1" or "system1_atom_idx=1"

Then we can go something like:

```
assert constraint.atom_idx.item() == system1_atom_idx
```

tests/test_constraints.py
```diff
@@ -909,6 +909,33 @@ def test_atom_indexed_constraint_update_and_select() -> None:
...
+
+    forces = torch.ones(sliced.n_atoms, 3, dtype=DTYPE)
+    constraint.adjust_forces(sliced, forces)
+    assert torch.all(forces[1] == 0.0)
```
curtischong: I think some comments would better explain why this asserts the correct behavior

tests/test_constraints.py
falletta: Sure!

tests/test_constraints.py
falletta: Ok I defined `expected_atom_idx = 1` which I then use for the asserts

tests/test_constraints.py
falletta: I think it's ok, this test is just a few lines of code so I'd prefer keeping it minimal

torch_sim/state.py
```diff
@@ -796,6 +796,14 @@ def _filter_attrs_by_index(
         if (c := con.select_constraint(atom_mask, system_mask))
     ]
 
+    # Remap constraint atom_idx to account for reordering by atom_indices
+    atom_remap = torch.empty(state.n_atoms, dtype=torch.long, device=state.device)
+    atom_remap[atom_indices] = torch.arange(len(atom_indices), device=state.device)
+    dense_to_reordered = atom_remap[torch.where(atom_mask)[0]]
```
curtischong: It's hard for me to understand what dense_to_reordered means. Maybe rename to new_atom_idx

curtischong (approved):
looks good! If we can get the final rename done as well that'd be great

torch_sim/state.py
falletta: Sure!

PR #460, opened 2026-02-21
Introduce PRNG to SimState and add reproducibility docs.

CompRhys:
The only messy bit is resumption as for serialization it seems that the only way to do it is with torch.save and I feel asking the user to store the pickle manually is awkward. 

--- 
### AI Overview

Every `SimState` now carries an optional `_rng` field (a `torch.Generator`) that controls all stochastic operations: momentum initialization, Langevin OU noise, V-Rescale Gamma draws, and C-Rescale barostat noise. No integrator init or step function accepts a `seed` or `prng` argument anymore — seeding is done exclusively through the state.

### The `rng` property

    state.rng = 42          # int → coerced to a seeded Generator
    state.rng = gen         # torch.Generator used directly
    state.rng = None        # reset; next access creates an unseeded Generator
    samples = state.rng     # lazily initialises if _rng is None/int, then returns it

- **Lazy**: if `_rng` is `None` (the default), accessing `state.rng` creates a new `torch.Generator` on the state's device and stores it back. No Generator is allocated until first use.
- **Coercing**: if `_rng` is an `int`, accessing `state.rng` converts it to a seeded Generator via `coerce_prng()` and stores it back, so subsequent accesses return the same (advancing) Generator.
- **Advancing**: because a single `torch.Generator` object is stored, its internal state advances with each draw, giving a proper random stream rather than re-seeding every step.

### Cloning

`state.clone()` deep-copies the Generator via `get_state()` / `set_state()`, producing an independent copy with identical initial RNG state. Drawing from one does not affect the other.

### Splitting

`state.split()` copies global attributes (including `_rng`) to every piece. All resulting single-system states share the same Generator **value** (copied), not the same object.

### Concatenating

`concatenate_states([s1, s2, ...])` takes global attributes from the **first** state. The resulting batch uses `s1`'s Generator; other states' Generators are discarded.

### Device movement

`state.to(device)` moves the Generator to the target device via `coerce_prng()`, which creates a new Generator on the target device and copies the RNG state if devices differ.

### Serialisation

    torch.save(state.rng.get_state(), "rng.pt")           # save
    gen = torch.Generator(device=state.device)
    gen.set_state(torch.load("rng.pt"))
    state.rng = gen                                        # restore

### What changed

- `_rng` moved from `MDState` to `SimState` (it's a global attribute, not MD-specific).
- All `seed=` / `prng=` parameters removed from integrator init functions.
- `initialize_momenta` takes `generator: torch.Generator | None` directly.
- V-Rescale Gamma sampling switched from `torch.distributions.Gamma` (unseeded) to `torch._standard_gamma(..., generator=rng)` so it's now fully seedable.
- `_rattle_sim_state` in `testing.py` refactored to use `state.rng` instead of saving/restoring global RNG state.
- `coerce_prng` handles cross-device Generator transfer.
- `_state_to_device` handles Generator device movement.

### DIFF
--- a/torch_sim/integrators/md.py
+++ b/torch_sim/integrators/md.py
@@ -1,10 +1,12 @@
 """Core molecular dynamics state and operations."""
 
+import warnings
 from collections.abc import Callable
 from dataclasses import dataclass
 
 import torch
 
+from torch_sim._duecredit import dcite
 from torch_sim.models.interface import ModelInterface
 from torch_sim.quantities import calc_kT
 from torch_sim.state import SimState
@@ -119,8 +121,8 @@ def initialize_momenta(
     device = positions.device
     dtype = positions.dtype
 
-    if isinstance(kT, torch.Tensor) and len(kT.shape) > 0:
-        # kT is a tensor with shape (n_systems,)
+    kT = torch.as_tensor(kT, device=device, dtype=dtype)
+    if kT.ndim > 0:
         kT = kT[system_idx]
 
     # Generate random momenta from normal distribution
@@ -165,6 +167,7 @@ def momentum_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
         MDState: Updated state with new momenta after force application
 
     """
+    dt = torch.as_tensor(dt, device=state.device, dtype=state.dtype)
     new_momenta = state.momenta + state.forces * dt
     state.set_constrained_momenta(new_momenta)
     return state
@@ -185,12 +188,15 @@ def position_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
         MDState: Updated state with new positions after propagation
 
     """
+    dt = torch.as_tensor(dt, device=state.device, dtype=state.dtype)
     new_positions = state.positions + state.velocities * dt
     state.set_constrained_positions(new_positions)
     return state
 
 
-def velocity_verlet[T: MDState](state: T, dt: torch.Tensor, model: ModelInterface) -> T:
+def velocity_verlet_step[T: MDState](
+    state: T, dt: float | torch.Tensor, model: ModelInterface
+) -> T:
     """Perform one complete velocity Verlet integration step.
 
     This function implements the velocity Verlet algorithm, which provides
@@ -214,6 +220,7 @@ def velocity_verlet[T: MDState](state: T, dt: torch.Tensor, model: ModelInterfac
         - Conserves energy in the absence of numerical errors
         - Handles periodic boundary conditions if enabled in state
     """
+    dt = torch.as_tensor(dt, device=state.device, dtype=state.dtype)
     dt_2 = dt / 2
     state = momentum_step(state, dt_2)
     state = position_step(state, dt)
@@ -223,6 +230,18 @@ def velocity_verlet[T: MDState](state: T, dt: torch.Tensor, model: ModelInterfac
     state.energy = model_output["energy"]
     state.forces = model_output["forces"]
     return momentum_step(state, dt_2)
+
+
+def velocity_verlet[T: MDState](
+    state: T, dt: float | torch.Tensor, model: ModelInterface
+) -> T:
+    """Deprecated alias for velocity_verlet_step."""
+    warnings.warn(
+        "velocity_verlet is deprecated. Use velocity_verlet_step instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return velocity_verlet_step(state=state, dt=dt, model=model)
 
 
 @dataclass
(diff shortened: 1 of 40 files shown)


torch_sim/integrators/nvt.py
```diff
@@ -67,7 +67,13 @@ def _ou_step(
     c2 = torch.sqrt(kT * (1 - torch.square(c1))).unsqueeze(-1)
 
     # Generate random noise from normal distribution
-    noise = torch.randn_like(state.momenta, device=state.device, dtype=state.dtype)
```
CompRhys, line 71: after https://github.com/pytorch/pytorch/issues/165865 randn_like is in torch 2.10 but I am not sure we want to pin to 2.10 given not all the models people want to use will support.

CompRhys: https://docs.pytorch.org/docs/2.9/generated/torch.randn_like.html vs https://docs.pytorch.org/docs/2.10/generated/torch.randn_like.html

torch_sim/testing.py
```diff
@@ -160,19 +160,40 @@ def make_sio2_sim_state(
...
+    Uses the state's ``rng`` (seeded with *seed*) so no global RNG state is touched.
+    """
     sim_state = sim_state.clone()
-    rng_state = torch.random.get_rng_state()
-    try:
-        torch.manual_seed(seed)
-        weibull = torch.distributions.weibull.Weibull(scale=0.1, concentration=1)
-        rnd = torch.randn_like(sim_state.positions)
-        rnd = rnd / torch.norm(rnd, dim=-1, keepdim=True)
-        shifts = weibull.sample(rnd.shape).to(device=sim_state.positions.device) * rnd
```
CompRhys, line 172: avoid `weibull.sample()` as it cannot be seeded.

torch_sim/integrators/nvt.py
```diff
@@ -531,9 +541,10 @@ def _vrescale_update(
     KE_new = dof * kT_tensor / 2
 
     # Generate random numbers
-    r1 = torch.randn(n_systems, device=device, dtype=dtype)
-    # Sample Gamma((dof - 1)/2, 1/2) = \sum_2^{dof} X_i^2 where X_i ~ N(0,1)
-    r2 = torch.distributions.Gamma((dof - 1) / 2, torch.ones_like(dof) / 2).sample()
```
CompRhys, line 540: avoid `Gamma.sample()` as it cannot be seeded.

torch_sim/state.py
```diff
@@ -371,6 +418,17 @@ def get_number_of_degrees_of_freedom(self) -> torch.Tensor:
             raise ValueError("Degrees of freedom cannot be zero or negative")
         return dof_per_system
 
+    @staticmethod
+    def _clone_attr(value: object) -> object:
+        """Clone a single attribute value, handling torch.Generator specially."""
```
CompRhys, line 423: forks have identical rng states.

torch_sim/integrators/md.py
```diff
@@ -89,12 +93,12 @@ def calc_kT(self) -> torch.Tensor:  # noqa: N802
         )
 
 
-def calculate_momenta(
+def initialize_momenta(
```
CompRhys, line 98: driveby: this was a misleading name.

abhijeetgangan:
LGTM. Note that setting deterministic mode can have performance penalties. Also, use of text vs binary file formats for restarting can have an effect.

torch_sim/integrators/npt.py
```diff
@@ -415,7 +430,10 @@ def _npt_langevin_velocity_step(
     c_2 = dt_atoms.unsqueeze(-1) * ((a * forces) + state.forces) / M_2.unsqueeze(-1)
 
     # Generate atom-specific noise
-    noise = torch.randn_like(state.momenta)
+    rng = state.rng
```
thomasloux, line 433: Is there a particular reason to not directly use state.rng and define a rng var ?

torch_sim/state.py
```diff
@@ -104,14 +133,32 @@ def charge(self) -> torch.Tensor: ...  # noqa: D102
...
+        ``torch.Generator`` is created on the same device as the state.
+        Assign ``None`` to reset, or assign a seeded ``torch.Generator``
+        or an ``int`` seed for reproducibility.
+        """
+        self._rng = coerce_prng(self._rng, self.device)
+        return self._rng
+
+    @rng.setter
+    def rng(self, value: int | torch.Generator | None) -> None:
+        self._rng = value
```
thomasloux, line 161: you probably want to coerce_prng only when setting the rng with a int value. Otherwise you're creating the object many time per MD step

thomasloux: If it's a way to handle `_rng=None`, you may want to do that in `post_init`

torch_sim/state.py
thomasloux: This should be fine, but I'm wondering whether there could be setup where one wants to clone state but still would need different prng (some kind of replica exchange)
Ok I'm thinking while writing, and it would be fine because if you batch your systems each system will get different random numbers.

torch_sim/state.py
CompRhys: I wanted to leave `_rng=None` alone when not used because then if you did clone an unseeded state all the downstream simstates would diverge

CompRhys:
```python
    if isinstance(rng, torch.Generator):
        if rng.device == device:
            return rng
        new = torch.Generator(device=device)
        new.set_state(rng.get_state())
        return new
```

in coerce if it's already a generator and on the right device we don't make a new object. It does add the two ifs into the step cost but I can't imagine that is significant?

torch_sim/integrators/npt.py
CompRhys: it only calls coerce once this way

Issue #463, opened 2026-02-23
Add MACE-POLAR input/output support across torch_sim APIs


janosh:
`torch_sim` already supports many aspects of MACE-POLAR (`charge`/`spin` plumbing exists), but we do not yet support all POLAR inputs and predicted properties in existing `torch_sim` APIs. This issue proposes a focused integration plan.

References:
- https://github.com/ACEsuit/mace-foundations/releases/tag/mace_polar_1
- https://mace-docs.readthedocs.io/en/latest/guide/polar_mace.html

### 1) Add external field to state + IO
- Extend `SimState` with per-system `external_field` (shape `[n_systems, 3]`)
- Include it in system-level attributes so split/slice/concat preserve it
- Validate shape in `SimState.__post_init__`
- Round-trip through ASE conversion:
  - `atoms_to_state`: read `atoms.info["external_field"]` (default zero vector)
  - `state_to_atoms`: write `atoms.info["external_field"]`

### 2) Pass POLAR inputs through `MaceModel`
- In `torch_sim.models.mace.MaceModel.forward`, pass:
  - `total_charge`
  - `total_spin`
  - `external_field` (included by default)

### 3) Standardize + preserve model outputs
- Keep canonical outputs stable:
  - `energy`, `forces`, `stress`
- Also propagate additional tensor outputs from MACE/POLAR in `MaceModel` results
  - e.g. `dipole`, `density_coefficients`, `spin_charge_density`, `charges`
- Goal: avoid hardcoding only a fixed small set of properties and support future POLAR heads without wrapper churn

### 4) Expose extra outputs in high-level APIs
- Ensure `ts.static(...)` returns extra model outputs in the per-system property dicts (in addition to existing core properties)
- Preserve correct per-system/per-atom splitting under batching/autobatching
- Keep trajectory reporting compatibility for these extra arrays

### 5) POLAR model discovery / convenience
- Add POLAR foundation aliases/checkpoint entries to `MaceUrls` (or equivalent in `mace.py`)
- Keep POLAR support in `MaceModel` rather than creating a separate wrapper class

### 6) Optional dependency docs/packaging
- Document and wire optional dependency required by POLAR runtime (`graph_electrostatics` / `graph_longrange`) in project extras/docs

### 7) Tests
- Add/extend tests for:
  - `external_field` IO round-trip
  - MACE-POLAR forward pass with charge/spin/field
  - presence + shape checks for electrostatic outputs
  - `ts.static` exposure of extra outputs
[...]

Issue #19, opened 2025-03-05
Write conversion format between HDF5 and popular formats


orionarcher:
Alternatively we could replace proprietary hdf5 format with https://www.nongnu.org/h5md/ and then leave ecosystem management to them.

PR #464, opened 2026-02-23
Introduce more extensible scheme for extra properties.

CompRhys:
This is a rough draft to try unlock #463

### DIFF
--- a/tests/models/test_fairchem.py
+++ b/tests/models/test_fairchem.py
@@ -1,7 +1,9 @@
 import traceback
+from collections.abc import Callable
 
 import pytest
 import torch
+from ase.build import bulk, fcc100, molecule
 
 import torch_sim as ts
 from tests.conftest import DEVICE, DTYPE
@@ -9,15 +11,11 @@ from tests.models.conftest import make_validate_model_outputs_test
 
 
 try:
-    from collections.abc import Callable
-
-    from ase.build import bulk, fcc100, molecule
     from fairchem.core.calculate.pretrained_mlip import (
         pretrained_checkpoint_path_from_name,
     )
     from huggingface_hub.utils._auth import get_token
 
-    import torch_sim as ts
     from torch_sim.models.fairchem import FairChemModel
 
 except (ImportError, OSError, RuntimeError, AttributeError, ValueError):
@@ -267,7 +265,7 @@ def test_fairchem_charge_spin(charge: float, spin: float) -> None:
         [mol],
         device=DEVICE,
         dtype=DTYPE,
-        system_extras={"charge": "charge", "spin": "spin"},
+        system_extras_map={"charge": "charge", "spin": "spin"},
     )
     assert state.charge is not None
     assert state.spin is not None


torch_sim/state.py
```diff
@@ -92,6 +93,8 @@ class SimState:
     spin: torch.Tensor | None = field(default=None)
     system_idx: torch.Tensor | None = field(default=None)
     _constraints: list["Constraint"] = field(default_factory=lambda: [])  # noqa: PIE807
+    _system_extras: dict[str, torch.Tensor] = field(default_factory=dict)
```
CompRhys, line 184: could consider removing privileged positions of `spin` and `charge`

torch_sim/runners.py
```diff
@@ -821,8 +821,25 @@ class StaticState(SimState):
                 else torch.full_like(sub_state.cell, fill_value=float("nan"))
             ),
         )
+        static_state.store_model_extras(model_outputs)
 
         props = trajectory_reporter.report(static_state, 0, model=model)
+
```
thomasloux, line 841: I would probably be more in favor of getting rid of trajectory_reporter here and just concatenate the results returned. We should just copy the logic of detecting whether the property is per atom or per system.

Trajectory_reporter have more functionalities than that but I'm not sure many people use it when running this static() function

thomasloux:
Nice work!

I think that most of the time extra predicted properties are not useful in MD, so that `_system_extras` and `_atom_extras` are not useful most of the time, but why not for analysis purposes. It's mainly the extra inputs that are important. More specifically, as torch_sim is transitioning to externally defined Model, they should be a very clear and canonical way to add extra_input in State that can be used easily but the models.

CompRhys:
> More specifically, as torch_sim is transitioning to externally defined Model, they should be a very clear and canonical way to add extra_input in State that can be used easily but the models.

I think here the aim was to just get something that would be general enough to be both input and output similar to `info` on ASE `atoms`. The prompting here was primarily focussed on hitting the generality point whilst trying to have a reasonable interface. The main questionable thing about this approach is that accessing the "extras" via the property syntax `simstate.external_E_field` isn't currently typed. 

Is this sufficient for the matris use-case? do you want to try explore that on top of this and then we can revisit this design? 

@Asecretboy would you be interested in adding a torchsim interface to https://github.com/HPC-AI-Team/MatRIS as an alternative to https://github.com/HPC-AI-Team/MatRIS/blob/c16f569ca08e6905e91b64e2ee68614303e46f7f/matris/applications/base.py#L24. TorchSim is an engine for batched MLIP workflows that enables fast MD with replicas/repeats or batched geometry optimization with minimal conceptual overhead c.f. ASE.

curtischong:
Just putting this here for documentation purposes - but this is similar to another PR that I wrote and closed earlier: https://github.com/TorchSim/torch-sim/pull/354

thomasloux:
I added a torch-sim interface https://github.com/HPC-AI-Team/MatRIS/pull/5
I didn't have time to test your PR, but it seems fine.

torch_sim/runners.py
orionarcher: That is currently the default behavior but it's a bit convoluted. If no reporter is given:
1. `static` will make one when `_configure_reporter` is called.
2. When calling report the file IO will be bypassed in favor of a concatenation as you describe.

`TrajectoryReporter` probably needs a rework. Both to make it compatible with `h5md` (#19) and to make it easier to report out model properties, as you've identified.

For the latter, the simplest solution (perhaps not the best) would be to add the `model_outputs` as a optional kwarg to the report function so that model predictions can be covered by the current API without needing an additional call.

torch_sim/state.py
orionarcher: It would honestly be a good test. If we can move `spin` and `charge` to extras without breaking anything that would help to validate this PR.

torch_sim/state.py
```diff
@@ -1116,6 +1281,14 @@ def concatenate_states[T: SimState](  # noqa: C901, PLR0915
     # Concatenate system indices
     concatenated["system_idx"] = torch.cat(new_system_indices)
 
+    # Concatenate extras
+    concatenated["_system_extras"] = {
+        key: torch.cat(tensors, dim=0) for key, tensors in system_extras_tensors.items()
+    }
```
orionarcher, line 1436: This should be a stack operation I believe? If we move charge & spin to extras it will give us a already extant test.

torch_sim/io.py
```diff
@@ -279,6 +293,24 @@ def atoms_to_state(
...
+                )
+
+    _atom_extras: dict[str, torch.Tensor] = {}
+    if atom_extras_keys:
+        for key in atom_extras_keys:
+            arrays = [at.arrays.get(key) for at in atoms_list]
+            if all(a is not None for a in arrays):
+                _atom_extras[key] = torch.tensor(
+                    np.concatenate(arrays), dtype=dtype, device=device
+                )
```
orionarcher, line 344: This makes me a bit nervous, there is no guarantee that ASE model interfaces will need the same shape/structure as TorchSim model interfaces. Feels like it opens up an easy place for bugs to slide into the interface. This could be a reason to leave `charge` and `spin` in their privileged positions. 

(confidence 50% on this opinion)

torch_sim/state.py
```diff
@@ -183,16 +189,132 @@ def __post_init__(self) -> None:  # noqa: C901
...
+    def atom_extras(self) -> dict[str, torch.Tensor]:
+        """Get the atom extras."""
+        return self._atom_extras
+
+    def set_extras(
+        self,
+        key: str,
+        value: torch.Tensor,
+        scope: Literal["per-system", "per-atom"],
+    ) -> None:
```
orionarcher: this doesn't appear to be used anywhere?

torch_sim/state.py
```diff
@@ -183,16 +189,132 @@ def __post_init__(self) -> None:  # noqa: C901
...
+                    f"Atom extras {key} leading dim must be "
+                    f"n_atoms={self.n_atoms}, got {value.shape[0]}"
+                )
+            self._atom_extras[key] = value
+        else:
+            raise ValueError(f"scope must be 'per-system' or 'per-atom', got {scope!r}")
+
+    def has_extras(self, key: str) -> bool:
+        """Check if an extras key exists."""
+        return key in self._system_extras or key in self._atom_extras
```
orionarcher, line 374: also doesn't appear to be used

orionarcher (commented):
Broadly this seems like quite a reasonable interface. A few comments.

One question: how would we handle fluctuating extras? e.g. a user wants to applying a fluctuating E field to a system, it occurs to me we don't handle that but perhaps it's an advanced use case beyond the support of the default runners.

abhijeetgangan:
>
> One question: how would we handle fluctuating extras? e.g. a user wants to applying a fluctuating E field to a system, it occurs to me we don't handle that but perhaps it's an advanced use case beyond the support of the default runners.

This could be like temperature which the runners and there by the step functions can take as inputs at each step. The only difference is that the model also has to take them as inputs in this case not just the step functions. I think the latter is not dealt with right now? This is assuming something like the E field doesn't change the step function.

Asecretboy:
> > More specifically, as torch_sim is transitioning to externally defined Model, they should be a very clear and canonical way to add extra_input in State that can be used easily but the models.
> 
> I think here the aim was to just get something that would be general enough to be both input and output similar to `info` on ASE `atoms`. The prompting here was primarily focussed on hitting the generality point whilst trying to have a reasonable interface. The main questionable thing about this approach is that accessing the "extras" via the property syntax `simstate.external_E_field` isn't currently typed.
> 
> Is this sufficient for the matris use-case? do you want to try explore that on top of this and then we can revisit this design?
> 
> @Asecretboy would you be interested in adding a torchsim interface to https://github.com/HPC-AI-Team/MatRIS as an alternative to https://github.com/HPC-AI-Team/MatRIS/blob/c16f569ca08e6905e91b64e2ee68614303e46f7f/matris/applications/base.py#L24. TorchSim is an engine for batched MLIP workflows that enables fast MD with replicas/repeats or batched geometry optimization with minimal conceptual overhead c.f. ASE.

Sorry for the delayed reply. We've been updating the MatRIS codebase and training some new model variants recently. Once that's done, I'll try to add a torchsim interface.

torch_sim/state.py
CompRhys: stack creates a new dimension which they requires the sizes to match

orionarcher:
@CompRhys what are steps from here to merge in your opinion?

Need any help or additional review?

CompRhys:
was looking at removing the spin/charge directly but nothing felt clean enough to want to do with it. The challenge is that this sort of generality is in some way opposite to type checking etc but need to hit a good balance. I would appreciate someone else taking over and maybe trying to get closer to a clean solution.

ultimately a test of whether this is good enough abstraction/clean interface is to remove spin and charge as direct kwargs but have all the tests of spin and charge still pass unchanged.

cw-tan:
Hi, just wondering what the state of this PR is. For some context, we have models like `allegro-pol` that predict extra outputs. We have a torchsim integration (https://github.com/mir-group/allegro-pol/blob/main/allegro_pol/integrations/torchsim.py), but I'm not sure what the right way is to dump the extra stuff during MD or other workflows. 

I just added a temp `save_extra_outputs` hook for now, but would like to migrate to a canonical torchsim framework for handling arbitrary extra inputs and outputs so that downstream clients can extend it to models that do more than just energies, forces and stresses. Thanks.

CompRhys:
I think this is probably 70% of the way there but I am not confident enough to push ahead with it without some more iteration. I am not sure I can put aside too much time in the next month to get to this given current project load. Very happy to taken criticisms and suggestions on how this is implemented or what usage patterns might be important downstream to try ensure we don't miss certain usage patterns. This feels like a fairly big change so important to get it right I believe.

cw-tan:
@CompRhys if it's helpful, our group could try using the `allegro-pol` torchsim integration as a trivial use case for this new extensible schema. Some people are also interested in using torchsim with NequIP-LES (https://github.com/ChengUCB/NequIP-LES/issues/2), so the extraction of Born effective charges in that scheme could also be another trivial test case to build confidence in the extension.

CompRhys:
@falletta is going to try take a look at fixing the remaining errors whilst keeping this backwards compatible. As in the comments above the go/no-go here is can we remove the `spin` and `charge` attributes as being privilaged in a backwards compatible manner with minimal test changes. If we can do that then I think this is clean enough to move forward.

I do recognize that there is pent up demand here (which is good) and the offer to help alpha test would be great. The metatomic/PET-MAD team at EPFL are also interested in this for returning uncertainties and MACE-POLAR demands the E_field as an input which was what triggered me trying to do this.

falletta:
ok guys, this [PR](https://github.com/TorchSim/torch-sim/pull/526) should fix the remaining issues on the core tests

torch_sim/runners.py
CompRhys: #536 can track this. I would need to go into more detail to understand what it going on here. I would like to merge this as a v0 to give a bit of certainty to people using downstream even if it's not yet fixed in stone.

CompRhys:
https://github.com/TorchSim/torch-sim/pull/532 provides the first useful example of this interface being used for partial charges and electrostatics. I think this is sufficient to merge as a v0 state and painpoints can be addressed as they arise.

torch_sim/state.py
```diff
@@ -67,6 +72,27 @@ def require_system_idx(system_idx: torch.Tensor | None) -> torch.Tensor:
     return system_idx
 
 
+def _wrap_init_for_extras(cls: type) -> None:
+    """Wrap a dataclass __init__ to route unknown kwargs into _system_extras."""
```
orionarcher: should we check shape before routing to atom or system extras?

torch_sim/state.py
```diff
@@ -246,16 +277,102 @@ def __post_init__(self) -> None:  # noqa: C901
...
+        * ``n_atoms``  → stored in ``_atom_extras``
+        * ``n_systems`` → stored in ``_system_extras``
+        * otherwise     → skipped (ambiguity or scalar)
+
+        When ``n_atoms == n_systems`` (single-atom system), the tensor is stored as
+        per-atom by convention.
+
+        Args:
+            model_output: Full dict returned by ``model.forward()``.
+        """
```
orionarcher, line 391: this is nice, this fixes many of the issues with reporting out non-canonical model properties

torch_sim/typing.py
```diff
@@ -25,14 +54,16 @@ class BravaisType(StrEnum):
...
+    CUBIC = "cubic"
+    HEXAGONAL = "hexagonal"
+    TRIGONAL = "trigonal"
+    TETRAGONAL = "tetragonal"
+    ORTHORHOMBIC = "orthorhombic"
+    MONOCLINIC = "monoclinic"
+    TRICLINIC = "triclinic"
+
 
+ExtrasMap = dict[str, str]
```
orionarcher: kinda feels like overkill because it requires an extra lookup to know the type, less readable imho

torch_sim/typing.py
```diff
@@ -14,6 +14,35 @@
...
+class SystemExtras(StrEnum):
+    """Blessed names for per-system :class:`~torch_sim.state.SimState` extras.
+
+    Stored in ``SimState._system_extras``; leading dimension is ``n_systems``.
+    """
+
+    CHARGE = "charge"  # TOTAL_CHARGE preferred for less ambiguity with partial charges
+    SPIN = "spin"  # TOTAL_SPIN preferred
+    TOTAL_CHARGE = "total_charge"
+    TOTAL_SPIN = "total_spin"
```
orionarcher, line 37: should there be a partial_charges quantity here too? As mentioned in the electrostatics PR?

orionarcher:
Will this close #463 or will additional work be needed in the `MaceInterface` upstream?

torch_sim/typing.py
CompRhys: Line 23. Partial charges are per atom

CompRhys:
> Will this close #463 or will additional work be needed in the `MaceInterface` upstream?

My plan was to close it because there is the upstream mace interface now (pending next pypi release) and so we wouldn't be implementing it and this creates the framework to do so. We already have the allegro-pol torchsim interface that we've tested a bit and seems to work so the building blocks are now there.

torch_sim/typing.py
CompRhys: will type as dict[enum, str]

torch_sim/state.py
CompRhys: Will add

PR #467, opened 2026-02-25
Use first-party implementation of SevenNet Model Interface

CompRhys:
Reduce maintenance surface area for SevenNet.


CompRhys:
@YutackPark FYI will merge when you push your next release and I can set a real pin :) (https://github.com/MDIL-SNU/SevenNet/releases)

YutackPark:
SevenNet is just released with TorchSim interface and you can safely updated the requirements to `>=0.12.1`. I tried to push here but hook installation never ends...

CompRhys:
thanks @YutackPark!

Issue #386, opened 2025-12-12
Portable batched neighbourt list


venkatkapil24:
Hi all, bringing to your attention `batch_nl`, a portable neighbour list that can perform a single-op calculation on a batch of structures. While it is *not* as fast as NVIDIA's neighbourlist, it can be a good fallback option for AMD, TPU, or Intel devices.  You can expect an order-of-magnitude difference over the current default.  

https://github.com/venkatkapil24/batch_nl

<img width="677" height="343" alt="Image" src="https://github.com/user-attachments/assets/edd388cb-4b7e-4299-8d67-56e4cb0f330c" />

abhijeetgangan:
With #348 merged in it would make sense to remove the default torch neighborlist with this. Will open a PR for integrating this.

PR #470, opened 2026-02-27
Eliminate for-loop in build_naive_neighborhood

CompRhys:
This PR is in response to https://github.com/TorchSim/torch-sim/issues/386, whilst it stops short of importing that package (not on pypi) it draws inspiration and uses pad_sequence to avoid a for loop in `build_naive_neighborhood` offering speed for what's expected to be a fairly minor memory cost. Ideally we would be able to eliminate the for-loops in `build_linked_cell_neighborhood` but not apparent how to do that.

This PR won't change much as the default fallback is to use `torch_nl_linked_cell` which uses unvectorized `build_linked_cell_neighborhood` rather than `torch_nl_n2` which is now vectorized fully.

Unsure if `torch_sim/neighbors/standard.py` should be cleaned up by this PR as it mostly overlaps with functions in `torch_nl.py` and `transforms.py`.


tests/test_transforms.py
```diff
@@ -1279,6 +1237,221 @@ def test_linked_cell_basic() -> None:
...
+        atoms, cutoff, self_interaction=self_interaction
+    )
+    dist_ref_sorted = np.sort(dist_ref)
+    dists_sorted = np.sort(dists)
+
+    # Our pairs must be a superset of ASE's
+    assert len(dists_sorted) >= len(dist_ref_sorted), (
+        f"pbc={pbc_val}: got {len(dists_sorted)} pairs, "
+        f"expected at least {len(dist_ref_sorted)}"
+    )
```
CompRhys: AI test - is this discriminative enough?

abhijeetgangan:
Like this change. `torch_sim/neighbors/standard.py`  can be cleaned up.

PR #473, opened 2026-02-28
Nudged Elastic Band

### LINKED ISSUE #90: Nudged elastic band implementation
Nudged elastic band would be an excellent addition to the project.

CompRhys:
This attempts to revive @mstapelberg's NEB work which was documented in https://github.com/TorchSim/torch-sim/issues/90. This is LLM assisted migration of the older code which does not yet work.


mstapelberg:
Hi @CompRhys happy to work on this in 2 weeks! Defending my thesis next Friday so slammed with that but will have time to work on this and a few other things if interested after the fact.

CompRhys:
Best of luck! the @ was just to try ensure you still got credit for the work that you did here on neb

CompRhys:
image
image
This matches ASE, matching ASE treats all images as a single opt batch. Doing each image with it's own fire state led to ~3x the number of opt steps being needed.

This design is really messy and shouldn't be merged. Results shown from the two examples. Those should be cleaned up also.

CompRhys:
image

I am unsure why at the end torchsim in this tutorial example stops being an envelope function to the ASE results.

CompRhys:
To get this to work a concept of `group_idx` needs to be added to the simstate and then passed to all the optimizers. This is because the whole path is optimized at once.

thomasloux:
Hey, I'm not super familiar with NEB. For instance, `group_idx` seems necessary but adds a lot of complexity. Could you add a short doc explaining this design choice, in this case probably by explaining why/how NEB needs to optimize the entire path directly. This overall looks great. But I would need more time to have a thorough read

torch_sim/autobatching.py
```diff
@@ -386,6 +386,28 @@ def calculate_memory_scalers(
...
+    """Split a state into independent optimizer groups for autobatching."""
+    if state.n_groups == state.n_systems and torch.equal(
+        state.group_idx,
+        torch.arange(state.n_systems, device=state.device, dtype=torch.int64),
+    ):
+        return state.split()
+    return [
+        state[torch.where(state.group_idx == group_idx)[0]]
+        for group_idx in range(state.n_groups)
+    ]
```
orionarcher: This will incur overhead and throttles the batch size by the group size.

orionarcher (changes requested):
I agree that we do need a group_idx but I don't think this is the right approach to operationalizing it.

As I understand it, if we split the state before autobatching then we limit the batch size to the group size, which will throttle performance in the case of small groups. 

Instead, I think we need to modify the logic of the autobatchers so that they can either split by system_idx or split by group_idx. Then we can re-use the same logic to pack the maximum number of groups into memory that we use to pack the maximum number of systems into memory. This will require a deeper rewrite to the autobatchers though.

I suggest having a autobatcher property on instantiation that selects "group_idx" or "system_idx" as the batching index. Or could be a bool `batch_by_group=False`.

I spent a fair amount of time thinking about this last year and I think everything should transition over nicely, no blockers on an architectural level.

CompRhys:
> I suggest having a autobatcher property on instantiation that selects "group_idx" or "system_idx" as the batching index. Or could be a bool `batch_by_group=False`.

I was trying to get it such that group_idx was the thing that autobatchers worked off and that the default behavior would be single system groups. I think that's cleaner than having switching logic?

PR #474, opened 2026-02-28
Add batch LJ 

abhijeetgangan:
## Summary

* Add proper batched version of lennard jones.

Number from RTX4070M for 32 atoms argon system.

image

## Checklist

### DIFF
--- a/examples/tutorials/reporting_tutorial.py
+++ b/examples/tutorials/reporting_tutorial.py
@@ -226,7 +226,7 @@ reporter = ts.TrajectoryReporter(
 )
 
 # Initialize a model for energy calculation
-lj_model = LennardJonesModel()
+lj_model = LennardJonesModel(device=state.device, dtype=state.positions.dtype)
 
 # Run simulation with property calculation
 for step in range(100):


abhijeetgangan:
I kept the unbatched forward because it's a basic test case if the batching works properly in the first place.

torch_sim/models/lennard_jones.py
```diff
@@ -138,7 +140,8 @@ class LennardJonesModel(ModelInterface):
         compute_stress (bool): Whether to compute stress tensor.
         per_atom_energies (bool): Whether to compute per-atom energy decomposition.
         per_atom_stresses (bool): Whether to compute per-atom stress decomposition.
-        use_neighbor_list (bool): Whether to use neighbor list optimization.
+        disable_neighbor_list (bool): Whether to disable neighbor-list acceleration.
```
CompRhys: would it be cleaner to have an all-to-all neighborlist fn and delete this arg?

abhijeetgangan: Yes, I could do that. With cheap potentials the neighborlist is the dominant cost so no point in doing brute force. Will apply to other classical force fields also.

torch_sim/models/lennard_jones.py
CompRhys: I would favor not keeping unbatched around long term but if there's short term value I would prefer this be called `UnbatchedLennardJonesModel` as my understanding is that we want the ts base case to be being batched?

abhijeetgangan: We can have one unbatched model just for sanity checks but yeah can go ahead and changing the naming schemes.

torch_sim/models/lennard_jones.py
```diff
@@ -441,3 +420,132 @@ def forward(
...
+        wrapped_positions = (
+            ts.transforms.pbc_wrap_batched(positions, sim_state.cell, system_idx, pbc)
+            if pbc.any()
+            else positions
+        )
+
+        if pbc.ndim == 1:
+            pbc_batched = pbc.unsqueeze(0).expand(sim_state.n_systems, -1)
+        else:
+            pbc_batched = pbc
```
thomasloux, line 473: pbc is still a general property for all systems if I'm not mistaken. So technically pbc is always ndim == 1.

torch_sim/models/lennard_jones.py
```diff
@@ -441,3 +420,132 @@ def forward(
...
+                forces = torch.zeros_like(positions)
+                forces.index_add_(0, mapping[0], -force_vectors)
+                forces.index_add_(0, mapping[1], force_vectors)
+                results["forces"] = forces
+
+            if self.compute_stress:
+                stress_per_pair = torch.einsum("...i,...j->...ij", dr_vec, force_vectors)
+                volumes = torch.abs(torch.linalg.det(row_cell))
+                stress = torch.zeros(
+                    (n_systems, 3, 3),
```
thomasloux, line 533: would be more readable to separate that into multiple methods/functions.

torch_sim/models/lennard_jones.py
```diff
@@ -441,3 +420,132 @@ def forward(
...
+            sim_state = ts.SimState(
+                positions=positions,
+                masses=torch.ones_like(positions),
+                cell=state_dict["cell"],
+                pbc=state_dict["pbc"],
+                atomic_numbers=state_dict["atomic_numbers"],
+                system_idx=state_dict.get("system_idx"),
+            )
+
+        if sim_state.system_idx is None and sim_state.cell.shape[0] > 1:
```
thomasloux, line 451: system_idx can't be None, except if manually modified.

torch_sim/models/lennard_jones.py
```diff
@@ -441,3 +420,132 @@ def forward(
                 results[key] = torch.cat([out[key] for out in outputs], dim=0)
 
         return results
+
+
+class LennardJonesModel(UnbatchedLennardJonesModel):
```
thomasloux, line 425: I'm not a fan of inheriting from Unbatched and make it batch. It would be better to have a `_LennardJonesModel` then have UnbachedLennardJonesModel and LenanrdJonesModel inheriting from it.

torch_sim/models/lennard_jones.py
abhijeetgangan: If you pass `state_dict` it can. The type checker caught it.

torch_sim/models/lennard_jones.py
abhijeetgangan: Yeah, we can make the general for any pair wise force fields.

CompRhys: https://github.com/TorchSim/torch-sim/tree/batched-pairfunction I had this as a wip locally based on wanting to batch all the other pairwise and delete more code but was waiting for #476 to merge first for qol on getting ty not to complain

abhijeetgangan: Yeah, let's get the merge. Maybe we should just batch all other classical force fields.

CompRhys: yep that was the aim with this PR, you could delete all the current classical implementations and then just have named backward compatible things with hard coded pair functions

CompRhys:
Merging in current state to with plan to revisit whether to consolidate after #483

PR #466, opened 2026-02-24
Standardize parameter typing to float | torch.Tensor when appropriate

CompRhys:
see #325

I am unsure if step functions should allow floats because then we have overhead at the integration level. Init functions are a one off cost not a linear cost.

PR #339, opened 2025-11-10
Run `ty` in lint CI

janosh:
with number of PRs, contributors and users increasing, it's more important to guard against easy-to-catch type errors. lessens burden on reviewers and allows to merge green PRs with more confidence

PR #476, opened 2026-03-01
Restore non-conflicting changes from #339 (ty linting setup) 

CompRhys:
In #339 we lost some intended changes from #460, #466 and #470. This PR tries to revert them whilst keeping ty checks to zero. Doing this required some funny business with `__set_attr__` on `SimState`.  There were also breaking changes to the soft-sphere model as making it type correct with multiple sources of species_idx was too messy and I think there was an uncaught bug in the logic also. 

There are a few changes that were left but might be worth reverting such as `_randn_for_state` using it's own state.rng does mean that we repeatedly call the coerce_prng method each call to the rng.

### DIFF
--- a/tests/models/test_fairchem.py
+++ b/tests/models/test_fairchem.py
@@ -22,8 +22,8 @@ try:
 
 except (ImportError, OSError, RuntimeError, AttributeError, ValueError):
     pytest.skip(
-        f"FairChem not installed: {traceback.format_exc()}",
-        allow_module_level=True,  # ty:ignore[too-many-positional-arguments]
+        f"FairChem not installed: {traceback.format_exc()}",  # ty:ignore[too-many-positional-arguments]
+        allow_module_level=True,
     )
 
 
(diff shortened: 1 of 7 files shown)


CompRhys:
## Problem:
ty excludes dataclass fields shadowed by `@property` descriptors from the synthesized `__init__`, causing unknown-argument errors for `pbc` and `system_idx` across all `SimState` subclasses.

## Solution:
Removed the `@property` descriptors for pbc and system_idx
Added `__setattr__` to `SimState` for runtime coercion (`bool/list[bool]` → `Tensor`, `None` → `torch.zeros(...)`)
Narrowed field declarations to torch.Tensor (so `state.pbc` is always `Tensor`)
Added a TYPE_CHECKING-only `__init__ `stub to `SimState` widening `pbc` to `Tensor | list[bool] | bool` and `system_idx` to `Tensor | None` at call sites

## Limitation:
The wider input types only apply to `SimState.__init__`. Subclasses get ty-synthesized `__init__`s from the narrow field type, so `MDState(..., pbc=True)` is still flagged. Fixing subclasses would require per-class stubs or accepting a wider field type. This isn't a major issue since most downstream things are made using `.from_state` and forcing them to be more strongly typed is fine but if someone really wishes then it will behave as the parent class and coerce.

orionarcher (approved):
Scrolled through and everything seemed right.

PR #477, opened 2026-03-01
Restore #460 seed-removal semantics after #339

janosh:
- restore intentional #460 API behavior where NVE/NVT/NPT init functions do **not** take a `seed` kwarg and instead use `state.rng`
- switch integrator init paths back to `initialize_momenta(..., generator=state.rng)` and remove the reintroduced `calculate_momenta` helper
- update affected examples/tutorials and test helper to follow state-bound RNG seeding (`state.rng = ...`)

## Context
These are surgical restorations of intentional changes from #460 that were unintentionally lost during the large `ty`-fixing merge in #339.

@CompRhys sorry about this oversight on my side while polishing the #339 LLM type fixes. I think this surgical restore might be easier to review than the broader revert path in #476.

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -660,6 +660,19 @@ def test_multiple_constraints_and_dof(
     assert torch.allclose(final_com, initial_com, atol=1e-5)
 
 
+def _assert_dof_per_system(
+    state: ts.SimState, constraint_list: list[Constraint], expected_dof: list[int]
+) -> None:
+    """Assert per-system DOF values for a given state and constraint list."""
+    computed_dof = count_degrees_of_freedom(state, constraint_list)
+    expected_dof_tensor = torch.tensor(
+        expected_dof,
+        device=state.device,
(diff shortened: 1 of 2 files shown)


thomasloux:
Overall looks good to me, but I haven't checked that it covered all cases of seed.

thomasloux:
@janosh actually I've added `count_degrees_of_freedom` when I added constraints. But to be honest, this function is not used in torch_sim and actually a user should use state.constraints = my_constraints and then use state.get_degrees_of_freedom. I'm preparing a PR to remove this function.

thomasloux:
If it happens that some people use this function, actually the fix is change the type annotation. It should return a torch.Tensor and not a int. That's actually the error on my side + changing the check for non negativity.

CompRhys:
> @CompRhys sorry about this oversight on my side while polishing the #339 LLM type fixes.

No harm :) as long as we end up eventually correct when we make a release that's all that matters

CompRhys:
Will merge this and let #479 resolve itself separately

PR #479, opened 2026-03-01
Remove unnecessary count dof function

thomasloux:
## Summary

After `ty` checks, I actually realized I added a `count_degrees_of_freedom` function in constraints.py but one should use directly the method state.get_degrees_of_freedom to get this information. I suggest to remove this unused and redundant function.

## Checklist

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -658,6 +658,7 @@ def test_multiple_constraints_and_dof(
     )
     assert torch.allclose(final_com, initial_com, atol=1e-5)
 
+
 @pytest.mark.parametrize(
     ("cell_filter", "fire_flavor"),
     [


thomasloux:
What I don't like with this function is that it may give the impression that you can provide arbitrary constraints and state, which technically you could. But this function does not test that constraints are compatible with the state. For example if you FixAtom with index 10 but your system has 2 atoms, it will still remove degrees of freedom.

We can keep it, but then it would add a checker for compatibility of constraints and state. This is the normal behaviour when doing state.constraints = myConstraints.

thomasloux:
I let this PR open for now so that we can discuss about it.

CompRhys:
closing in favour of #480 because it's nice to leave functional blocks of helper code for hackers.

PR #480, opened 2026-03-01
Enforce compatible constraints in DOF helper

janosh:
## Summary
- keep `count_degrees_of_freedom` as a public helper, but validate explicitly supplied constraints against the provided `state` before counting
- centralize shared per-system DOF math in `_dof_per_system` and reuse it from both `count_degrees_of_freedom` and `SimState.get_number_of_degrees_of_freedom`
- add regression tests for out-of-bounds constraints and malformed `get_removed_dof` outputs (wrong shape, negative, non-tensor, non-finite), plus helper/method parity and strict-vs-clamped behavior

## Context
This follows the API discussion in #479 and the end-of-thread discussion in #477: keep the helper for external use, but make it safe and explicit about compatibility checks instead of allowing arbitrary mismatched state/constraint pairs.

Refs:
- https://github.com/TorchSim/torch-sim/pull/479
- https://github.com/TorchSim/torch-sim/pull/477

## Test plan
- `pytest tests/test_constraints.py -q`

### DIFF
--- a/torch_sim/constraints.py
+++ b/torch_sim/constraints.py
@@ -616,21 +616,21 @@ def count_degrees_of_freedom(
     """Count per-system degrees of freedom with compatibility checks.
 
     This helper computes one DOF value per system. When ``constraints`` are
-    supplied explicitly, it validates that they are compatible with ``state``
-    before counting.
+    supplied, it validates that they are compatible with ``state`` before
+    counting.
 
     Args:
         state: Simulation state
-        constraints: Optional constraints to evaluate. If ``None``, uses
-            ``state.constraints``.
+        constraints: Constraints to evaluate. If ``None``, returns unconstrained
+            DOF (3 * n_atoms_per_system). Use ``state.get_number_of_degrees_of_freedom()``
+            to count with state-attached constraints.
 
     Returns:
         Degrees of freedom per system as a tensor of shape (n_systems,)
     """
-    constraints_to_count = state.constraints if constraints is None else constraints
     if constraints is not None:
-        _validate_constraints_for_dof(state, constraints_to_count)
-    return torch.clamp(_dof_per_system(state, constraints_to_count), min=0)
+        validate_constraints(constraints, state)
+    return torch.clamp(_dof_per_system(state, constraints), min=0)
 
 
 def _dof_per_system(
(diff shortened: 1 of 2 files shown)


torch_sim/constraints.py
```diff
@@ -613,28 +613,74 @@ def __repr__(self) -> str:
...
+    dof_per_system = 3 * state.n_atoms_per_system
     if constraints is not None:
         for constraint in constraints:
-            total_dof -= constraint.get_removed_dof(state)
+            dof_per_system -= constraint.get_removed_dof(state)
+    return dof_per_system
+
 
-    return torch.clamp(total_dof, min=0)
+def _validate_constraints_for_dof(state: SimState, constraints: list[Constraint]) -> None:
```
thomasloux: I think that `validate_constraints` is enough. The other if-statements should be tests perform by pytest.

torch_sim/constraints.py
```diff
@@ -613,28 +613,74 @@ def __repr__(self) -> str:
...
-        constraints: List of active constraints (optional)
+        constraints: Optional constraints to evaluate. If ``None``, uses
+            ``state.constraints``.
 
     Returns:
         Degrees of freedom per system as a tensor of shape (n_systems,)
     """
-    # Start with unconstrained DOF per system
-    total_dof = 3 * state.n_atoms_per_system
+    constraints_to_count = state.constraints if constraints is None else constraints
```
thomasloux: Why not but to me it is simpler to only consider the input constraints. Otherwise the user should use state.get_degrees_of_freedom()

torch_sim/constraints.py
```diff
@@ -613,28 +613,74 @@ def __repr__(self) -> str:
...
+    state: SimState, constraints: list[Constraint] | None = None
+) -> torch.Tensor:
+    """Compute unconstrained-minus-removed DOF per system."""
+    dof_per_system = 3 * state.n_atoms_per_system
     if constraints is not None:
         for constraint in constraints:
-            total_dof -= constraint.get_removed_dof(state)
+            dof_per_system -= constraint.get_removed_dof(state)
+    return dof_per_system
+
```
thomasloux, line 645: I'm not convinced by the value of separating these two helper functions

thomasloux: Ok I saw that you used it with state.get_degrees_of_freedom. Then I'm fine with the 2 functions

CompRhys:
@thomasloux mergeable?

janosh:
i didn't have time to look into the comments yet. unless you think they're not actionable, not mergeable yet

janosh:
took a closer look now. great suggestions, thanks @thomasloux! addressed in 6d01cb2:

- `_validate_constraints_for_dof` removed — agreed that the type/shape/device/finiteness checks on `get_removed_dof()` output are implementation contract validation, not user-input validation. better validated in unit tests, not at runtime. pept only `validate_constraints()` for index-bounds checking
- `state.constraints` fallback removed — `count_degrees_of_freedom(state)` now returns unconstrained DOF. Users who want state-attached constraints should use `state.get_number_of_degrees_of_freedom()`.
- i kept `_dof_per_system`. as you noted, it's reused by `SimState.get_number_of_degrees_of_freedom()`

PR #483, opened 2026-03-02
Implement general pair-potential and pair-force

CompRhys:
Batch all the things.

Adds general Pair Potential and Pair Force models (allows asym forces). Testing consistency against older classical models for now. Considering leaving classical potentials for now but aim would be to deprecate and remove the implementations and just use the energy functions/force functions here. MultiSoftSphere has largest API changes to the args for setting up the matricies.


torch_sim/models/soft_sphere.py
```diff
@@ -121,7 +121,7 @@ def soft_sphere_pair_force(
     """
 
     def fn(dr: torch.Tensor) -> torch.Tensor:
-        return (-epsilon / sigma) * (1.0 - (dr / sigma)).pow(alpha - 1)
+        return (epsilon / sigma) * (1.0 - (dr / sigma)).pow(alpha - 1)
```
CompRhys, line 124: testing against pair potential showed a bug in soft-sphere stress.

abhijeetgangan: Good catch. Looking at the original paper there shouldn't be a minus - https://doi.org/10.1103/PhysRevE.68.011306

PR #484, opened 2026-03-02
Configure Basic Logging.

CompRhys:
Log all the things.

using f-strings in logs are bad because it forces them to instantiate even if filtered, should really remove the ruff rule ignore added. 

logs by default are written to a shared file which might be weird if multiple things are run simultaneously.

what do we want to log and at what level?


torch_sim/autobatching.py
orionarcher, line 1171: Maybe we can add a log here of the number of states that completed and how many are in the current batch?

torch_sim/runners.py
```diff
@@ -683,6 +698,10 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
                     f"All systems have reached the maximum number of steps: {max_steps}.",
                     stacklevel=2,
                 )
+                logger.warning(
+                    "optimize: all systems reached max_steps=%d without converging",
```
orionarcher:
```suggestion
                    "optimize: all systems in batch reached max_steps=%d without converging",
```

torch_sim/autobatching.py
orionarcher, line 732: Can we log the completed batch and next batch sizes, akin to the suggestion in InFlight

PR #486, opened 2026-03-02
Bump tables version and use py314 for highest tests

CompRhys:
bleeding edge development.


thomasloux:
love these commit names haha

PR #488, opened 2026-03-03
Batched pairfunction cleanup

CompRhys:
Replace all the old classical pair potential backends with general backend, make this differentiable to enable `diff_sim.py` to work.

### DIFF
--- a/torch_sim/models/pair_potential.py
+++ b/torch_sim/models/pair_potential.py
@@ -9,21 +9,60 @@ It also provides :class:`PairForcesModel` for potentials defined directly as for
 (e.g. the asymmetric particle-life interaction) that cannot be expressed as the
 gradient of a scalar energy.
 
-Concrete pair functions live in their respective model modules:
+The pair function signature required by :class:`PairPotentialModel` is:
+``pair_fn(distances, atomic_numbers_i, atomic_numbers_j) -> pair_energies``,
+where all arguments are 1-D tensors of length n_pairs and the return value is a
+1-D tensor of pair energies. Additional parameters (e.g., ``sigma``, ``epsilon``)
+can be bound using :func:`functools.partial`.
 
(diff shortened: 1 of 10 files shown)


CompRhys:
The PR was looking so clean until these tests broke. They all use LJ models and have numerical consistency issues. Not sure what the bug might be :(

examples/tutorials/diff_sim.py
```diff
@@ -9,17 +9,19 @@
 # </details>
 
 # %%
+import typing
```
abhijeetgangan, line 12: You can remove the example because it seems out of scope for torchsim.

orionarcher: I think having a differentiable example for classical potentials is nice

tests/test_integrators.py
```diff
@@ -339,7 +339,7 @@ def test_nvt_nose_hoover(ar_double_sim_state: ts.SimState, lj_model: LennardJone
     temperatures_list = [t.tolist() for t in temperatures_tensor.T]
     assert torch.allclose(
         temperatures_tensor[-1],
-        torch.tensor([299.9910, 299.6800], dtype=dtype),
+        torch.tensor([300.0096, 299.7024], dtype=dtype),
```
orionarcher, line 342: Is this degree of precision justified?

torch_sim/models/pair_potential.py
```diff
@@ -9,26 +9,24 @@
 (e.g. the asymmetric particle-life interaction) that cannot be expressed as the
 gradient of a scalar energy.
```
orionarcher, line 10: More documentation of new design perhaps?

tests/test_integrators.py
CompRhys: This was a magic seeded test and the change in the backend caused the numbers to jump. We could say +/- 1 on the all close and remove the seed but then we would have to contend with it potentially becoming flaky

PR #489, opened 2026-03-04
Add log line clarifying BinningAutoBatcher is initialization-only in optimize

orionarcher:
## Summary

- Adds a single `logger.info` call in `optimize` (runners.py) just before the `_chunked_apply` call that initializes optimizer state
- Makes clear in logs that BinningAutoBatcher output during this step is one-time initialization, and that InFlightAutoBatcher will be used for the actual optimization loop

## Motivation

Users reading logs saw BinningAutoBatcher output and were confused about which autobatcher was active during optimization. This one-line log message removes that ambiguity with no refactoring.

### DIFF
--- a/torch_sim/trajectory.py
+++ b/torch_sim/trajectory.py
@@ -186,6 +186,7 @@ class TrajectoryReporter:
             ValueError: If filenames are not unique
         """
         self.finish()
+        self.trajectories = []  # drop refs so HDF5 finalizes before new opens
 
         filenames = (
             [filenames] if isinstance(filenames, (str, pathlib.Path)) else list(filenames)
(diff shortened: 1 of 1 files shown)


orionarcher:
@CompRhys this was not ready to merge as

```
        # TODO FIX THIS
        if hasattr(tables, "file") and (
            handles := tables.file._open_files.get_handlers_by_name(str(filename))
        ):
            list(handles)[-1].close()
```

was part of an incomplete second change. Next time can mark as draft. Will deal with file handling in a separate PR.

CompRhys:
oops, I thought the deleted TODO was that the list creation above fixed any issue.

orionarcher:
unclear... I was having some HDF5 issues and this was the vibe coded solution that didn't actually fix the problem. More digging needed. Certainly possible these changes are valid but did not address my problem.

PR #490, opened 2026-03-04
Fix FIRE NaN velocity sentinel corrupting retained systems during autobatcher swaps

janosh:
Fixes a bug in `_ase_fire_step` where the NaN-velocity initialization sentinel skips the entire FIRE step (force transformation, power calculation, dt/alpha/n_pos updates, velocity mixing) for **all** systems when **any** system has NaN velocities.

This matters when the `InFlightAutoBatcher` swaps in a new state mid-optimization — the new state has NaN velocities from `fire_init`, which causes retained systems to skip a FIRE step entirely.

### Evidence

Clone-and-compare test: two identical systems, 10 FIRE steps, then inject NaN velocities into system 1 only and run one more step. System 0 should be completely unaffected:

| Metric | Main (buggy) | With fix |
|--------|-------------|----------|
| Position diff | `3.76e-04` | `0.0` |
| dt diff | `1.33e-02` | `0.0` |
| alpha diff | `9.70e-04` | `0.0` |
| n_pos diff | `1` | `0` |

`test_fire_nan_velocities_dont_affect_other_systems` fails on all 4 parametrizations on main, passes on all 4 with the fix.

### Fix

Decouple NaN zeroing from the FIRE logic branch. Only skip FIRE mixing when `nan_velocities.all()` (first step, matching ASE). When a subset has NaN (autobatcher swap), zero them and run normal FIRE — newly zeroed systems naturally get `power=0` → negative mask → dt decrease and velocity reset.

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -716,7 +716,7 @@ def test_fire_vv_negative_power_branch(
 
 
 @pytest.mark.parametrize("fire_flavor", get_args(FireFlavor))
-@pytest.mark.parametrize("cell_filter", [None, ts.CellFilter.frechet])
+@pytest.mark.parametrize("cell_filter", [None, ts.CellFilter.unit, ts.CellFilter.frechet])
 def test_fire_nan_velocities_dont_affect_other_systems(
     ar_supercell_sim_state: SimState,
     lj_model: ModelInterface,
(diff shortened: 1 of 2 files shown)


janosh:
failing CI is intermittent MACE download issue. @orionarcher do you want to review? if not, @abhijeetgangan feel free to merge?

PR #491, opened 2026-03-04
Fix GPU memory leak from missing .detach() in model wrappers

reillyosadchey:
## Summary

- Several model wrappers return tensors still attached to the autograd computation graph, causing the entire forward-pass graph to be retained in GPU memory across simulation steps
- Added `.detach()` calls to model outputs in `fairchem`, `orb`, `metatomic`, `fairchem_legacy`, and `graphpes_framework`
- The `fairchem_legacy` wrapper also used `.clone()` without `.detach()` on inputs, retaining graph references via `self.data_object`

## Details

| Model | Issue | Severity |
|---|---|---|
| `fairchem.py` | energy/forces/stress not detached | High |
| `orb.py` | predictions + conservative forces/stress not detached | High |
| `metatomic.py` | forces/stress not detached (energy was correct) | Medium |
| `fairchem_legacy.py` | `.clone()` without `.detach()` on inputs | Medium |
| `graphpes_framework.py` | external library predictions not detached | Low |

Without `.detach()`, `.to(dtype=...)` and `.view()` create new graph nodes that still reference the full forward pass. Integrators store these into `state.energy`/`state.forces`/`state.stress`, keeping the graph alive until the next step overwrites them — and longer if trajectory loggers or MC routines `.clone()` the state.

## Test plan

- [ ] Existing tests pass (no functional change — detach only affects graph retention)
- [ ] Verify GPU memory stays flat over long simulations with FairChem/Orb models

### DIFF
--- a/torch_sim/models/fairchem.py
+++ b/torch_sim/models/fairchem.py
@@ -243,15 +243,15 @@ class FairChemModel(ModelInterface):
 
         # Convert predictions to torch-sim format
         results: dict[str, torch.Tensor] = {}
-        results["energy"] = predictions["energy"].detach().to(dtype=self._dtype)
-        results["forces"] = predictions["forces"].detach().to(dtype=self._dtype)
+        results["energy"] = predictions["energy"].to(dtype=self._dtype)
+        results["forces"] = predictions["forces"].to(dtype=self._dtype)
 
         # Handle stress if requested and available
         if self._compute_stress and "stress" in predictions:
-            stress = predictions["stress"].detach().to(dtype=self._dtype)
+            stress = predictions["stress"].to(dtype=self._dtype)
             # Ensure stress has correct shape [batch_size, 3, 3]
             if stress.dim() == 2 and stress.shape[0] == len(atomic_data_list):
                 stress = stress.view(-1, 3, 3)
             results["stress"] = stress
 
-        return results
+        return {k: v.detach() for k, v in results.items()}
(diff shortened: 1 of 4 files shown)


CompRhys:
Thank you for the PR I will merge it once tests and lint pass.

It is worth noting that we are attempting to move all the models you fixed this for to external posture. These fixes may also need to be applied upstream.

In the pair potential backend refactor I am doing I also included a [retrain graph argument ](https://github.com/TorchSim/torch-sim/pull/488/changes#diff-e6724f79e75d566c0accf7d9f7e1b2dc76aa08b94881a644f8a2659757a6ad76R418) to allow for diff-sim tutorial. I think that's a very niche use so wouldn't add the flag here but just an FYI

torch_sim/models/graphpes_framework.py
```diff
@@ -185,4 +185,5 @@ def forward(self, state: ts.SimState, **_kwargs: object) -> dict[str, torch.Tens
         if not isinstance(cutoff, torch.Tensor):
             raise TypeError("GraphPES model cutoff must be a tensor")
         atomic_graph = state_to_atomic_graph(state, cutoff)
-        return self._gp_model.predict(atomic_graph, self._properties)  # type: ignore[return-value]
+        preds = self._gp_model.predict(atomic_graph, self._properties)  # ty: ignore[call-non-callable]
+        return {k: v.detach() for k, v in preds.items()}
```
CompRhys, line 189: It feels cleaner to be to detatch everything like this all in one line at the end? Could you update all the models to follow this pattern?

CompRhys: if it's already detached then the op is idempotent and so no harm.

reillyosadchey: Good call

CompRhys:
Thanks!

PR #494, opened 2026-03-05
Check if models are detached

CompRhys:
Move to external posture undermined by memory issues of several local interfaces potentially not detaching tensors.

default is set here to not break CI for others using the testing utils we provide at this point.

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -629,6 +629,10 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
     if isinstance(initial_state, OptimState):
         state = initial_state
     else:
+        logger.info(
+            "optimize: initializing optimizer state via chunked apply "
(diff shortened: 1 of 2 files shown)


orionarcher:
Is the intent here to add `retain_graph` as a formal part of the `ModelInterface`?

CompRhys:
> Is the intent here to add `retain_graph` as a formal part of the `ModelInterface`?

Not really, I added it in https://github.com/TorchSim/torch-sim/pull/488 to the pairwise to allow the diff-sim tutorial to work. It's a sort of reasonable kwarg name to use for that but more a "secret menu" thing. This PR was mostly in response to https://github.com/TorchSim/torch-sim/pull/491. It does seem reasonable to set the default behavior that results should be detached and then there should be some way for people to hack around that if they really want.

PR #497, opened 2026-03-09
Use upstreamed MACE model

CompRhys:
Uses interface introduced in https://github.com/ACEsuit/mace/pull/1390. Awaiting release for minimal pin

### DIFF
--- a/examples/scripts/1_introduction.py
+++ b/examples/scripts/1_introduction.py
@@ -6,7 +6,10 @@ This script demonstrates the fundamental usage of TorchSim with:
 """
 
 # /// script
-# dependencies = ["scipy>=1.15", "mace-torch>=0.3.12"]
+# dependencies = [
+#     "torch_sim_atomistic[mace, io]",
+#     "mace-torch @ git+https://github.com/CompRhys/mace.git@main",
+# ]
 # ///
 
 import itertools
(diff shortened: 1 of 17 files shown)


examples/scripts/1_introduction.py
```diff
@@ -6,7 +6,10 @@
 """
 
 # /// script
-# dependencies = ["scipy>=1.15", "mace-torch>=0.3.12"]
+# dependencies = [
+#     "torch_sim_atomistic[mace, io]",
+#     "mace-torch @ git+https://github.com/ACEsuit/mace.git@develop",
```
CompRhys: can remove all these once a new mace release comes. I would guess that we should delay to merge this until then but that may not be for a few months

tests/models/test_mace.py
```diff
@@ -15,38 +20,65 @@
...
-    raw_mace_omol = mace_omol(model="extra_large", return_raw_model=True)
-    HAS_MACE_OMOL = True
-except (ImportError, OSError, RuntimeError, AttributeError, ValueError):
-    raw_mace_omol = None
-    HAS_MACE_OMOL = False
 
-raw_mace_mp = mace_mp(model=MaceUrls.mace_mp_small, return_raw_model=True)
-raw_mace_off = mace_off(model=MaceUrls.mace_off_small, return_raw_model=True)
-DTYPE = torch.float64
+def _download_with_retry(fn: Callable, **kwargs: Any) -> Any:
```
CompRhys, line 34: this didn't help to make the test not flaky.

PR #498, opened 2026-03-10
update momenta before integration in npt NH

alphalm4:
Hello, thanks for offering such a great project!

I've got catastrophic failure in MD when using NPT Nose-Hoover integration. I believe the momenta update seems accidently dropped before position update. Adding momenta update gives no such failure.

## Summary

* Add momenta update before position update
* Add stress attribute in NPT nose hoover state

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@ docs/reference/torch_sim.*
 *.hdf5
 *.traj
 
+# ignore torch.save outputs
+*.pt
+
 # coverage
 coverage.xml
 .coverage*
--- a/docs/user/reproducibility.md
+++ b/docs/user/reproducibility.md
@@ -79,17 +79,34 @@ Because TorchSim runs batched simulations, all systems in a batch share a single
 
 If strict reproducibility is required, keep your batching setup fixed.
 
-### Serialising the RNG state
+### Serialising state for reproducible restarts
 
-If you wish to be able to resume a session and ensure determinism you need to persist and reload the `torch.Generator` state. This can be done using `torch.save()` and `torch.Generator().set_state()`:
+To resume a simulation and ensure determinism you need to persist and reload the complete state, including the `torch.Generator` RNG state. The simplest approach is to save the full state dict with `torch.save()`:
 
 ```python
+from dataclasses import asdict
+from torch_sim.integrators import MDState
+
 # save
+torch.save(asdict(state), "checkpoint.pt")
+
+# restore (weights_only=False needed for torch.Generator in PyTorch 2.6+)
+restored = MDState(**torch.load("checkpoint.pt", weights_only=False))
+```
+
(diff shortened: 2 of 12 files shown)


orionarcher:
What model are you using?

alphalm4:
> What model are you using?

@orionarcher  I'm using SevenNet-Omni with the inference channel `omol25_low`. I'm currently porting my MD jobs from LAMMPS to torch-sim to take advantage of its batch performance. For reference, I’ve attached trajectories from before/after the commit, along with the reference MD trajectory from LAMMPS.

[for_ts_pr.zip](https://github.com/user-attachments/files/25874887/for_ts_pr.zip)

CompRhys:
image
Looking at the data you share does suggest this is resulting in a trajectory closer to the lammps reference

orionarcher:
@abhijeetgangan @thomasloux what are your thoughts?

thomasloux:
I think he's right. I'll have to get a second read in the source equations. Because it's really hard to read.

Note: adding stress in state is potentially a good idea. Although you probably want it for all NPT MD. So potentially a NptMdState (terrible name) could be created to save the stress, all NPT integrator state would inherit from it. It would be easier to save it in trajectory.

thomasloux:
We don't exactly have the exact same convention as the source. I added the source, I'm almost sure the source is the exact same equation for NVT nose hoover. But there is some slight differences with torch sim NPT nose hoover. 
I think there is a swap of name between L1 and L2. 

The equation should be checked carefully. I'm not even sure that this fix correspond exactly to the equations.

Thanks for finding this issue!

thomasloux:
As a note, most of integrators except C/V-rescale come from JAX-MD. In JAX-MD the code actually indeed perform the same implementation as you suggested: https://github.com/jax-md/jax-md/blob/a2c11c18ce65606983824c5cb0d17f5569abb3a1/jax_md/simulate.py#L910

They don't use `state.velocities` but get the results of L2 (so momenta) and provide directly `momenta/mass`, but it is equivalent to your proposal @alphalm4

thomasloux:
In the meantime, could you give isotropic NPT C-Rescale a try to check that it has a behaviour coherent with what you would expect. Of course it's going to be different than NH, but I would really like to know whether my implementation is indeed correct :)

CompRhys:
> As a note, most of integrators except C/V-rescale come from JAX-MD. In JAX-MD the code actually indeed perform the same implementation as you suggested: https://github.com/jax-md/jax-md/blob/a2c11c18ce65606983824c5cb0d17f5569abb3a1/jax_md/simulate.py#L910
> 
> They don't use `state.velocities` but get the results of L2 (so momenta) and provide directly `momenta/mass`, but it is equivalent to your proposal @alphalm4

based on this comment I would merge?

thomasloux:
Let me have the week end to check the sources and equations. I would also prefer to create the NPTMDState with stress, but yes I think we'll be able to merge it. I can add a commit in this PR for the previous class I mentioned.

CompRhys:
thanks for finding this!

Issue #485, opened 2026-03-02
Why checking for only single system when providing system list in `initialize_state`


thomasloux:
```
def initialize_state(  # noqa: C901, PLR0911
    system: StateLike,
    device: torch.device | None = None,
    dtype: torch.dtype | None = None,
) -> SimState:
    # TODO: create a way to pass velocities from pmg and ase

    if isinstance(system, SimState):
        return system.clone().to(device, dtype)

    if isinstance(system, list | tuple):
        if len(system) == 0:
            raise ValueError("Cannot initialize state from an empty list.")
        states = [s for s in system if isinstance(s, SimState)]
        **if len(states) == len(system) and not all(s.n_systems == 1 for s in states):
            raise ValueError(
                "When providing a list of states, to the initialize_state function, "
                "all states must have n_systems == 1. To fix this, you can split the "
                "states into individual states with the split_state function."
            )**
        if len(states) == len(system):
            return ts.concatenate_states(states)
```
Why does this function impose this constraint? Not sure it's a problem for anyone in practise, but this seems quite arbitrary. Actually `concatenate_states` handle list of state with arbitrary system sizes.

orionarcher:
If tests pass with a zero length system I'd be fine removing it. I think this is just a hedge against edge cases.

Issue #379, opened 2025-12-10
Remove last_permutation from SwapMCState initialization


curtischong:
When doing a MC swap, we need to pass in `last_permutation`. But I think the SwapMCState can just figure out an initial last_permutation automatically (no need to pass it into the state at first?). Is there a strong reason why we need to pass it in?

```
@dataclass(kw_only=True)
class SwapMCState(SimState):
    """State for Monte Carlo simulations with swap moves.

    This class extends the SimState to include properties specific to Monte Carlo
    simulations, such as the system energy and records of permutations applied
    during the simulation.

    Attributes:
        energy (torch.Tensor): Energy of the system with shape [batch_size]
        last_permutation (torch.Tensor): Last permutation applied to the system,
            with shape [n_atoms], tracking the moves made for analysis or reversal
    """

    energy: torch.Tensor
    last_permutation: torch.Tensor

    _atom_attributes = SimState._atom_attributes | {"last_permutation"}  # noqa: SLF001
    _system_attributes = SimState._system_attributes | {"energy"}  # noqa: SLF001
```

Maybe a better approach is to make it `| None` so people don't need to pass it in when they create this swap state?

orionarcher:
Yeah that makes sense, we can modify the `__post_init__` to add in a default last swap.

PR #501, opened 2026-03-15
Nits rollup

CompRhys:
Show a way to save and restart that gives same results. This should resolve https://github.com/TorchSim/torch-sim/issues/43.

Allow lists of different size simstates to be combined. This should solve #485.

Allow MC state to create the last permutation itself. This solve solve #379. --> solution here is a different pattern to https://github.com/TorchSim/torch-sim/pull/464 and so may not be long lived.

### DIFF
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -1225,8 +1225,6 @@ def initialize_state(
 
     if isinstance(system, list | tuple) and all(isinstance(s, SimState) for s in system):
         system: list[SimState] = typing.cast("list[SimState]", system)
-        if len(system) == 0:
-            raise ValueError("Cannot initialize state from an empty list.")
         return ts.concatenate_states(system)
 
     converters = [


torch_sim/integrators/npt.py
```diff
@@ -574,17 +574,15 @@ def npt_langevin_init(
     model_output = model(state)
 
     # Initialize momenta if not provided
-    momenta = getattr(
-        state,
-        "momenta",
-        initialize_momenta(
```
CompRhys, line 580: this always eval'd which advances rng state.

torch_sim/state.py
```diff
@@ -1225,12 +1225,8 @@ def initialize_state(
...
     if isinstance(system, list | tuple) and all(isinstance(s, SimState) for s in system):
         system: list[SimState] = typing.cast("list[SimState]", system)
-        if not all(state.n_systems == 1 for state in system):
-            raise ValueError(
-                "When providing a list of states, to the initialize_state function, "
-                "all states must have n_systems == 1. To fix this, you can split the "
-                "states into individual states with the split_state function."
-            )
+        if len(system) == 0:
+            raise ValueError("Cannot initialize state from an empty list.")
```
CompRhys:
```suggestion
```

PR #507, opened 2026-03-17
Adjust test tol to resolve false positive test failures

CompRhys:
reduce tol to fp32 levels of 1e-6 rtol and 1e-7 atol.

allow for a zero force provided not all are zero.

### DIFF
--- a/tests/models/test_pair_potential.py
+++ b/tests/models/test_pair_potential.py
@@ -199,7 +199,7 @@ def test_autograd_force_fn_matches_potential_model(
     out_pp = model_pp(sim_state)
     out_pf = model_pf(sim_state)
 
-    assert (out_pp["forces"] != 0.0).any(), (
+    assert (out_pp["forces"] != torch.zeros_like(out_pp["forces"])).any(), (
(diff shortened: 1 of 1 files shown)


tests/models/test_pair_potential.py
```diff
@@ -199,7 +199,7 @@ def force_fn(dr: torch.Tensor, zi: torch.Tensor, zj: torch.Tensor) -> torch.Tens
     out_pp = model_pp(sim_state)
     out_pf = model_pf(sim_state)
 
-    assert (out_pp["forces"] != 0.0).all()
```
thomasloux, line 202: `!=` on float value is not great, could you replace it by `>=tol` with tol=1e-6 for instance

CompRhys: this is a check for exact zero's which would be a logic failure as opposed to numerically close to zero which would be fine. It being brittle is okay the problem was that seemingly we got one force that evaluated as equal to zero in the latest torch

thomasloux: ok then could you just add a short comment about that, because it's really unusual to do that, good for me otherwise

PR #509, opened 2026-03-17
Fix `FixSymmetry` constraint `system_idx` remapping on reordered state slicing

danielzuegner:
### Problem

When slicing a `SimState` with out-of-order system indices (e.g. `state[[3, 1, 4]]`), `FixSymmetry`'s per-system data (`rotations`, `symm_maps`, `reference_cells`) gets applied to the **wrong systems**.

This is triggered during relaxation with `BinningAutoBatcher`, which produces non-sorted system indices from its descending-volume bin-packing algorithm.

### Root cause

`_filter_attrs_by_index` in `state.py` converts `system_indices` to a boolean mask before passing to `select_constraint`. Boolean masks lose ordering information — `torch.where(mask)` always returns indices in ascending order. While constraint `atom_idx` remapping already existed to correct for this, the analogous `system_idx` remapping was missing.

### Fix

Build both `atom_remap` and `system_remap` tables upfront, then remap both `atom_idx` and `system_idx` on each constraint in a single pass after `select_constraint`. This also consolidates the previously separate atom/system remapping loops.

### Test

Added two tests for out-of-order slicing:
- `test_fix_symmetry_system_idx_remapped_on_reordered_slice` — creates a 2-system state with distinct `FixSymmetry` rotations per system, slices with `state[[1, 0]]`, and verifies each output system is paired with the correct rotation and reference cell.
- `test_fix_com_system_idx_remapped_on_reordered_slice` — constrains only system 0 with `FixCom`, slices with `state[[1, 0]]`, and verifies `system_idx` is remapped to `[1]`.

Disclaimer: This is human-checked and verified LLM code :)
## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,8 @@ from torch_sim.models.lennard_jones import LennardJonesModel
 from torch_sim.testing import SIMSTATE_GENERATORS
 
 
+torch.set_num_threads(4)
+
 DEVICE = torch.device("cpu")
 DTYPE = torch.float64
 
--- a/tests/models/test_pair_potential.py
+++ b/tests/models/test_pair_potential.py
@@ -146,7 +146,7 @@ def test_half_list_matches_full(si_double_sim_state: ts.SimState, key: str) -> N
     model_half = PairPotentialModel(**common, reduce_to_half_list=True)
     out_full = model_full(si_double_sim_state)
     out_half = model_half(si_double_sim_state)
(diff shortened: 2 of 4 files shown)


CompRhys:
@falletta I think you changed this code in https://github.com/TorchSim/torch-sim/pull/457 can you take a look at this

falletta:
I tested the code on my use case, which required the fix in #457, and everything looks good to me.

CompRhys:
Offline @falletta confirmed this fix is consistent with his fix for his workflow. Thanks for the fix!

PR #510, opened 2026-03-17
Use upstreamed Metatomic model

CompRhys:
Uses interface added in https://github.com/metatensor/metatomic/pull/167#event-23646360964


CompRhys:
https://pypi.org/project/metatomic-torchsim/ is currently a placeholder wait on that to get real release.

PR #512, opened 2026-03-17
Add filtering of model tests based on file import closure

CompRhys:
This should mean that we only test models which have been affected by file changes wrt main. 

Actually this is insufficient with external posture as it doesn't check the external torchsim closure.


CompRhys:
concluded that the nature of filtering on tests makes this impossible to have a minimal solution.

Issue #437, opened 2026-02-03
Make neighborlist work with both wrapped and unwrapped coordinates


abhijeetgangan:
See #436 and #423 for more context. I think the current fix works but if the users access the neighborlist directly it might give wrong results.

CompRhys:
I think the source of the issues identified in #423 was actually that in #294 we swapped to default unwrapped positions but the test suite was insufficient to expose some resulting edge cases errors.

abhijeetgangan:
Yes, that was an issue indeed but I think there is definitely an unexpected behavior comparing two version of the neighborlist. Repro [here](https://github.com/NVIDIA/nvalchemi-toolkit-ops/issues/18) without torchsim.

alphalm4:
Dear devs, I'm here to ask about precision things regarding cell wrap-unwrap operations (rather than opening new issue since i feel here might be a correct place)

Related issue is that current `sevenn.torchsim.SevenNetModel` incorrectly creates graph using unwrapped positions,
https://github.com/MDIL-SNU/SevenNet/issues/299

I bypassed this by manually inserting `ts.transforms.pbc_wrap_batched` before graph generation.
https://github.com/MDIL-SNU/SevenNet/commit/ea447648247039abff38eee55dc9041f7f5705f6

But then high wrap-then-unwrap precision error from cell transform was occurred, since SevenNet only supports `float32` operation and all torch-sim operations inherit the position's precision (am i right?). Upcasting the operation resolved this.
https://github.com/TorchSim/torch-sim/commit/a858fbdb196b1e1c358fea8ab8a70175778c7a73

Does the related PR resolves this model-dependent precision things?

CompRhys:
The open PR should address https://github.com/MDIL-SNU/SevenNet/issues/299 as all options will wrap and unwrap correctly - alchemiops, vesin and the native. It also fixes the CPU import for alchemiops meaning that we will essentially never use the native. As such the precision changes you describe in native I don't think would impact anything. So that could remain an issue.

Perhaps we should add another test case to the validate model function to test this wrapped vs unwrapped equivalence as part of the interface definition.

PR #519, opened 2026-03-21
Add pbc_wrap_batched_and_get_lattice_shifts to handle unwrapped positions in nl

CompRhys:
Addresses https://github.com/TorchSim/torch-sim/issues/437

### DIFF
--- a/torch_sim/models/interface.py
+++ b/torch_sim/models/interface.py
@@ -381,7 +381,7 @@ def validate_model_outputs(  # noqa: C901, PLR0915
     # This catches models that fail to apply periodic boundary conditions.
     shifted_state = si_state.clone()
     lattice_vec = shifted_state.cell[0, :, 0]  # column convention
-    shifted_state.positions[0] = shifted_state.positions[0] + lattice_vec
+    shifted_state.positions[0] = shifted_state.positions[0] + 3 * lattice_vec
     shifted_output = model.forward(shifted_state)
     if not torch.allclose(
         shifted_output["energy"], si_model_output["energy"], atol=VALIDATE_ATOL


torch_sim/transforms.py
```diff
@@ -139,43 +139,68 @@ def pbc_wrap_batched(
...
+    pbc_mask = pbc_per_atom & active_per_atom
+
+    int_shifts = torch.where(pbc_mask, torch.floor(frac), torch.zeros_like(frac))
+    wrapped_frac = frac - int_shifts
+
+    cell_per_atom = cell_T[system_idx]
+    wrapped_pos = torch.bmm(cell_per_atom, wrapped_frac.unsqueeze(2)).squeeze(2)
+    out = torch.where(active_per_atom, wrapped_pos, positions)
+    shifts = torch.where(active_per_atom, int_shifts, torch.zeros_like(int_shifts))
+    return out, shifts
```
CompRhys, line 210: key idea was just to keep these extra shifts around and adjust the nl shifts accordingly.

CompRhys:
alchemiops can run on CPU but many of our test cases hit it's default max neighbor amounts

tests/test_fix_symmetry.py
```diff
@@ -34,7 +34,7 @@ def make_structure(name: str, repeats: int = REPEATS) -> Atoms:
         "fcc": lambda: bulk("Cu", "fcc", a=3.6),
         "hcp": lambda: bulk("Ti", "hcp", a=2.95, c=4.68),
         "diamond": lambda: bulk("Si", "diamond", a=5.43),
-        "bcc": lambda: bulk("Al", "bcc", a=2 / np.sqrt(3), cubic=True),
```
CompRhys, line 37: too dense lead to nbr explosion exceeding alchemiops padding heuristic

torch_sim/models/interface.py
```diff
@@ -371,3 +376,34 @@ def validate_model_outputs(  # noqa: C901, PLR0915
         raise ValueError(f"{fe_model_output['forces'].shape=} != (1, 3)")
     if stress_computed and fe_model_output["stress"].shape != (1, 3, 3):
         raise ValueError(f"{fe_model_output['stress'].shape=} != (1, 3, 3)")
+
+    # Translating one atom by a full lattice vector should not change outputs.
```
CompRhys, line 380: @alphalm4 This test seemingly passes with the upstream sevennet so I would guess that no additional changes are needed for the precision issue you mentioned at least at the 0.1 meV level (1e-4)

alphalm4: Great! I think the validation condition should be stricter in this [line](https://github.com/TorchSim/torch-sim/blob/a89ebe65fbb05be0a19aefccaf0d4d6b9e986061/torch_sim/models/interface.py#L384) like below,
since I found only using `+ lattice_vec` luckly PASSES even though NL things are not patched.

`shifted_state.positions[0] = shifted_state.positions[0] + 10 * lattice_vec`

I also checked that the current PR version, together with `nvalchemi-toolkit-ops==0.3.0`  and `vesin[torch]==0.5.3` passes all sevenn+torchsim tests with all four `torchsim_nl`s. Thanks for considering this.

Issue #453, opened 2026-02-12
Wrong unit of default b_tau in npt_langevin_init


chlwjd1234:
In `torch_sim.integrators.npt.npt_langevin_init`, the default value of `b_tau` is currently defined as

```python
if b_tau is None:
    b_tau = 1 / (1000 * dt)  # Default barostat time constant
```

However, b_tau is later used to construct the barostat mass as:

```python
cell_masses = (n_atoms_per_system + 1) * batch_kT * b_tau * b_tau
```

Since cell_masses should be in dimension of  (energy) · (time)² , 
it should be changed as

```python
b_tau = 1000 * dt
```

orionarcher:
@abhijeetgangan

thomasloux:
There is indeed something weird here. More investigation needs to be done on it.
If I remember well, the implementation of NPT Langevin corresponds exactly to press/langevin in LAMMPS. LAMMPS has all of its integrators using parameters with time units, but from memory this is not the first way NPT Langevin has been described, so we end up with some weird units like these ones.
Referring the original PR that exactly discuss b_tau (in their case, Pdamp) with time units and also at some point it was redefined as 1/time.
https://github.com/lammps/lammps/pull/3791

thomasloux:
I also point to LAMMPS docs which gives some more information about the units and how Pdamp should be setup: https://docs.lammps.org/fix_press_langevin.html

PR #521, opened 2026-03-24
Add integration tests checking the physical observables resulting from different the integrators

thomasloux:
## Summary

I've added tests to evaluate the correctness of integrators. I'm using physical_validation, used in gromacs test suite. The code is WIP, so that people can try it easily and update some parameters. In the future, the systems studied and the parameters will need to be fixed, or used the standard parameters. Note: this is important because the default parameters may be changed in an undesirable way.

## What to do next?
- Check that the chosen parameters are correct default one.
- Run for longer to have convergence of the integrator.
- Choose the optimal pressure to correctly evaluate NPT integrators.
- Change default parameters for some integrators, most importantly for NPT NH.

## Notes
- Parameters for NPT langevin are most likely wrong, as noted in #453.
- C-Rescale Anisotropic does not seem to work, crystal general tend to grow too large. I haven't checked the trajectories, but this is something I've realized during development. Maybe we should add a warning when a user chooses NPT anisotropic for now.
- Langevin parameters need to be more studied, the cell volume variation is far from correct.
- I should add how to estimate the isothermal_compressibility for c-rescale, which is easy and fast using QHA and phonopy.
- I've already tested changes in #520 and it works great. Actually I didn't even try without these changes, but I'm quite convinced the changes in #520 are correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,4 +56,3 @@ uv.lock
 
 # treemap html file
 docs/_static/torch-sim-pkg-treemap.html
-


thomasloux:
# In the previous commit:
- I added equations in the docs. It's easier to read then checking the papers each time.
- I fix Anisotropic NPT C-rescale: sign error (yes that's stupid). Now it works
- Fix some factor issues in NPT Langevin. @abhijeetgangan Can you have a look as I think you first implemented them?
- I've updated the default parameters, they work well with LJ Argon. I don't know for other MLIP, but that's the best I can do for now.

# What's next?
## Regarding integrators
- NPT Langevin does not provide correct pressure sampling (more precisely volume fluctuations). I'm trying to compare the implementation to LAAMPS but could not yet find the issue.
 
## Docs
- Docs need to be super explicit about units. Any parameters in integrator must be in the time (or inverse time) units so ps * MetalUnits.time. I highly suspect most people make the error of using `ts.integrate(timestep=dt)` with `dt=0.001` and then use some `factor*dt` for integrator argument, because timestep is automatically converted to torch-sim units (multiplied by MetalUnits.time in `integrate`).

thomasloux:
Ok I've managed to make all integrator pass:

Changelog:
NPT Langevin
- There was an error with `cell_positions` shape, initialized as a volume [n_system, 1, 1] and turn into [n_system, 3, 3] when added with `force_pressure` that was [n_system, 3, 3]
- Change the volume control to strain: otherwise the volume fluctuation was too low. The mean was correct but not fluctuation. Now with strain the correct thermodynamic system is sampled
Side note: these problems were in the code from the very first implementation.
- Added a version fully isotropic NPT Langevin Strain, and a version like LAAMPS that can control independently x,y,z. This make up 2 integrators for NPT Langevin, we can keep it like that or modify it to be one with an argument.

Others:
- Correct unwrap_positions that was wrong because test was wrong when evaluating modifying cells (ie. NPT simulation)

Note:
- Lennard Jones Model just got modified to be built from pair_potentials. That's great because it actually corrected a bug in Lennard Jones Model that was providing wrong forces: with the neighbor list it would correct the energy with a factor 1/2 because this matrix is symmetric, but this correction was not on forces, nor stress. That explains why it took me so long to make these tests, the model was not even right haha.
- Lennard Jones is crazy fast on gpu, the performance hardly depends on the number of atoms in the graph (which is the main bottleneck). So tests can run on gpu on relatively large cells. This is not true on cpu, that's why I force the test on GPU.
- On my h200: 1 min = 10_000 steps of most integrators except Nose Hoover (the substeps take time), so the entire integrator test is actually relatively fast to run, less than an hour with a GPU.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
 [project.optional-dependencies]
 test = [
     "torch-sim-atomistic[io,symmetry,vesin]",
+    "physical-validation>=1.0.5",
```
CompRhys, line 42: nit: This package seems unsupported

torch_sim/integrators/npt.py
```diff
@@ -153,268 +174,156 @@ def _npt_langevin_cell_beta(
...
 
     Returns:
-        NPTLangevinState: Updated state with new positions
+        Updated state with new positions
     """
-    # Calculate effective mass term by system
-    # Map masses to have batch dimension
-    M_2 = 2 * state.masses.unsqueeze(-1)  # shape: (n_atoms, 1)
-
-    # Calculate new cell length scale (cube root of volume for isotropic scaling)
```
CompRhys, line 348: the new beta coming in externalizes this to another method correct?

torch_sim/integrators/npt.py
```diff
@@ -532,48 +408,34 @@ def npt_langevin_init(
...
-    # Set default values if not provided
     if alpha is None:
-        alpha = 1.0 / (100 * dt)  # Default friction based on timestep
+        alpha = 1.0 / (5 * dt)
     if cell_alpha is None:
-        cell_alpha = alpha  # Use same friction for cell by default
+        cell_alpha = 1.0 / (30 * dt)
     if b_tau is None:
-        b_tau = 1 / (1000 * dt)  # Default barostat time constant
+        b_tau = 300 * dt
```
CompRhys, line 441: should this mean we should link to close #453?

torch_sim/integrators/npt.py
```diff
@@ -696,72 +565,510 @@ def npt_langevin_step(
...
-    # Scale cell uniformly in all dimensions
-    scaling = (V / V_0) ** (1.0 / dim)  # shape: (n_systems,)
 
-    # Apply scaling to reference cell to get new cell
-    new_cell = torch.zeros_like(state.cell)
-    for sys_idx in range(state.n_systems):
-        new_cell[sys_idx] = scaling[sys_idx] * state.reference_cell[sys_idx]
+# =============================================================================
+# NPT Langevin Strain integrator — isotropic logarithmic strain coordinate
+# =============================================================================
```
CompRhys, line 640: nit: can we remove CC headers and titles? Things like titles if used we need to keep consistent and so easier just not to have them imo

torch_sim/integrators/npt.py
thomasloux: Yes. By the way, the dimension units of these equations are sometimes a mystery to be honest.

thomasloux:
By the way, how do you want to handle the integrator names regarding whether they provide isotropic or anisotropic (length only or length+angle) ? NPT Crescale has these 3, now NPT Langevin supports isotropic and independant length.

thomasloux:
By the way, I need to correct the pytest tests. I was lazy running them, but now it only takes 15 minutes on a gpu, so that's fairly short. I'll add a doc with my jupyter notebook to show the plots. Note: there still pass on the notebook I use to make the tests.

pyproject.toml
thomasloux: I believe that's the last version on pypi and it's working on my computer. What is supposed to be unsupported?

torch_sim/integrators/npt.py
thomasloux: That's a bit weird but actually the random noise beta (both for cell or positions) is supposed to be the same one in equation of position and velocities: _npt_langevin_cell_position_step and _npt_langevin_cell_velocity_step for instance.

torch_sim/integrators/npt.py
thomasloux: done

PR #523, opened 2026-03-25
Upstream Metatomic Model

CompRhys:
better branch name for #510. Waiting on 0.0.2 on https://pypi.org/project/metatomic-torchsim/


Luthaf:
The code has been released as v0.1.1! https://pypi.org/project/metatomic-torchsim/

CompRhys:
Do you have a recommended model pt to use in testing here? I think the "pet-mad" keys no longer work with the new modularized packages. The ckpt from hugginface we used before needs more work to load into an `AtomsiticModel`

Luthaf:
I was going to point you toward upet, but it looks like you already found it! We also have a toy model in https://github.com/metatensor/lj-test/ that is intended to be used in tests.

pyproject.toml
```diff
@@ -52,8 +52,8 @@ io = ["ase>=3.26", "phonopy>=2.37.0", "pymatgen>=2025.6.14"]
 symmetry = ["moyopy>=0.7.8"]
 mace = ["mace-torch>=0.3.15"]
 mattersim = ["mattersim>=0.1.2"]
-metatomic = ["metatomic-torch>=0.1.3", "metatrain[pet]>=2025.12"]
-orb = ["orb-models>=0.6.0"]
+metatomic = ["metatomic-torchsim>=0.1.1", "metatomic-ase>=0.1.0", "upet>=0.2.0"]
```
Luthaf, line 55: The idea of metatomic/upet being separate packages is that some people will use other metatomic models that are not PET models, or not from the upet repository; so strictly speaking the upet dependency is not required here, except for testing purposes, but this is up to you!

CompRhys: I added the ase and upet to the same group for the purposes of having them installed for testing. Otherwise we would need to boost the number of extras groups 2x per model which seemed like a big step. 

Maybe there's a way to do it with https://peps.python.org/pep-0723/ like we do for examples but I think we can come back to that at a later point?

Luthaf: With the way this repository is setup, PEP 723 sounds like the best option yes! Otherwise we gave good success adding test-only dependencies to a `tox` environment specification. But this works for now!

CompRhys: the solution I would like doesn't yet exist there are no pep 723 aware test frameworks 🫠

PR #530, opened 2026-04-04
Add sum model interface

CompRhys:
Allows combination of models such as electrostatics or dispersion forces within TorchSim ecosystem.

Prompted by https://github.com/TorchSim/torch-sim/discussions/500

### DIFF
--- a/torch_sim/models/interface.py
+++ b/torch_sim/models/interface.py
@@ -234,10 +234,14 @@ class SumModel(ModelInterface):
         self._compute_stress = all(m.compute_stress for m in models)
         self._compute_forces = all(m.compute_forces for m in models)
 
+    def _children(self) -> list[ModelInterface]:
+        """Return child models with proper typing for static analysis."""
+        return list(self.models.children())  # type: ignore[return-value]
+
     @ModelInterface.compute_stress.setter
     def compute_stress(self, value: bool) -> None:  # noqa: FBT001
         """Propagate ``compute_stress`` to all child models that support it."""
-        for m in self.models:
+        for m in self._children():
             try:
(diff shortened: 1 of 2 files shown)


torch_sim/models/interface.py
```diff
@@ -171,6 +183,129 @@ def forward(self, state: SimState, **kwargs) -> dict[str, torch.Tensor]:
...
+                    raise
+        self._compute_forces = value
+
+    @property
+    def retain_graph(self) -> bool:
+        """Whether any child model retains the computation graph."""
+        return any(getattr(m, "retain_graph", False) for m in self._children())
+
+    @retain_graph.setter
+    def retain_graph(self, value: bool) -> None:
```
abhijeetgangan, line 269: The graph corresponding to the largest cutoff should be used. Then the nbrs for individual models can be masked out based on individual cutoff of the model. I don't think is logic is there currently?

abhijeetgangan:
Looks great!

torch_sim/models/interface.py
CompRhys: This retain graph is the torch autograd graph. I agree that there could be some benefit to doing the largest nbr-lists first and sharing but the trouble is that with the push to externalize to reduce maintenance we don't really want to have such a strict API contract to allow passing filtered edge idx in the forward pass of the model. I can't see a way to cache. I did think about putting a LRU cache on torchsim_nl but as the cutoff would change it wouldn't help in 99% of cases

abhijeetgangan: Ah got it. Alternative solution might be the dual cutoff neighborlist but we can do that in a separate PR.

PR #531, opened 2026-04-04
Add D3 Dispersion Model

CompRhys:
Adds Nvidia Alchemiops D3 in response to #500

PR #532, opened 2026-04-04
Add Electrostatics using extensible extras

CompRhys:
Leverage #464 to add electrostatics from Alchemiops

merge #530 and #531 into main first then merge #464 then finally this can be merged.

### DIFF
--- a/torch_sim/_duecredit.py
+++ b/torch_sim/_duecredit.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeVar
 
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+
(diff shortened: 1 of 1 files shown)


abhijeetgangan:
What is the use case for the standalone versions of these?

CompRhys:
> What is the use case for the standalone versions of these?

1. This is a first test of extensible extras working that occurred to me to add whilst doing the dispersion for #500.

2. The Alchemiops roadmap it is likely to add more and more fast primitives over time and people who want to use them will wrap things from alchemiops into their own packages proliferating lots of interfaces to the same code. Centralizing the interface here with SumModel (vs SumCalculator for ASE) just reduces the additional effort for people to compose there models. For example every package I follow had to cut a release on the 0.2.0 -> 0.3.0 alchemiops breaking change. It's much less maintainence to have just one TorchSim Model for each primitive that needs to be fixed in such instances.

PR #548, opened 2026-04-20
Add isolated atom into the interface test

CompRhys:
Add and isolated atom edge case. Get ahead of https://github.com/orbital-materials/orb-models/pull/158 such that we can release 0.6.0 in a more bug-free state


thomasloux (approved):
Could you add a comment above the line you removed the .squeeze() ? I didn't see the change immediately comparing with current Orb repo

PR #551, opened 2026-04-21
Cut v0.6.0 release

CompRhys:
Cut a new release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,81 @@
 
 ## Unreleased
 
+## v0.6.0
+
+This release is shaped by three major pushes:
+
+1. **External model posture.** Model integrations are increasingly maintained in the model's own package rather than in-tree. Orb (#456), SevenNet (#467), and Metatomic (#523) are migrated to upstream-maintained torch-sim interfaces, Nequix lands as a new model via the same pattern (#511), and testing utilities for this external-model posture are added (#433, released in v0.5.2).
+2. **Extensible extras framework.** A new scheme for attaching extra atom/system properties to `SimState` (#464) unlocks long-range physics on top of a common base: Electrostatics (#532), D3 Dispersion (#531), and a `SumModel` interface for composing force fields (#530, #537).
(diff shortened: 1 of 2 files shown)


CompRhys:
@Luthaf @haozeke this release will break because of your pinned deps :)

Luthaf:
Is TorchSim doing [semver](https://semver.org/) or not? I.e. does going from 0.5.x to 0.6.0 indicate breaking changes, or "a collection of major improvements"?

What the pin is trying to prevent is some user installing an old version of metatomic-torchsim with a recent version of torchsim that contains breaking changes and is no longer compatible.

So if you use semver we can update the pin to be <0.7, and otherwise we can maybe drop the pin. Although then I'm wondering how you want to deal with future breaking changes in torchsim's API?

CompRhys:
my understanding of semver is that until we issue a 1.0.0 all 0.x.y are allowed to be breaking because the API is still in flux. Once we get a 1.0.0 release then we would increment the first digit for breaking changes but as there is still flux in the public API (e.g. adding the "extras") then we're still a little way off doing that.

Luthaf:
Yes, sorry, I'm typically using the semver extension from [cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements) and did not make that clear:

> Versions are considered compatible if their left-most non-zero major/minor/patch component is the same. 

I find this helps with not having to pin to one specific version, and matches how people tend to think/use versions in practice.

------

This is mostly unrelated to this discussion though, my main question is how do you plan to handle breaking changes in TorchSim? I.e. what happen if you add a new required parameter to a function/rename a class?

I'm happy to remove the upper-version pin in metatomic-torchsim if you think this will work better overall!

CompRhys:
I don't think there's actually any way to avoid breaking CI if we have upper pins. Atleast whilst we wrap the interface in the package we would catch any issues with the latest release incompatibility with you guys before we cut a release and so on the assumption we continue to actively maintain torchsim in a responsive manner then I think it's fine to remove the upper pin on the basis that we're not planning to make many breaking changes even if we're still not close to 1.x.x and so not having the upper pin reduces the total number of PRs needed.

Luthaf:
metatomic-torchsim v0.1.2 is now on PyPI if you want to try CI again!

Issue #503, opened 2026-03-16
Update deformation gradient logic


CompRhys:
Initially we focussed on matching the Frechet filter in ASE. That implementation fixes the reference cell as the initial cell. This can lead to blowup of optimization if the cells change too much. 

Look at alternatives such as trust regions

PR #552, opened 2026-04-28
Fix bugs with constrained GeomOpt & FixSymmetry

PascalSalzbrenner:
### Bugs

I fixed three connected bugs causing high failure rates for constrained high-throughput geometry optimisations.

- In both FIRE & LBFGS, `cell_positions` were not updated after the cell was adjusted by constraints, which meant the next iteration started with stale positions. Now they are resynced after the adjustment.
- In LBFGS, the history vectors mixed two different coordinate frames because `prev_positions/forces` were stored in the un-adjusted coordinate frame. Now they are stored after the adjustment, consistently with the new positions.
- `FixSymmetry's` `max_cumulative_strain` clamp prevented full geometry optimisation. Speculatively, this was added to prevent strain blow-ups which the other bug fixes seem to have fixed. Removed `max_cumulative_strain` clamp & `reference_cells` which were used in the strain clamping. In my testing with heavily distorted structures, they all converge now.

### Testing

- Added constrained optimisation tests for `FixSymmetry`
- Add history vectors test for LBFGS optimiser
- Updated `FixSymmetry` tests to remove references to `max_cumulative_strain` clamp & `reference_cells`

## Checklist


CompRhys:
Thanks for the PR! I was out on holiday last week but will take a look in the next few days

CompRhys:
@janosh ~~Can you provide a repo for the issue motivating #447 as this removes a bunch of what I merged in there as a temp improvement.~~ I confused #447 with #450. Please ignore

torch_sim/constraints.py
```diff
@@ -932,20 +913,22 @@ def adjust_stress(self, state: SimState, stress: torch.Tensor) -> None:
             rots = self.rotations[ci].to(dtype=dtype)
             stress[si] = symmetrize_rank2(state.row_vector_cell[si], stress[si], rots)
 
-    def adjust_cell(self, state: SimState, cell: torch.Tensor) -> None:
+    def adjust_cell(
+        self, state: SimState, cell: torch.Tensor, max_delta_component: float = 0.25
```
CompRhys, line 917: note this is still essentially hard coded but just as a default rather than inline.

CompRhys:
Related: https://github.com/TorchSim/torch-sim/issues/503 -- we should work out how we want to better manage the reference cell in the frechet relaxation. Leaving it fixed the entire time despite matching ASE's implementation seems flawed for systems that relax substantially. 

LGTM

PR #560, opened 2026-05-13
Fix alchemiops dispersion/electrostatics nl issue.

CompRhys:
Unclear issue. Cutoff sometimes float sometimes Tensor inconsistently in codebase. ~~Partial fix is to increase minimum torch pin for 3.14 such that alchemiops works given torch.compile issues but that doesn't address the issues with vesin output ordering or the float/tensor arg standardization for cutoff.~~ Moved this to #566.

PR now just changes the tensor -> float cutoff and addresses some issues that were not previously tested for vesin nl paths.

### DIFF
--- a/examples/benchmarking/neighborlists.py
+++ b/examples/benchmarking/neighborlists.py
@@ -238,7 +238,7 @@ def _benchmark_backend(
     positions: torch.Tensor,
     cell: torch.Tensor,
     pbc: torch.Tensor,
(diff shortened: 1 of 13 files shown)


CompRhys:
No particular rush to merge, breaking change in that cutoff types are standardized towards float therefore probably needs more eyes.

orionarcher (approved):
Seems fine to me

PR #561, opened 2026-05-14
docs(optimize): document post-hoc convergence check

### LINKED ISSUE #542: Convergence and crystal relaxation
The suggested torch-sim interface to relax crystal is to use `ts.optimize`. This function returns the state relaxed, *whether or not the state has reached convergence*. I believe this function should probably return a `convergence` boolean tensor, or the docs should encourage the user to run:
```
convergence = convergence_fn(final_state)
```

lil-lon:
## Summary
Closes https://github.com/TorchSim/torch-sim/issues/542 with the docs-only approach as suggested.

* Added a Notes section to the `optimize` docstring describing the behavior where `final_state` is returned even when `max_steps` is reached, and documented the post-hoc check
* Mirrored the same explanation in the high-level tutorial's Geometry Optimization section

I also checked that the documents are properly rendered locally.

## Checklist

### DIFF
--- a/examples/tutorials/high_level_tutorial.py
+++ b/examples/tutorials/high_level_tutorial.py
@@ -347,6 +347,10 @@ that instead of taking `n_steps` and `temperature`, `optimize` takes a `converge
 that determines when the optimization is converged. By default, the `convergence_fn` will
 wait until the energy difference between steps is less than 1 meV.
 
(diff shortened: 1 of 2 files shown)


thomasloux (approved):
Much clearer docs on the convergence behaviour, thanks!

PR #562, opened 2026-05-14
Simplify to_constant_volume_bins

curtischong:
## Summary

By standardizing all of the different formats into a single datatype before running the algorithm, we no longer need isinstance checks, which simplifies the algorithm (since the algorithm is only handling one type of object)

## Checklist


torch_sim/autobatching.py
```diff
@@ -78,132 +79,63 @@ def to_constant_volume_bins(  # noqa: C901, PLR0915
...
-
-        # sort weights decreasingly
-        n_dcs = _rev_argsort_bins(vals)
-
-        weights = _get_bins(vals, n_dcs)
-        keys = _get_bins(keys, n_dcs)
-
-        bins = [[]] if is_tuple_list else [{}]
-    else:
-        weights = sorted(items, key=lambda x: -x)
```
curtischong, line 126: all of the above logic to get this reverse weights list is simplified to `entries.sort(key=lambda e: -e[0])` below

torch_sim/autobatching.py
```diff
@@ -78,132 +79,63 @@ def to_constant_volume_bins(  # noqa: C901, PLR0915
...
-    elif lower_bound >= upper_bound:
-        raise ValueError("lower_bound is greater or equal to upper_bound")
-
-    valid_ndcs = list(valid_ndcs)
+    is_dict = isinstance(items, dict)
+    is_tuple_list = not is_dict and (weight_pos is not None or key is not None)
+    if not is_dict and not is_tuple_list and isinstance(items[0], (tuple, list)):
+        raise ValueError("Must provide weight_pos or key for tuple list")
 
-    weights = _get_bins(weights, valid_ndcs)
```
curtischong, line 145: The above logic to get all of the valid indices is simplified to:

```
    if lower_bound is not None:
        entries = [e for e in entries if e[0] > lower_bound]
    if upper_bound is not None:
        entries = [e for e in entries if e[0] < upper_bound]
```

torch_sim/autobatching.py
```diff
@@ -78,132 +79,63 @@ def to_constant_volume_bins(  # noqa: C901, PLR0915
...
-
-    if isinstance(items, dict):
-        keys = _get_bins(keys, valid_ndcs)
```
curtischong, line 148: this is no longer needed since we operate on the new tuple objects, which automatically carries the keys in the entries list

torch_sim/autobatching.py
```diff
@@ -78,132 +79,63 @@ def to_constant_volume_bins(  # noqa: C901, PLR0915
...
+        entries = [e for e in entries if e[0] > lower_bound]
+    if upper_bound is not None:
+        entries = [e for e in entries if e[0] < upper_bound]
+
+    # Pack heaviest first, opening a new bin only when nothing fits.
+    entries.sort(key=lambda e: -e[0])
+    bin_entries: list[list[tuple[float, Any]]] = []
+    bin_sums: list[float] = []
+    for weight, payload in entries:
+        candidates = [i for i, s in enumerate(bin_sums) if s + weight <= max_volume]
```
curtischong: one thing I noticed that was missing in the original implementation (I didn't add it here). but I think we need to have an assert to ensure that all weights are <= max_volume? bc if there contains an item that exceeds max_volume, we can never fit it into a bin, even a new bin

torch_sim/autobatching.py
```diff
@@ -78,132 +79,67 @@ def to_constant_volume_bins(  # noqa: C901, PLR0915
...
+        entries = [e for e in entries if e[0] < upper_bound]
+
+    # Pack heaviest first, opening a new bin only when nothing fits.
+    entries.sort(key=lambda e: -e[0])
+    bin_entries: list[list[tuple[float, Any]]] = []
+    bin_sums: list[float] = []
+    for weight, payload in entries:
+        # get all bin indices that can fit this payload
+        candidate_bin_indices = [
+            i for i, s in enumerate(bin_sums) if s + weight <= max_volume
```
curtischong, line 124: one thing I noticed that was missing in the original implementation (I didn't add it here). but I think we need to have an assert to ensure that all weights are <= max_volume? bc if there contains an item that exceeds max_volume, we can never fit it into a bin, even a new bin

curtischong: maybe we add this line around 120?

```python
    # warn on oversized items — they'll still be placed in a bin by themselves,
    # but that bin will exceed max_volume
    for weight, payload in entries:
        if weight > max_volume:
            logger.warning(
                "item %r has weight %s > max_volume %s; placing in its own bin anyway",
                payload,
                weight,
                max_volume,
            )
```
Right now, we allow items to overflow the max bin size. some tests fail if we add this line and throw an exception

curtischong: I actually wanted this to be an exception, but I can see cases where people might want the weight to exceed the max volume so maybe it's not necessary to add this in.

PR #564, opened 2026-05-21
Detach state tensors after each step to prevent computation graph buildup

Panyalak:
## Summary
- Add `SimState.detach_()` method that detaches all tensor fields and extras from the computation graph in-place
- Call `state.detach_()` after each step in `integrate()` and `optimize()` loops, gated on `model.retain_graph`
- Prevents progressive slowdown in long MD optimization runs

## Test plan
- [x] All core tests pass (runners, optimizers, integrators, autobatching, state)
- [x] Run 100k-step MD to confirm slowdown is resolved with detach fix
Detach_vs_no-detach
The spike around step 70k probably came from the HPC system.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/torch_sim/runners.py
+++ b/torch_sim/runners.py
@@ -383,6 +383,9 @@ def integrate[T: SimState](  # noqa: C901
                 **integrator_kwargs,
             )
 
+            if not getattr(model, "retain_graph", False):
+                state.detach_()
+
             if trajectory_reporter:
                 trajectory_reporter.report(state, step, model=model)
 
@@ -696,6 +699,9 @@ def optimize[T: OptimState](  # noqa: C901, PLR0915
             if hasattr(state, "energy"):
                 last_energy = state.energy
             state = step_fn(state=state, model=model, **optimizer_kwargs)
+
+            if not getattr(model, "retain_graph", False):
+                state.detach_()
 
             if trajectory_reporter:
                 trajectory_reporter.report(
--- a/torch_sim/state.py
+++ b/torch_sim/state.py
@@ -409,7 +409,7 @@ class SimState:
         This severs the autograd graph that accumulates when model forward passes
         produce graph-attached tensors (energy, forces, stress) stored on the state.
         Without detaching, each simulation step builds an ever-growing computation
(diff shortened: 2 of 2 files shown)


orionarcher:
Do you have any evidence of slowdown or a before/after comparison? I don't think gradients are preserved by default right now

Panyalak:
I just added the plot before adding the detach function and after to the post. I performed MD simulations on 448 atoms structure at 1 fs time step for 100,000 steps and the trajectory is written every step.

I also think that the gradients are not preserved by default. Actually, I tried to log CPU and GPU usage to see what actually happened and found something interesting that the GPU usage becomes lower and lower over time. I am not sure why this happened. But the detach function that I used somehow solved this issue.
CPU_and_GPU_usage

CompRhys:
Currently only the pairwise model implements `retain_graph` do you see this also with any of the MLIP?

Panyalak:
Sorry for my late reply. I performed all the tests using MACE-MH-1 model, I have not tried using other models yet.

CompRhys:
I think probably the issue lies on MACE not detaching rather than in the code here. When we default to retrain graph false if missing it's because that is the expectation that it will be detached unless implementing a retain_graph variable.

Panyalak:
That makes sense. I will perform further tests using other models to make sure!

CompRhys:
https://github.com/TorchSim/torch-sim/blob/400c4072aa00d5edb1bf75598c569bd150d8a11d/torch_sim/models/interface.py#L472 So we have a pathway to check by default it they're detatched but I had the default on that function be False to soften a breaking change we made. It should be updated to be True now. I am not sure if MACE authors actually make use of the testing utils we provide here so maybe it wouldn't cause it to be caught upstream but this is a good reminded for me to swap this default.

Panyalak:
Thanks so much Rhys!

CompRhys:
Can you confirm that my diagnosis was correct and then I can close this out?

Panyalak:
I have just finished testing. The slowdown issue is now gone.

PR #566, opened 2026-05-22
Pin torch version due to 3.14 compile issues

CompRhys:
These issues were surfaced in an alchemiops error for highest deps. They are no longer apparent in the resolved pattern afaict but perhaps worth guarding.


orionarcher (approved):
Should create issue to unpin as a reminder perhaps

PR #571, opened 2026-06-05
Fix D3, Ewald, PME, DSF stress sign

teddykoker:
## Summary

 * ~~Shortly after D3 was added to torch-sim, [alchemi updated the sign convention of the D3 dispersion virial](https://github.com/NVIDIA/nvalchemi-toolkit-ops/pull/56), so now the D3 model in torchsim will produce the wrong sign.~~
 * Alchemi uses [negative strain derivate of energy for virial](https://github.com/NVIDIA/nvalchemi-toolkit-ops/blob/803008406345cf9d8b5c48a5afc2653d0f29d2bf/docs/userguide/about/conventions.md#virial), torch-sim uses positive for stress, so virial must be negated.
 * I added a test showing this compared to finite displacement, can be removed if preferred. 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "h5py>=3.15.1",
     "numpy>=1.26,<3; python_version < '3.13'",
     "numpy>=2.3.2,<3; python_version >= '3.13'",
-    "nvalchemi-toolkit-ops[torch]>=0.3.0",
+    "nvalchemi-toolkit-ops[torch]>=0.3.1",
     "tables>=3.11.1",
     "torch>=2",
     "tqdm>=4.67",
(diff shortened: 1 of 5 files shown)


abhijeetgangan (approved):

CompRhys:
Looks good, can you update the minimum pin to 0.3.1 which I think is the first release after that PR changing the convention. 

I am unsure why the lowest test here of dispersion doesn't fail if this sign convention changed as the lowest pin should be downloading 0.3.0 which should have the convention opposite to your fix here.

teddykoker:
Ah I misread the alchemi PR above. The D3 code *always* used the negative derivative for virial, the change from 0.3.0 to 0.3.1 was updating the DSF and LJ modules to match. This means D3 always had the incorrect sign, but DSF was once correct and now wrong after 0.3.1. I bumped up the pin to this version and also fixed DSF, Ewald, and PME.

CompRhys:
Okay I think then I had the wrong sign on everything thats for catching!

PR #572, opened 2026-06-11
Fix wrong default value in doc-string

t-reents:
Nothing fancy, but I finally got back to using `torch-sim` and noticed this small mismatch in the doc-string, which was introduced when the default value was updated in #404


CompRhys:
Thanks

Issue #575, opened 2026-06-13
Naive neighbor list is wrong for non-periodic systems carrying a cell


lil-lon:
## Summary

nvalchemi-toolkit-ops's batch naive neighbor list method (`batch_naive_neighbor_list`), which is torch_sim's default neighbor list, silently returns an incorrect neighbor list (and therefore wrong energies/forces) for non-periodic systems that still carry a non-zero cell.
Issue in the upstream repository: https://github.com/NVIDIA/nvalchemi-toolkit-ops/issues/104

It shows up in common setups:

- a molecule (`pbc=[F,F,F]`) carrying a dummy cell (e.g. `ase.Atoms.get_cell(complete=True)` returns a 1 Å identity)
- a slab (`pbc=[T,T,F]`) whose atoms straddle the non-periodic z boundary

It is silent: the neighbor count can look right, but actually distances/energies etc are wrong.
Downstream, e.g. a MACE single point on an isolated molecule comes back with the wrong energy.

Note that only the default (naive) backend is affected.
The other backends, such as `vesin_nl` and alchemiops `cell_list`, return correct values.

### Code sample

Minimal reproduction: a `pbc=False` H2O molecule carrying a 1 Å dummy cell, run through `ts.static` with the MACE-MPA-0 model while swapping the neighbor list backend, and compared against a plain ASE MACE single point.

```python
import torch
from ase.build import molecule
from mace.calculators import mace_mp

import torch_sim as ts
from torch_sim.models.mace import MaceModel
from torch_sim.neighbors import (
    alchemiops_nl_cell_list,
    alchemiops_nl_n2,
    vesin_nl,
)

device = "cuda"
dtype = torch.float64
MODEL = "medium-mpa-0"

# Non-periodic H2O carrying a 1 A dummy cell. pbc stays False, so the energy
# must be cell-independent and match a plain ASE MACE single point.
atoms = molecule("H2O")
atoms.set_cell([1.0, 1.0, 1.0])

# Reference: the standard ASE MACE calculator, which uses its own
# neighbor list and is unaffected by the dummy cell.
ref_atoms = atoms.copy()
ref_atoms.calc = mace_mp(
    model=MODEL, dispersion=False, device=device, default_dtype="float64"
)
ref_energy = ref_atoms.get_potential_energy()

raw_model = mace_mp(
    model=MODEL,
    dispersion=False,
    device=device,
    default_dtype="float64",
    return_raw_model=True,
)


[...]

PR #576, opened 2026-06-13
fix(neighbors): zero dummy cell for non-periodic systems in alchemiops naive NL

lil-lon:
## Summary
Related to https://github.com/TorchSim/torch-sim/issues/575.

## Fix
The alchemiops naive backend returns a wrong neighbor list for non-periodic systems carrying a non-zero (dummy) cell, because the kernel wraps on every axis ignoring pbc.

## Test
Extended `test_neighbor_list_implementations` with a `molecule_dummy_cell` axis: non-periodic molecules carrying a dummy cell must still match the ASE reference across all backends.
Confirmed that newly added tests fail without this patch.

## Note
Slabs / partial pbc (e.g. `[T, T, F]`) are not covered and still need the upstream fix (NVIDIA/nvalchemi-toolkit-ops#104).
So, even after this PR is merged, we cannot close the above issue, but this PR would resolve the major path where molecule is called with dummy cell.

## Checklist

### DIFF
--- a/tests/test_neighbors.py
+++ b/tests/test_neighbors.py
@@ -286,11 +286,13 @@ def test_neighbor_list_invariant_under_lattice_image_shifts(
     _nl_backends_x_cutoffs(),
 )
 @pytest.mark.parametrize("self_interaction", [True, False])
+@pytest.mark.parametrize("molecule_dummy_cell", [0.0, 1.0])
 def test_neighbor_list_implementations(
     *,
     nl_implementation: Callable[..., tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
     cutoff: float,
     self_interaction: bool,
+    molecule_dummy_cell: float,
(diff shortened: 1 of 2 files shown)


CompRhys:
Thanks! when the upstream issue is resolved will you make another PR for the slab case?

lil-lon:
Thank you too for the review and merge!

> when the upstream issue is resolved will you make another PR for the slab case?

Yes. Once the upstream issue is fixed, the kernel will handle both the slab case and the non-periodic molecule case (the case I patched in this PR) natively.
I plan to open a follow-up PR to bump the nvalchemi-toolkit-ops dependency and add unit tests covering slab-like partial-pbc systems. (Also, it's possible to drop this patch after the fix if you want.)

This PR only patched the molecule case since it was easy to fix; the slab fix is more complex, so I thought it was better to wait for the upstream fix.

PR #577, opened 2026-06-16
fix(deps): pin phono3py>=4.0.0 in mattersim extra

lil-lon:
## Summary

`uv sync` fails both locally and in CI (e.g. https://github.com/TorchSim/torch-sim/actions/runs/27552957567/job/81444030054) because uv resolves `phono3py==3.2.0`, whose sdist is unbuildable.
Since `mattersim` doesn't pin `phono3py`, this PR adds `phono3py>=4.0.0` (simply picked the available latest major version) to the `mattersim` extra so uv picks a buildable wheel.

## Checklist


CompRhys:
thanks!

Issue #580, opened 2026-06-25
Missing device in `FixCom` / `get_centers_of_mass`


sklenard:
Some `torch.zeros()` tensors are not placed on the correct device when adding a `FixCom`  constraint or when calling `get_centers_of_mass()` leading to a `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!` when running on GPU. This is somehow related to issue #371.

Minimal reproduction:

```python
import torch
from ase.build import bulk

import torch_sim as ts
from torch_sim.constraints import FixCom
from torch_sim.models.lennard_jones import LennardJonesModel
from torch_sim.transforms import get_centers_of_mass
from torch_sim.units import MetalUnits

n_steps = 2000
dt = 0.001            # ps
temperature = 1000.0  # K
dtype = torch.float64
device = "cuda"

atoms = bulk("Ar", "fcc", a=5.26, cubic=True) * (2, 2, 2)  # 32 atoms
atoms.rattle(stdev=0.05, seed=0)

calc = LennardJonesModel(
    sigma=3.405,
    epsilon=0.0104,
    dtype=dtype,
    device=device,
    compute_forces=True,
)

# --- 3. Build SimState from atoms, attach FixCom, record initial COM ---
state = ts.initialize_state(atoms, device=device, dtype=dtype)
state.constraints = [FixCom([0])]

initial_com = get_centers_of_mass(
    state.positions,
    state.masses,
    state.system_idx,
    state.n_systems,
)

state = ts.integrate(
    system=state,
    model=calc,
    n_steps=n_steps,
    timestep=dt,
    temperature=temperature,
    integrator=ts.Integrator.nvt_nose_hoover,
    pbar=True,
)

final_com = get_centers_of_mass(
    state.positions,
    state.masses,
    state.system_idx,
    state.n_systems,
)

print("Initial COM:", initial_com.tolist())
print("Final   COM:", final_com.tolist())
print("COM drift  :", (final_com - initial_com).abs().max().item())
```

CompRhys:
Thanks for the report! would you be willing to submit a PR to fix this?

sklenard:
Sure !

PR #583, opened 2026-07-01
Fix missing device in FixCom and get_centers_of_mass()

sklenard:
## Summary

Related to #580.
Add `device` to all `torch.zeros()` calls in `FixCom` and in `get_center_of_mass`, fixing a crash on CUDA when using the `FixCom` constraint.

## Checklist

### DIFF
--- a/torch_sim/transforms.py
+++ b/torch_sim/transforms.py
@@ -1620,13 +1620,15 @@ def get_centers_of_mass(
         torch.Tensor: A tensor of shape (n_structures, 3) containing
             the center of mass coordinates for each structure.
     """
-    coms = torch.zeros((n_systems, 3), dtype=positions.dtype).scatter_add_(
+    coms = torch.zeros(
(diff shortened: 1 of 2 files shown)


torch_sim/transforms.py
curtischong, line 1: I feel like linked_cell may also need a `device=` arg. it'd be great if we can add it in this PR!

curtischong (approved):
Overall it looks good. I checked all the torch.zeros calls in both files and aside from the `linked_cell` one, you seem to have got them all

torch_sim/transforms.py
curtischong: https://github.com/TorchSim/torch-sim/blob/main/torch_sim/transforms.py#L891 and https://github.com/TorchSim/torch-sim/blob/main/torch_sim/transforms.py#L971

sklenard: Good catch, I'll add that fix now!

CompRhys:
lgtm, set to auto merge

PR #587, opened 2026-07-03
Convert ts.integrate param units

### LINKED ISSUE #579: Batched MD with SevenNet+D3 can produce NaN trajectories while ASE remains finite
## Summary

I am seeing reproducible NaNs during TorchSim batched MD with SevenNet v0.13.0 + D3(BJ)-PBE. The same structures and MD settings remain finite when run through the ASE SevenNet+D3 path.

The TorchSim call follows the documented batched-MD style:

```python
ts.integrate(system=list_of_atoms, ...)
```

where `list_of_atoms` contains 16 independent ASE `Atoms` objects.

## System

- 16 independent amorphous `Na16Ta16Cl96` structures
- 128 atoms per structure
- Initial structures look reasonable; initial nearest-neighbor distances are about 2.26 Angstrom
- SevenNet OMNI checkpoint with `modal="omat24"`
- D3(BJ)-PBE enabled

## MD settings

- Ensemble: fixed-cell NVT
- Integrator: Nose-Hoover chain
- Temperature: 300 K
- Timestep: 2 fs
- Thermostat damping: 200 fs
- Warmup: 200 steps
- Measured segment: 1000 steps

## Environment

Primary test:

- GPU: NVIDIA RTX 5090
- `sevenn 0.13.0`
- `torch 2.11.0+cu128`
- `torch-sim-atomistic 0.6.0`
- `ase 3.28.0`

Repeat test:

- GPU: NVIDIA RTX 4090
- Same package versions: `sevenn 0.13.0`, `torch 2.11.0+cu128`, `torch-sim-atomistic 0.6.0`, `ase 3.28.0`

## Observed behavior

On the RTX 5090 machine, for the same batch of 16 structures:

- TorchSim + `cueq`: MD completes, but some batch elements end with NaN energy/temperature.
- TorchSim + `flash`: MD completes, but some batch elements end with NaN energy/temperature.
- ASE + SevenNet+D3 + `flash`: all 16 structures remain finite under the same structures, potential, and MD settings.

On the RTX 4090 machine, repeating TorchSim + `cueq` also produces NaN energy/temperature in some batch elements. This suggests the issue is not specific to a single RTX 5090/Blackwell machine.

## Static single-point comparison

For the same structure, TorchSim SevenNet+D3 and ASE SevenNet+D3 agree closely in static single-point calculations:

- Energy difference: about `0.000679 meV/atom`
- Max force-component difference: about `1.62e-5 eV/Angstrom`
- Max stress-component difference: about `1.13e-8 eV/Angstrom^3`

So the problem appears more likely to be in the TorchSim MD state update / batched integration path than in the static SevenNet+D3 energy/force evaluation itself.

## Expected behavior

With the same initial structures, potential, and fixed-cell NVT settings, TorchSim batched MD should keep energy, temperature, and coordinates finite. If one batch element becomes numerically unstable, it would also be helpful to have a diagnostic exception or warning rather than silently returning a final state with partial-batch NaNs.

## Actual behavior

TorchSim batched MD completes and returns results, but some batch elements end with NaN energy/temperature. The ASE MD control remains finite under the same conditions.

I can provide the benchmark config, structures, or a smaller reproducer if that would be useful.

curtischong:
## Summary

We currently are not converting the parameters in ts.integrate (they are supposed to accept normal units)
Note: this is in contrast to the internal api (direct integrator functions like `ts.npt_nose_hoover_isotropic_step` where they expect the units to already be converted).

This PR goes through all the params and adds a type annotation to these args in the integrator functions to properly convert the units.

Note: additional kwargs like `a = kwargs.get("momenta")` in https://github.com/TorchSim/torch-sim/blob/main/torch_sim/integrators/nvt.py#L344 and https://github.com/TorchSim/torch-sim/blob/main/torch_sim/integrators/npt.py#L1733 do not have annotated conversion param like the rest. I think this is fine since users are not manually initing the momenta for individual atoms in the ts.integrate api. - they're most likely loaded in from another simulation (so I think they'll have the proper units)

This PR is to fix the issue outlined in https://github.com/TorchSim/torch-sim/issues/579 where a user was changing the settings for a simulation and since we didn't properly convert the units to internal MetalUnits, their simulation was unstable. This change fixes the issue noticed in their md workflow

## Checklist

### DIFF
--- a/torch_sim/integrators/md.py
+++ b/torch_sim/integrators/md.py
@@ -4,6 +4,7 @@ import logging
 import warnings
 from collections.abc import Callable
 from dataclasses import dataclass
+from typing import Annotated
 
 import torch
 
@@ -15,6 +16,13 @@ from torch_sim.units import MetalUnits
 
 
 logger = logging.getLogger(__name__)
+
+
+# The metadata is the factor that converts the kwarg to internal units
+TimeArg = Annotated[float | torch.Tensor | None, MetalUnits.time]
+InverseTimeArg = Annotated[float | torch.Tensor | None, 1 / MetalUnits.time]
+PressureArg = Annotated[float | torch.Tensor, MetalUnits.pressure]
+InversePressureArg = Annotated[float | torch.Tensor | None, 1 / MetalUnits.pressure]
 
 
 @dataclass(kw_only=True)
--- a/torch_sim/integrators/md.py
+++ b/torch_sim/integrators/md.py
@@ -19,10 +19,12 @@ logger = logging.getLogger(__name__)
 
 
 # The metadata is the factor that converts the kwarg to internal units
-TimeArg = Annotated[float | torch.Tensor | None, MetalUnits.time]
-InverseTimeArg = Annotated[float | torch.Tensor | None, 1 / MetalUnits.time]
+TemperatureArg = Annotated[float | torch.Tensor, MetalUnits.temperature]
+EnergyArg = Annotated[float | torch.Tensor, MetalUnits.energy]
+TimeArg = Annotated[float | torch.Tensor, MetalUnits.time]
+InverseTimeArg = Annotated[float | torch.Tensor, 1 / MetalUnits.time]
 PressureArg = Annotated[float | torch.Tensor, MetalUnits.pressure]
-InversePressureArg = Annotated[float | torch.Tensor | None, 1 / MetalUnits.pressure]
+InversePressureArg = Annotated[float | torch.Tensor, 1 / MetalUnits.pressure]
 
 
 @dataclass(kw_only=True)
@@ -111,7 +113,7 @@ def initialize_momenta(
     positions: torch.Tensor,
     masses: torch.Tensor,
     system_idx: torch.Tensor,
-    kT: float | torch.Tensor,
+    kT: EnergyArg,
     generator: torch.Generator | None = None,
 ) -> torch.Tensor:
     """Initialize particle momenta based on temperature.
@@ -164,7 +166,7 @@ def initialize_momenta(
     )
 
 
-def momentum_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
+def momentum_step[T: MDState](state: T, dt: TimeArg) -> T:
     """Update particle momenta using current forces.
 
     This function performs the momentum update step of velocity Verlet integration
@@ -185,7 +187,7 @@ def momentum_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
     return state
 
 
-def position_step[T: MDState](state: T, dt: float | torch.Tensor) -> T:
+def position_step[T: MDState](state: T, dt: TimeArg) -> T:
     """Update particle positions using current velocities.
 
     This function performs the position update step of velocity Verlet integration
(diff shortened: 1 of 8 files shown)


torch_sim/integrators/md.py
```diff
@@ -17,6 +18,13 @@
 logger = logging.getLogger(__name__)
 
 
+# The metadata is the factor that converts the kwarg to internal units
+TimeArg = Annotated[float | torch.Tensor | None, MetalUnits.time]
```
CompRhys: `1 / MetalUnits.time` are these helpful as type hints? it's just a float and it doesn't necessarily add much imo?

torch_sim/runners.py
```diff
@@ -320,6 +320,16 @@ def integrate[T: SimState](  # noqa: C901
             f"integrator must be key from Integrator or a tuple of "
             f"(init_func, step_func), got {type(integrator)}"
         )
+
+    # Like `timestep` above, unit-carrying kwargs (e.g. `tau`, `gamma`,
+    # `external_pressure`) must be converted to internal units per
+    # INTEGRATOR_UNIT_KWARGS
+    init_kwargs = dict(init_kwargs or {})
```
CompRhys: nit the dict call isn't needed?

CompRhys:
The alternative here is simply to delete the concept of the unit convertors within TS that this PR attempts to fix and then leave it up to the user to ensure that they're numerically consistent between their model and integration parameters. @abhijeetgangan would you be opposed to that as a lower maintenance and LLOC solution?

torch_sim/integrators/md.py
curtischong: Actually, this is how this annotated system works. inside the `_collect_unit_kwargs` above, we read the metadata in `for meta in getattr(hint, "__metadata__", ()):`

By specifying the unit conversion alongside the typehint, it's 1) obvious what the unit conversion is, and 2) makes it simple to find the proper conversion float since it's right alongside the type definition.

i.e. in this setup, we don't need to define a new dictionary that maps each type to its conversion coefficient

curtischong:
I think we need to keep the metalunits bc it does simplify the formulas. here's an example branch where I asked an ai to remove the units https://github.com/TorchSim/torch-sim/compare/main...curtischong:torch-sim:remove-metal-units.

Maybe there's a better way though

curtischong:
I think I'm just hesitant because I'm not familiar with all the units. The diff doesn't look too bad and as long as we tell users the expected units for each param we should be fine (they won't really see all the internal unit conversions anyway). I'll put up a PR so we can run tests on the branch

torch_sim/runners.py
curtischong: Good observation. We modify the dict below `kwargs[key] = kwargs[key] * meta.factor` so this is more like a copy so the params aren't changed. I think this is fine because if ppl are using the same parameter dict across diff runs, we wouldn't one run to modify the input dict.

init_kwargs = {} if init_kwargs is None else init_kwargs.copy()

I'll add a comment

CompRhys:
@sek1ro-yuzzz could you test your SevenNet NaN issue on this branch please?

curtischong:
I tested this PR with this script and it passes on this branch, but fails on main

[repro_579.py](https://github.com/user-attachments/files/30362304/repro_579.py)

CompRhys:
thanks @curtischong!

PR #589, opened 2026-07-04
fix(autobatching): recognise Warp OOM in memory estimation and free cache before the real run

niklashoelter:
## Problem

`InFlightAutoBatcher` / `BinningAutoBatcher` estimate a safe batch size by replaying the model forward on geometrically-growing batches until it OOMs, then backing off. Two defects on that path make models whose neighbor lists run on **Warp / nvalchemiops** (e.g. ORB v3) crash instead of backing off:

1. **The Warp OOM isn't recognised.** `determine_max_batch_size` only matched the substring `"CUDA out of memory"`. Warp's allocator raises `"Failed to allocate <n> bytes on device '…'"`, which doesn't match, so the OOM propagates and kills the run *during memory estimation*. In addition, the re-`raise` sat **inside** the `for msg in oom_error_message` loop, so passing a broadened list of signatures never worked — the first non-matching entry re-raised before later ones were checked.

2. **PyTorch's caching allocator starves the external Warp pool after estimation.** Once (1) is fixed, estimation survives but the crash simply *moves* to the first real optimization forward pass, failing on a tiny (~2.5 KB) Warp allocation. The estimation forwards leave PyTorch's caching allocator holding almost all of device memory, and Warp allocates from a *separate* `cudaMallocAsync` pool that then has nothing left.

Concretely, ORB v3 conservative on a 4272-molecule batch reproducibly died with `RuntimeError: Failed to allocate 2556 bytes on device 'cuda:N'`.

## Fix (`torch_sim/autobatching.py`)

- Add `DEFAULT_OOM_ERROR_MESSAGES = ("CUDA out of memory", "Failed to allocate")` and use it as the default `oom_error_message` for `determine_max_batch_size`, `InFlightAutoBatcher`, and `BinningAutoBatcher`.
- Re-raise only after checking **all** signatures (`if any(msg in exc_str …): return safe` then `raise`), so a multi-signature list actually works.
- Release memory at the end of `estimate_max_memory_scaler` (shared by both batchers) with `torch.cuda.synchronize()` + `torch.cuda.empty_cache()`, guarded by `torch.cuda.is_available()`, so the real run isn't starved.

## Tests

Adds regression tests in `tests/test_autobatching.py`: both PyTorch and Warp OOM wordings are recognised, and a genuine non-OOM error still propagates. `pytest tests/test_autobatching.py` → 38 passed; ruff clean.

## Validation

ORB v3 conservative over a 4272-structure library on a dedicated GPU: **4/4 full runs complete `4272/4272`** with the fix; the same code with the memory release disabled crashes with `Failed to allocate 2556 bytes`.

### DIFF
--- a/tests/test_autobatching.py
+++ b/tests/test_autobatching.py
@@ -488,6 +488,63 @@ def test_determine_max_batch_size_fibonacci(
     assert max_size == 2
 
 
+@pytest.mark.parametrize(
+    "oom_message",
+    [
+        "CUDA out of memory. Tried to allocate 20.00 MiB",
+        # Warp / nvalchemiops allocator (used by ORB v3 neighbor lists) phrases
+        # OOM differently and must still be recognised by the default matcher.
+        "Failed to allocate 2556 bytes on device 'cuda:0'",
+    ],
+)
+def test_determine_max_batch_size_recognises_oom_variants(
+    si_sim_state: ts.SimState,
+    lj_model: LennardJonesModel,
+    monkeypatch: pytest.MonkeyPatch,
+    oom_message: str,
+) -> None:
+    """OOM is detected for both PyTorch and Warp-style allocator messages.
+
+    Regression test: the default ``oom_error_message`` must cover the Warp
+    allocator wording, and a non-matching first entry in the message list must
+    not short-circuit the check before later entries are compared.
+    """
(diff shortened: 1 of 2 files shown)


CompRhys:
LGTM

PR #591, opened 2026-07-18
fix(neighbors): partial pbc and imports

### LINKED ISSUE #575: Naive neighbor list is wrong for non-periodic systems carrying a cell

lil-lon:
## Summary

Closes #575, fixes a partial-pbc bug in vesin_nl, and updates import paths for current releases.

### Bug fix
* Require nvalchemi-toolkit-ops >= 0.4.0 (https://github.com/NVIDIA/nvalchemi-toolkit-ops/releases/tag/v0.4.0), which fixes the naive kernel wrapping positions on non-periodic axes, and drop the workaround for non-periodic systems
  * The bump also picks up upstream DFT-D3 and electrostatics updates, and I confirmed that the existing dispersion and electrostatics tests pass with 0.4.0
  * Note: 0.4.0 deprecates the direct-output flags (compute_forces / compute_virial) used in EwaldModel and PMEModel, so their calls now emit a DeprecationWarning; this issue is left as a follow-up
* Fix `vesin_nl` collapsing per-axis pbc into a single bool, and require vesin >= 0.5.8 for the upstream partial-pbc fixes (https://github.com/Luthaf/vesin/releases/tag/v0.5.8)
  * 0.6.0 requires cut_off argument as a key-work argument (https://github.com/Luthaf/vesin/blob/v0.6.0/python/vesin/vesin/_neighbors.py#L73-L81), so fixed it

### Import path fix and refactoring
* Support the vesin-torch 0.6 module move (`vesin.torch` -> `vesin_torch`), keeping the old path as a fallback
* Import nvalchemiops from its public path (`nvalchemiops.torch.neighbors`) and drop the legacy fallback

### Test
* Add partial-pbc slab tests and fold the dummy-cell molecule variants into the fixture for simplicity and test efficiency
* codecov fails because no tests verify the package import behavior. If needed, I'd add corresponding tests

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "h5py>=3.15.1",
     "numpy>=1.26,<3; python_version < '3.13'",
     "numpy>=2.3.2,<3; python_version >= '3.13'",
-    "nvalchemi-toolkit-ops[torch]>=0.3.1",
+    "nvalchemi-toolkit-ops[torch]>=0.4.0",
     "tables>=3.11.1",
     "torch>=2",
     "tqdm>=4.67",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ test = [
     "pytest>=8",
     "spglib>=2.6",
 ]
-vesin = ["vesin[torch]>=0.5.3"]
+vesin = ["vesin[torch]>=0.5.8"]
 io = ["ase>=3.26", "phonopy>=3.0.0", "pymatgen>=2026.3.23"]
 symmetry = ["moyopy>=0.7.8"]
 mace = ["mace-torch>=0.3.16"]
--- a/tests/test_neighbors.py
+++ b/tests/test_neighbors.py
@@ -6,7 +6,7 @@ import numpy as np
 import pytest
 import torch
 from ase import Atoms
-from ase.build import bulk, molecule
+from ase.build import bulk, fcc111, molecule
 from ase.neighborlist import neighbor_list
 
 from tests.conftest import DEVICE, DTYPE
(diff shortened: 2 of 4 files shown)


CompRhys:
Thanks for the follow on PR, I am currently on holiday and will review/merge next week.

CompRhys (approved):
lgtm

PR #593, opened 2026-07-23
Remove metal units

curtischong:
## Summary

As Rhys pointed out, it's not 100% necessary and the entire concept was originally ported over from ASE. Having values passed in two kinds of units (metal vs non-metal) was giving us issues because some interfaces had inconsistent units https://github.com/TorchSim/torch-sim/issues/579

## Checklist


CompRhys:
other PR is smaller footprint and therefore I think that it's probably better solution.

Issue #584, opened 2026-07-02
Per-system pbc: batching periodic and non-periodic systems together


lil-lon:
In the review of #320 there was a short discussion about making `pbc` a per-system attribute so that systems with different pbc can be mixed in one batch (https://github.com/TorchSim/torch-sim/pull/320#issuecomment-3500245561).
It was deferred to a separate issue at the time, but I don't think that issue was ever opened, so raising it here.

I think this would be a useful addition. One concern in that discussion was that mixed pbc may complicate the neighbor list calculations (https://github.com/TorchSim/torch-sim/pull/320#issuecomment-3508563598), but the current neighbor list API already supports per-system pbc (e.g. `batch_naive_neighbor_list` in nvalchemi-toolkit-ops takes `pbc` of shape `(num_systems, 3)`: https://github.com/NVIDIA/nvalchemi-toolkit-ops/blob/2b7c3c3adfb1ca84b886eecbf14bc60ff6ba1dc2/nvalchemiops/torch/neighbors/batch_naive.py#L765-L767), and I think other methods support similar APIs, so it can be handled without making the neighbor list logic so complicated.

What do you think? If we agree on a direction I'd be happy to work on a PR.

CompRhys:
I would be happy to see this PR, I think it's unlikely that the native torch NL we have would support but I also don't believe that they're likely to be used in most cases. Primary question would be how to add the check to the NL that do not support so they raise errors correctly.

lil-lon:
Thanks! I'd like to start working on a PR. Sharing the rough direction first:

- Make `SimState.pbc` a per-system `(n_systems, 3)` tensor. The constructor keeps accepting `bool` / `list[bool]` / `(3,)` and broadcasts, so the main breaking change is the shape of `state.pbc` for code reading it directly.
- Neighbor lists turned out to need no changes: all backends including the native torch NL already support per-system pbc, and existing tests already cover mixed batches. So I don't think NL-level error checks are needed.
- Core MD and optimization are pbc-agnostic and work with mixed batches as-is. Features that assume periodicity need case-by-case handling (details in the PR): Ewald/PME would zero-fill non-periodic systems per system, matching their current all-non-periodic behavior, and workflows requiring full periodicity (phonons, elastic) get an entry-point check with a clear error.
[...]

PR #594, opened 2026-07-25
feat(state): make pbc a per-system attribute

lil-lon:
## Summary

* Make `SimState.pbc` a per-system attribute of shape `(n_systems, 3)`; a bool, `list[bool]`, or `(3,)` tensor is broadcast to all systems. Moving it from `_global_attributes` to `_system_attributes` lets slicing, splitting, popping, and concatenation handle it like `cell`.
* `atoms_to_state` / `structures_to_state` accept inputs with different pbc instead of raising, and `state_to_atoms` / `state_to_structures` / trajectory files store each system's own pbc row.
* Ewald and PME models evaluate the periodic systems of a mixed batch and zero-fill systems with no periodic axis.
* Add `require_full_pbc`; elastic tensor calculation now rejects systems that are not fully periodic.
* The orb / NequIP / metatomic / fairchem wrappers convert pbc back to the legacy `(3,)` row on a shallow copy (`_to_legacy_pbc_state`) before delegating, because those integrations broadcast `state.pbc` themselves assuming the old shape; mixed-pbc batches raise a clear error there.
* Declare `pbc` after `system_idx` so `__setattr__` can broadcast a `(3,)` row already during `__init__`, leaving a single broadcast site (`__post_init__` only validates the shape). Simplify pbc handling in `autobatching` and `io` now that pbc is always a tensor.

Related issue: #584 

## Compatibility

Breaking:

* External code reading state.pbc as a global (3,) tensor must index it per system. The bundled orb / NequIP / metatomic / fairchem wrappers already adjusted this.
* `get_attrs_for_scope`: `pbc` moved from `"global"` to `"per-system"`.

Unchanged:

* bool / `list[bool]` / `(3,)` inputs are still accepted and broadcast to all systems.
* Trajectory files still store a `(3,)` pbc array; existing files stay readable.

## Discussion points

* `_to_legacy_pbc_state` hands third-party forward() a SimState whose pbc temporarily violates the `(n_systems, 3)` invariant. The preferred fix is upstream support for per-system pbc in the integrations; until then the shim keeps the current orb / NequIP / metatomic / fairchem releases working unchanged, since they only read plain attributes off the state.
* `state.pbc = <1-D value>` after construction broadcasts any length without validation (`__post_init__` runs only at construction). Missing validation is pre-existing, but should `__setattr__` enforce a length-3 row?

## Checklist

### DIFF
--- a/examples/scripts/7_others.py
+++ b/examples/scripts/7_others.py
@@ -56,9 +56,6 @@ n_atoms = state.n_atoms
 cutoff = 4.0
 self_interaction = False
 
-# Ensure pbc has the correct shape [n_systems, 3]
-pbc_tensor = torch.tensor(pbc).repeat(state.n_systems, 1)
-
 log.info(f"Batched system with {state.n_systems} structures:")
 for i, atoms in enumerate(atoms_list):
     log.info(f"  Structure {i}: {atoms.get_chemical_formula()} ({len(atoms)} atoms)")
@@ -66,7 +63,7 @@ for i, atoms in enumerate(atoms_list):
 # Method 1: Linked cell neighbor list (efficient for large systems)
 log.info("Calculating neighbor lists with linked cell method...")
 mapping, mapping_system, shifts_idx = torch_nl_linked_cell(
-    pos, cell, pbc_tensor, cutoff, system_idx, self_interaction
+    pos, cell, pbc, cutoff, system_idx, self_interaction
 )
 cell_shifts = transforms.compute_cell_shifts(cell, shifts_idx, mapping_system)
 dds = transforms.compute_distances_with_cell_shifts(pos, mapping, cell_shifts)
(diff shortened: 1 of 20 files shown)


CompRhys:
is this ready to merge?

lil-lon:
Yes, ready for review.

One thing I'd like your eyes on: some potentials don't support mixed pbc, so they won't benefit from this change until upstream adds support. To keep them working, `_to_legacy_pbc_state` feeds them a state whose pbc temporarily violates the `(n_systems, 3)` invariant, which I'm slightly uneasy about (see Discussion points; mixed-pbc batches raise a ValueError there).
For potentials that support mixed pbc, I'm also considering a test that compares a mixed-pbc batch against per-system results.

CompRhys:
> some potentials don't support mixed pbc, so they won't benefit from this change until upstream adds support. To keep them working, `_to_legacy_pbc_state` feeds them a state whose pbc temporarily violates the `(n_systems, 3)` invariant, which I'm slightly uneasy about (see Discussion points; mixed-pbc batches raise a ValueError there).

Yeah this is likely a reason to hold off merging this for the time being. I hope to have some more time to review soon but have been busy with an international move.

lil-lon:
No rush at all, and good luck with the move!

I think the concern is reasonable, but the invariant break is confined to the wrapped forward calls and is meant to be temporary until the upstream integrations accept per-system pbc. I checked that this fix should be straightforward: the necessary change is only a 2-3 line shape check.

Overall I think the benefits outweigh this temporary cost. Besides improving batching efficiency for mixed datasets, storing pbc per system makes it behave like cell everywhere: slicing, splitting, popping, and concatenation handle it consistently. In particular, `concatenate_states` used to silently adopt the first state's pbc for mixed inputs.

Issue #597, opened 2026-08-14
VACF calculation does not seem to respect batching


jwjeffr:
I've been trying to use the `properties.correlations.VelocityAutoCorrelation` calculator in batched MD simulations, but it seems that only one VACF trajectory is reported. I think it's intuitive that the `vacf` property would, in this case, be a tensor of shape `(*, n_systems)`.

Is this intended behavior? If not, I'm happy to write up a small example. If so, I think this would be a very nice enhancement. I have a custom implementation for this I'm also happy to show off, but it's a little hacky.

CompRhys:
If it's summing multiple systems into a single result then that's definitely not intended behaviour. Please share any relevant examples and then if you have a fix we can see whether or not there's a clean way to do it to fix in prod. I don't think this code will have been looked at much/at-all since initial release.

jwjeffr:
See the example below. Batched MD for different temperatures, where I'd expect that the vacf would have a temperature index

```py
r"""
Script is computing the VACF for bulk tungsten at different temperatures using a MACE-MP potential

Environment
-----------
Python: 3.13.13
GPU/CUDA: NVIDIA H100 80GB HBM3, CUDA version 13.0
Packages:
    torch-sim-atomstic==0.6.1
    ase==3.29.0
    mace-torch==0.3.16
"""

from ase.build import bulk
from mace.calculators.foundations_models import mace_mp
from torch_sim.models.mace import MaceModel
from torch_sim.properties.correlations import VelocityAutoCorrelation
import torch_sim as ts


DEVICE: str = "cuda"
SAMPLE_STRIDE: int = 10
NUM_STEPS: int = 1_000
TEMPERATURES: list[int] = [300, 600, 900, 1200]


def main():

    vacf_calc = VelocityAutoCorrelation(window_size=100, device=DEVICE, use_running_average=True)
    reporter = ts.TrajectoryReporter(
        filenames=[f".vacf_window_counter{temp:0d}.h5" for temp in TEMPERATURES],
        prop_calculators={SAMPLE_STRIDE: {"vacf": vacf_calc}}
    )

    mace = mace_mp(model="large", return_raw_model=True)
    tungsten = bulk("W", cubic=True).repeat((4, 4, 4))
    _ = ts.integrate(
        system=[tungsten.copy()] * len(TEMPERATURES),
        model=MaceModel(model=mace, device=DEVICE),
        n_steps=NUM_STEPS,
        timestep=1.0e-3,
[...]

PR #598, opened 2026-08-15
(feat) optional mass weighting in VACF

jwjeffr:
## Summary

This PR adds an optional mass weight option to the VACF calculation in `properties.correlations.VelocityAutoCorrelation`, similar to the implementation in `pwtools`, which is necessary for getting accurate spectra in multicomponent systems.

## Feature

* Added mass weighting as a Boolean flag which defaults to `False` for backwards compatibility. I.e., if the user passes `mass_weighted=True` into the initializer, then the VACF is:

$$ C(t) \propto \sum_i m_i \left\langle \mathbf{v}_i(0)\cdot\mathbf{v}_i(t)\right\rangle$$

rather than the unweighted case where $m_i = 1$. The main changes are [here](https://github.com/jwjeffr/torch-sim/blob/24e157ff4de639f2f7aa5abb835f12783ce499ed/torch_sim/properties/correlations.py#L467-L474), and are similar to the implementation in `pwtools` [here](https://github.com/elcorto/pwtools/blob/a42f7e0a831b7cf6d11fa5dfde7d07cffab24d2b/src/pwtools/pydos.py#L57-L67).

## Checklist

### DIFF
--- a/tests/test_correlations.py
+++ b/tests/test_correlations.py
@@ -28,9 +28,19 @@ class MockState:
     the components needed for correlation calculations.
     """
 
-    def __init__(self, velocities: torch.Tensor, device: torch.device) -> None:
+    def __init__(
+        self,
+        velocities: torch.Tensor,
+        device: torch.device,
+        masses: torch.Tensor | None = None,
+    ) -> None:
         """Initialize mock state with provided data."""
         self.velocities = velocities
+        self.masses = (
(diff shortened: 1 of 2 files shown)


CompRhys:
Does this PR not address #597 correct this is just a fix to running a single system through VACF?

jwjeffr:
Correct, this is a different feature entirely

I'm hesitant to open a PR for #597 at this point just b/c it's not obvious to me how it can be cleanly implemented yet

PR #602, opened 2026-08-29
Fix/ase fire deform grad forces

curtischong:
## Summary

The problem lies in this code https://github.com/curtischong/torch-sim/blob/98c3ca52d5e6e45030efbbc3442ba99f91e2a335/torch_sim/optimizers/fire.py#L323-L328:

```python
cur_deform_grad = cell_filters.deform_grad(
  state.row_vector_cell,
  getattr(state, "reference_row_vector_cell", state.row_vector_cell), # L325
)
```

Since the fire state (which is a child of CellOptimState) never has `reference_row_vector_cell`,  the `state.row_vector_cell` is always used instead.

Here is the definition of CellOptimState, we can see it doesn't have `reference_row_vector_cell`.

https://github.com/curtischong/torch-sim/blob/98c3ca52d5e6e45030efbbc3442ba99f91e2a335/torch_sim/optimizers/cell_filters.py#L427

Adding the DeformGradMixin to CellOptimState gives it the `reference_row_vector_cell` attribute.

I found this bug bc I saw that torchsim didn't match the ASE implementation. I ran the test script below to show that our optimizer diverges from ASE's because of this bug.

This was the script used to replicate the bug on main

```python
"""Does torch-sim's ase_fire + frechet cell filter track ASE's FIRE + FrechetCellFilter?

ASE's cell filters hand FIRE ``forces @ deform_grad`` for the atomic rows, where
``deform_grad`` maps the reference cell onto the current cell. Both runs start
from the same 5%-strained, rattled Ar supercell with matched Lennard-Jones
parameters and FIRE settings, so the trajectories should be identical.
"""

import numpy as np
import torch
from ase.build import bulk
from ase.calculators.lj import LennardJones
from ase.filters import FrechetCellFilter
from ase.optimize import FIRE

import torch_sim as ts

lj = dict(sigma=3.405, epsilon=0.0104, cutoff=2.5 * 3.405)
dt, max_step = 1.0, 10.0  # large so the cell moves appreciably each step
pos_tol, cell_tol = 1e-8, 1e-6  # Ã; fp noise is ~1e-11 / ~1e-7 respectively

atoms = bulk("Ar", "fcc", a=5.26 * 1.05, cubic=True).repeat((2, 2, 2))
atoms.rattle(stdev=0.1, seed=0)
atoms.calc = LennardJones(sigma=lj["sigma"], epsilon=lj["epsilon"], rc=lj["cutoff"])
ase_fire = FIRE(FrechetCellFilter(atoms), logfile=None, dt=dt, maxstep=max_step)

model = LennardJonesModel(**lj, dtype=torch.float64, compute_stress=True)
state = ts.fire_init(
    ts.io.atoms_to_state(atoms, device="cpu", dtype=torch.float64),
    model,
    cell_filter=ts.CellFilter.frechet,
    dt_start=dt,
)

print(f"torch-sim {ts.__version__} ase_fire + CellFilter.frechet  vs  ASE FIRE + FrechetCellFilter")
print(f"{len(atoms)}-atom Ar supercell, 5% strained + rattled, LJ, {dt=}, {max_step=}\n"
print(f"{'step':>4} | {'max |pos_ts - pos_ase| (Ã)':>26} | {'max |cell_ts - cell_ase| (Ã)':>28} | status")
print("-" * 80)
all_match = True
for step in range(1, 6):
    ase_fire.step()
    state = ts.fire_step(state, model, max_step=max_step)
    pos_err = np.abs(state.positions.numpy() - atoms.positions).max()
    cell_err = np.abs(state.cell[0].numpy() - atoms.cell.array).max()
    match = pos_err < pos_tol and cell_err < cell_tol
    all_match &= match
    print(f"{step:>4} | {pos_err:>26.1e} | {cell_err:>28.1e} | {'match' if match else 'DIVERGED'}")

print(
    "\nStep 1 matches regardless: the cell still equals the reference cell, so"
    " forces @ deform_grad == forces. From step 2 on the cell has moved and the transform matters."
)
print(
    "\nRESULT: torch-sim ase_fire "
    + ("TRACKS" if all_match else "DOES NOT TRACK")
    + " ASE FIRE + FrechetCellFilter "
    + f"(tolerance: pos < {pos_tol:.0e} Ã, cell < {cell_tol:.0e} Ã)"
)
```

### Results:

main:
```
step | max |pos_ts - pos_ase| (Ã) | max |cell_ts - cell_ase| (Ã) | status
--------------------------------------------------------------------------------
   1 |                    1.4e-12 |                      1.8e-12 | match
   2 |                    4.8e-05 |                      2.1e-06 | DIVERGED
   3 |                    1.6e-04 |                      8.5e-06 | DIVERGED
   4 |                    3.9e-04 |                      2.8e-05 | DIVERGED
   5 |                    8.4e-04 |                      1.1e-04 | DIVERGED
```

this branch:
```
step | max |pos_ts - pos_ase| (Ã) | max |cell_ts - cell_ase| (Ã) | status
--------------------------------------------------------------------------------
   1 |                    1.4e-12 |                      1.8e-12 | match
   2 |                    2.7e-12 |                      1.0e-08 | match
   3 |                    3.6e-12 |                      6.0e-08 | match
   4 |                    2.0e-12 |                      1.5e-07 | match
   5 |                    4.6e-11 |                      1.9e-07 | match
```

### DIFF
--- a/torch_sim/optimizers/bfgs.py
+++ b/torch_sim/optimizers/bfgs.py
@@ -247,25 +247,8 @@ def bfgs_step(  # noqa: C901, PLR0915
     )
 
     if isinstance(state, CellBFGSState):
-        # Get current deformation gradient
-        # reference_cell.mT: [S, 3, 3], row_vector_cell: [S, 3, 3]
-        cur_deform_grad = cell_filters.deform_grad(
-            state.reference_cell.mT, state.row_vector_cell
-        )  # [S, 3, 3]
-
-        # Transform forces to scaled coordinates
-        # forces: [N, 3], cur_deform_grad[system_idx]: [N, 3, 3]
-        forces_scaled = torch.bmm(
-            state.forces.unsqueeze(1),  # [N, 1, 3]
-            cur_deform_grad[state.system_idx],  # [N, 3, 3]
-        ).squeeze(1)  # [N, 3]
-
-        # Current fractional positions
-        # positions: [N, 3] -> frac_positions: [N, 3]
-        frac_positions = torch.linalg.solve(
-            cur_deform_grad[state.system_idx],  # [N, 3, 3]
-            state.positions.unsqueeze(-1),  # [N, 3, 1]
-        ).squeeze(-1)  # [N, 3]
+        forces_scaled = state.deform_grad_forces()  # [N, 3]
+        frac_positions = state.frac_positions()  # [N, 3]
 
         # Pack into dense tensors [N, 3] -> [S, M, 3] -> [S, D]
         # For cell state, prev_positions is already fractional (stored that way)
@@ -495,15 +478,7 @@ def bfgs_step(  # noqa: C901, PLR0915
         # Apply position step in fractional space, then convert to Cartesian
         new_frac = frac_positions + flat_step  # [N, 3]
 
-        new_deform_grad = cell_filters.deform_grad(
-            state.reference_cell.mT, state.row_vector_cell
-        )  # [S, 3, 3]
-        # new_positions = new_frac @ deform_grad^T
-        new_positions = torch.bmm(
-            new_frac.unsqueeze(1),  # [N, 1, 3]
-            new_deform_grad[state.system_idx].transpose(-2, -1),  # [N, 3, 3]
-        ).squeeze(1)  # [N, 3]
-        state.set_constrained_positions(new_positions)  # [N, 3]
+        state.set_constrained_positions(state.positions_from_frac(new_frac))  # [N, 3]
     else:
         state.prev_positions = state.positions.clone()  # [N, 3]
         state.prev_forces = state.forces.clone()  # [N, 3]
--- a/torch_sim/optimizers/cell_filters.py
+++ b/torch_sim/optimizers/cell_filters.py
@@ -17,7 +17,7 @@ import torch
 import torch_sim.math as tsm
 from torch_sim.models.interface import ModelInterface
 from torch_sim.optimizers.state import BFGSState, FireState, LBFGSState, OptimState
-from torch_sim.state import SimState
+from torch_sim.state import DeformGradMixin, SimState
 
 
 MAX_LOG_DEFORM = 2.0
@@ -299,7 +299,7 @@ def unit_cell_step[T: AnyCellState](state: T, cell_lr: float | torch.Tensor) -> 
         cell_lr = cell_lr.expand(state.n_systems)
 
     # Get current deformation gradient
-    cur_deform_grad = deform_grad(state.reference_cell.mT, state.row_vector_cell)
+    cur_deform_grad = state.deform_grad()
 
     # Calculate cell positions from current deformation gradient
     cell_factor_expanded = state.cell_factor.expand(state.n_systems, 3, 1)
@@ -371,7 +371,7 @@ def compute_cell_forces[T: AnyCellState](
 
     if is_frechet:
         # Frechet cell force computation
-        cur_deform_grad = deform_grad(state.reference_cell.mT, state.row_vector_cell)
+        cur_deform_grad = state.deform_grad()
         ucf_cell_grad = torch.bmm(
             virial, torch.linalg.inv(torch.transpose(cur_deform_grad, 1, 2))
         )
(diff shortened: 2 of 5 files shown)


torch_sim/optimizers/cell_filters.py
```diff
@@ -424,10 +424,9 @@ def get_cell_filter(cell_filter: "CellFilter | tuple") -> CellFilterFuncs:
 
 
 @dataclass(kw_only=True)
-class CellOptimState(OptimState):
+class CellOptimState(OptimState, DeformGradMixin):
```
curtischong, line 427: adding this mixin is the key fix of this PR, it gives the `CellOptimState` the `reference_row_vector_cell` attribute.

torch_sim/optimizers/bfgs.py
```diff
@@ -247,18 +247,8 @@ def bfgs_step(  # noqa: C901, PLR0915
     )
 
     if isinstance(state, CellBFGSState):
-        # Get current deformation gradient
-        # reference_cell.mT: [S, 3, 3], row_vector_cell: [S, 3, 3]
-        cur_deform_grad = cell_filters.deform_grad(
-            state.reference_cell.mT, state.row_vector_cell
```
curtischong, line 253: this is such a subtle bug that ONLY affected FIRE, and NOT the BFGS or L-BFGS optimizers since the BFGS optimizers read the `reference_cell` directly. whereas if you look at fire, it looked for `getattr(state, "reference_row_vector_cell", `

curtischong: I could've fixed this bug by making the fire implementation match the other 2, but it's cleaner to just add the DeformGradMixin to the CellOptimState and use shared helper functions

torch_sim/optimizers/cell_filters.py
```diff
@@ -453,6 +452,22 @@ class CellOptimState(OptimState):
         "frechet_method",
     }
 
+    def deform_grad_forces(self) -> torch.Tensor:
```
curtischong, line 455: I considered adding this function to DeformGradMixin but decided against it since it needs forces and system_idx which are missing from DeformGradMixin but CellOptimState provides

torch_sim/optimizers/cell_filters.py
```diff
@@ -453,6 +452,22 @@ class CellOptimState(OptimState):
...
+        Mirrors the transform ASE's ``get_forces_unitcellfilter`` and
+        ``get_forces_frechet`` apply to the atomic forces. Equals ``forces`` when
+        the cell is undeformed relative to the reference cell.
+
+        Returns:
+            The transformed atomic forces, shape (n_atoms, 3)
+        """
+        # per-atom row vector @ its system's deform_grad:
+        # (n_atoms, 1, 3) @ (n_atoms, 3, 3) -> (n_atoms, 1, 3) -> (n_atoms, 3)
+        return torch.bmm(
```
curtischong, line 467: ase's version of this function is here:
https://gitlab.com/ase/ase/-/blob/3.29.0/ase/_4/optimize/cellutil.py#L213

torch_sim/optimizers/bfgs.py
```diff
@@ -247,25 +247,8 @@ def bfgs_step(  # noqa: C901, PLR0915
...
-        )  # [S, 3, 3]
-
-        # Transform forces to scaled coordinates
-        # forces: [N, 3], cur_deform_grad[system_idx]: [N, 3, 3]
-        forces_scaled = torch.bmm(
```
curtischong, line 258: fire, bfgs, and l-bfgs all individually calculate frac_positions which we will use the `state.frac_positions()` helper function now. I think this bug arose since we had different implementations to calculate forces_scaled, so factoring out all this logic into the same helpers is defensive programming

curtischong:
This PR comes before https://github.com/curtischong/torch-sim/pull/18 (it's rebased off this one for clarity), but I'd probably review that one too if you're in a reviewing spree this weekend rhys

CompRhys:
Thanks for report, Can you look into why https://github.com/curtischong/torch-sim/blob/98c3ca52d5e6e45030efbbc3442ba99f91e2a335/tests/test_optimizers_vs_ase.py didn't catch this regression? I am fairly certain that the extensive work of @t-reents did ensure we matched to within floating point accuracy and that test was supposed to prevent regression.

t-reents:
It seems that the thresholds are pretty large, so they would only capture the obvious failures/deviations (see https://github.com/TorchSim/torch-sim/pull/200#issue-3071056083) that we observed before achieving the agreement between ASE and torch-sim:

https://github.com/TorchSim/torch-sim/blob/17c14a054c9857d326f6974abe804829aeb1e285/tests/test_optimizers_vs_ase.py#L190-L220

Moreover, I just checked the old versions and we initially always relied on the `DeformGradMixin`, so the correct `reference_row_vector_cell` was used back then (IIRC, ase and torch-sim also agreed with much tighter thresholds when working on it, which would support this statement). See for example v0.3:

https://github.com/TorchSim/torch-sim/blob/cde91e98d93318a9a0a598701b3f18f03b068d88/torch_sim/optimizers.py#L1559-L1563

PR #603, opened 2026-08-31
alchemiops.py: Flag as unavailable on non-NVIDIA torch builds

yantar92:
## Summary

On ROCm builds, nvalchemiops correctly imports but does not work during runtime, yielding

  File "<...>/lib/python3.12/site-packages/torch_sim/neighbors/__init__.py", line 86, in torchsim_nl
    return alchemiops_nl_n2(
           ^^^^^^^^^^^^^^^^^
  File "<...>/lib/python3.12/site-packages/torch_sim/neighbors/alchemiops.py", line 62, in alchemiops_nl_n2
    res = _batch_naive_neighbor_list(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<...>/lib/python3.12/site-packages/nvalchemiops/torch/neighbors/batch_naive.py", line 1513, in batch_naive_neighbor_list
    compute_naive_num_shifts(cell, cutoff, pbc)
  File "<...>/lib/python3.12/site-packages/nvalchemiops/torch/neighbors/neighbor_utils.py", line 326, in compute_naive_num_shifts
    wp_device = wp.device_from_torch(device)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<...>/lib/python3.12/site-packages/warp/_src/torch.py", line 39, in device_from_torch
    return warp._src.context.runtime.cuda_devices[torch_device.index]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
IndexError: list index out of range

## Checklist

* [X] Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google), but not in `_import_nvalchemiops_batch_neighbors`, which I did not touch.
* [X] Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code. (no new failures)

<!-- We highly recommended installing the `prek` hooks running in CI locally to speedup the development process. Simply run `pip install prek && prek install` to install the hooks which will check your code before each commit. -->

### DIFF
--- a/tests/test_neighbors.py
+++ b/tests/test_neighbors.py
@@ -528,6 +528,27 @@ def test_fallback_when_alchemiops_unavailable(monkeypatch: pytest.MonkeyPatch) -
     assert mapping2.shape[1] > 0
 
 
+def test_alchemiops_import_guard_non_nvidia_builds(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """nvalchemiops is flagged unavailable on non-NVIDIA GPU torch builds.
+
+    On ROCm builds ``torch.version.cuda`` is ``None`` even though a GPU is
+    present, and on CPU-only builds ``torch.cuda.is_available()`` is ``False``.
+    alcheimops works on CPUs, and on NVIDIA GPUs, but not on non-NVIDIA GPUs.
+    """
+    from torch_sim.neighbors.alchemiops import _import_nvalchemiops_batch_neighbors
+
+    # Simulate a ROCm build: HIP is available, but there is no CUDA runtime.
+    monkeypatch.setattr(torch.version, "cuda", None)
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: True)
+    assert _import_nvalchemiops_batch_neighbors() is None
+
+    # Simulate a CPU build.
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: False)
+    assert _import_nvalchemiops_batch_neighbors() is not None
+
+
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="GPU not available for testing")
 def test_torchsim_nl_gpu() -> None:
     """Test that torchsim_nl works on GPU (CUDA/ROCm)."""
--- a/torch_sim/neighbors/alchemiops.py
+++ b/torch_sim/neighbors/alchemiops.py
@@ -17,6 +17,13 @@ _batch_cell_list: object | None = None
 
 def _import_nvalchemiops_batch_neighbors() -> tuple[object, object] | None:
     """Return ``(batch_cell_list, batch_naive_neighbor_list)`` if importable."""
+    # nvalchemiops is NVIDIA-CUDA-only (built on warp) for GPUs
+    # It does not work on non-NVIDIA builds.  In particular, on ROCm
+    # builds, nvalchemiops will fail during runtime, even though
+    # Python bindings are imported without issues.
+    if torch.version.cuda is None\
+       and torch.cuda.is_available():  # On ROCm, == True
+        return None
     try:
         from nvalchemiops.torch.neighbors import batch_cell_list as bcl
         from nvalchemiops.torch.neighbors import batch_naive_neighbor_list as bnl


CompRhys:
Thanks for the contribution, I totally understand the issue but I am not sure this is the right fix because on CPU achemiops should be the default so the check needs to check if there is a non-cuda accelerator not just if there is not a cuda accelerator.

yantar92:
Oops. Did not realize that alchemiops works on CPUs. I have updated the commit.

lil-lon:
Btw, if you want to detect ROCm, you can use `torch.version.hip` variable.

yantar92:
> Btw, if you want to detect ROCm, you can use `torch.version.hip` variable.

I have considered it, but AFAIU alchemiops will fail on any GPU that is not CUDA.

CompRhys:
I think that perhaps it's hard to do in logic for everyone's preferences so perhaps set an environment variable 'TS_USE_ALCHEMIOPS=false' and skip on that is most direct approach and easy to document. If not set default to true.  [edited for clarity]

yantar92:
I am ok with that approach, but that would require more significant changes.

1. `ALCHEMIOPS_AVAILABLE` is currently used for load-time checks. They will likely need to be moved to runtime.
2. Failure to run alcheimops-related functions should better provide some hints about existance of this user toggle. At least, that will greatly simplify debugging the issue I encountered.

CompRhys:
> 1. `ALCHEMIOPS_AVAILABLE` is currently used for load-time checks. They will likely need to be moved to runtime.

I don't think that adding another check for setting ALCHEMIOPS_AVAILABLE  is a significant change it is less lines of code that this solution, indeed it could be a single line change without needing explicit tests imo. My opposition to this solution is that relying on 'torch.cuda.is_available()' to be true for a non-cuda device to get the logic tree to work for cpus is just very indirect and impossible to parse without knowing that amd just hides behind the cuda api and I would hope amd would avoid that obvious misdirection in future api updates.

> 2. Failure to run alcheimops-related functions should better provide some hints about existance of this user toggle. At least, that will greatly simplify debugging the issue I encountered.

Any brief note about the env variable would be surfaced instantly by an agent in debugging so I think that the env variable covers that documentation side. However if you're using TS on ROCm a lot it could also be cool to create a dedicated docs page called something like 'TorchSim on AMD' to capture all the learnings you have.

yantar92:
> I don't think that adding another check for setting ALCHEMIOPS_AVAILABLE is a significant change it is less lines of code that this solution, indeed it could be a single line change without needing explicit tests imo.

Do I understand correctly that you are suggesting something like

```python
if not TS_USE_ALCHEMIOPS:
    return None
```

inside `_import_nvalchemiops_batch_neighbors`?

My concern here is that `_import_nvalchemiops_batch_neighbors` is called early, during `torch_sim.neighbors` module init. It means that the users will have to carefully arrange setting the proposed new flag to be set **before** loading anything that triggers this import. That's why I though that it will require something more complex that a one-liner change. (Just in case, I have no problem with the proposed approach with a flag, just discussing the implementation).

> My opposition to this solution is that relying on 'torch.cuda.is_available()' to be true for a non-cuda device to get the logic tree to work for cpus is just very indirect and impossible to parse without knowing that amd just hides behind the cuda api and I would hope amd would avoid that obvious misdirection in future api updates.

Fair. What about the alternative proposed by @lil-lon with `torch.version.hip`? I am thinking about both disabling alchemiops for ROCm and providing the user-level flag.

> Any brief note about the env variable would be surfaced instantly by an agent in debugging so I think that the env variable covers that documentation side.

I personally prefer to more direct approach, but you are the one to decide.

> However if you're using TS on ROCm a lot it could also be cool to create a dedicated docs page called something like 'TorchSim on AMD' to capture all the learnings you have.

This is the first critical failure I encountered, really. Usually, on ROCm some NVIDIA-only optimizations are simply disabled (with a warning), but that does not crash the code.

yantar92:
> Do I understand correctly that you are suggesting something like

Oh, you meant **environment** variable. Then, my concern is indeed moot and the code change will be simple.

CompRhys:
> you meant **environment** variable

yep, sorry I wasn't clear!

yantar92:
I plan to amend my patches and implement the environment variable, but will do it a bit later.

CompRhys:
thanks, on both PRs please make sure that all linting/tests pass (You can ignore the metatomic tests and I can make a judgement call on coverage check)

CompRhys:
Closing in favour of https://github.com/TorchSim/torch-sim/issues/606 becoming the default given it is ROCm and CUDA compatible and allows elimination of dependency on alchemiops.

PR #604, opened 2026-09-01
Added DeePMD-kit PyTorch backend integration

rahulumrao:
## Summary

* Feature 1
* Fix 1

## Checklist


docs/reference/deepmd.rst
CompRhys, line 1: we don't use rst.

pyproject.toml
```diff
@@ -54,6 +54,7 @@ mace = ["mace-torch>=0.3.16"]
 # phono3py>=4.0.0: some older 3.x sdists (e.g. 3.2.0) fail to build and break uv resolution.
 mattersim = ["mattersim>=1.2.5", "phono3py>=4.0.0"]
 metatomic = ["metatomic-torchsim>=0.1.1", "metatomic-ase>=0.1.0", "upet>=0.2.0"]
+deepmd = ["deepmd_torchsim>=0.1.1"]
```
CompRhys: make sure to capture any deps clashes below.

tests/models/test_deepmd.py
CompRhys, line 1: We have standard testing approach for models and their interfaces. Using that infrastructure is a prerequisite to merging here.

docs/reference/deepmd.rst
rahulumrao: Removed the rst file. The module` torch_sim/models/deepmd.py `carries the docstring.

tests/models/test_deepmd.py
rahulumrao: Rewritten on the standard infrastructure and now pushed into GitHub. Test used the `DPA-3.1-3M` universal model, and passed all 22/22 test locally.

pyproject.toml
rahulumrao: Verified dependency compatibility, the added `deepmd = ["deepmd_torchsim>=0.1.1"]` doesn't introduce any conflicts. Because `deepmd_torchsim` is a thin wrapper that places no direct constraints on deepmd-kit or torch, it won't clash with the other extras.

I intentionally leave deepmd-kit out of the direct dependencies since its PyTorch backend requires specific, torch versions. Offloading those constraints to `deepmd_torchsim` own deepmd extra keeps torch-sim's dependency tree clean and decoupled.

CompRhys: Please pin a minimum version that works such that all someone would need to do if importing from top level is install the right option. Otherwise there's no benefit to having the ease of use first party import now we adopted an external model posture

In terms of clashes please check once you've pinned a minimum working set which models deepmd can solve the deps alongside and mark the uv clashes in the pyproject toml. This is also important for UX who use TS for interfacing to many models.

rahulumrao: Done! `torch-sim[deepmd]` now installs `deepmd-kit==3.1.3` with `torch==2.10.0`. All model extras work with `deepmd` except `fairchem`, which has a PyTorch version conflict. Added fairchem to the conflicts list.

CompRhys:
Closing in favour of #606 becoming the default given it is ROCm and CUDA compatible and allows elimination of dependency on alchemiops.

PR #605, opened 2026-09-10
Enhance pymatgen io functions to support system and atom extras

akwarii:
## Summary

Extends pymatgen related functions in `torch_sim/io.py` with system_extras_map/atom_extras_map keyword arguments, matching the API already available for ASE.

Right now this mimics the ASE implementation, i.e. the user must manually pass the fields (s)he want to copy. Would you consider making automatic copying of all fields the default behavior instead? It would be very valuable in high-throughput workflows such as atomate2, where the conversion is often called by an internal wrapper the end user has no way to pass a map through to.

## Checklist

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -7,6 +7,7 @@ import pytest
 import torch
 from ase import Atoms
 from ase.build import molecule
+from pymatgen.core import Lattice, Structure
 
 import torch_sim as ts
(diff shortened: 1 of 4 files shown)


akwarii:
The tests errors seems unrelated to my PR

CompRhys:
LGTM, thanks for the contribution!

